PR #15, opened 2018-03-28
implement OpenAPI from single config

tomkralidis:
- generates OpenAPI config from a single configuration
- adds a pygeoapi subcommand


jorgejesus:
This PR also seems to clean the missing description on pygeoapi-config.yml. Using 0.0.0.0 is necessary for docker creation

PR #21, opened 2018-04-02
fix travis

tomkralidis:


pygeoapi/provider/sqlite.py
```diff
@@ -105,7 +105,6 @@ def __load(self):
         else:
             raise InvalidProviderError
 
-        conn.enable_load_extension(True)
```
jorgejesus: dont know what is the default load extension value but better to have it on the code

tomkralidis: Good catch; added back.

PR #36, opened 2018-04-29
Remove GeoJSON dependecy from sqlite module

### LINKED ISSUE #35: sqlite depends on generic geojson package - package name collition
Sqlite provider uses package geojson,  same folder contains module geojson and in python2.7 this module is imported causing problems (Python3 has no problems)

Drop dependence of geojson use only json python package

jorgejesus:
Removal of GeoJSON package dependence in favour of generic JSON package

Prevention of package/module name colision

Closes #35


requirements.txt
```diff
@@ -4,4 +4,4 @@ Flask
 Shapely
 Flask_Cors
 PyYAML
-geojson
+json
```
tomkralidis: `json` ships with Python and thus doesn't need to be specified in `requirements.txt`.

PR #68, opened 2018-10-02
add X-Powered-By header to all HTTP responses

tomkralidis:


tomkralidis:
Note that travis error is unrelated (will fix in next PR).

PR #72, opened 2018-11-13
Postgis provider support

jorgejesus:
PostGIS support using generic SQL in-sync queries using psycopg2 library

Provider supports schema and implements server side cursors for item pagination


openapi.yml
```diff
@@ -0,0 +1 @@
+[2018-09-07T20:37:57Z] {/home/jesus/git/pygeoapi/pygeoapi/provider/elasticsearch_.py:69} ERROR - Cannot connect to Elasticsearch
```
tomkralidis: Is this file needed?

pygeoapi-openapi.yml
```diff
@@ -0,0 +1,273 @@
+components:
```
tomkralidis: Is this file needed also? We have a utility to generate this already.

pygeoapi/provider/__init__.py
```diff
@@ -37,7 +37,8 @@
     'Elasticsearch': 'pygeoapi.provider.elasticsearch_.ElasticsearchProvider',
     'GeoJSON': 'pygeoapi.provider.geojson.GeoJSONProvider',
     'GeoPackage': 'pygeoapi.provider.geopackage.GeoPackageProvider',
-    'SQLite': 'pygeoapi.provider.sqlite.SQLiteProvider'
+    'SQLite': 'pygeoapi.provider.sqlite.SQLiteProvider',
+    'PostgreSQL': 'pygeoapi.provider.postgresql.PostgreSQLProvider'
```
tomkralidis: Can we try to keep to alphabetical order?

pygeoapi/provider/postgresql.py
```diff
@@ -0,0 +1,295 @@
...
+        :returns: GeoJSON FeaturesCollection
+        """
+        LOGGER.debug('Querying PostGIS')
+
+        if resulttype == 'hits':
+
+            with DatabaseConnection(self.conn_dic,
+                                    self.table, context="hits") as db:
+                cursor = db.conn.cursor(cursor_factory=RealDictCursor)
+                sql_query = SQL("select count(*) as hits from {}").\
```
tomkralidis: Can we implement binding here to make safer against SQL injections?

jorgejesus: Identifier(self.table) on line 182 should take care of the SQL injection as indicated in http://initd.org/psycopg/docs/sql.html  as explained on class psycopg2.sql.SQL(string) and on  class psycopg2.sql.Identifier(*strings)

Issue #26, opened 2018-04-10
enhance HTML representations


tomkralidis:
The `f=html` responses need a better layout.  The goal here should be to enhance layout (especially the maps) without introducing too many external dependencies.

jorgejesus:
Something every simple like: https://bulma.io/documentation/

jorgejesus:
Suggestion from Paul van Genuchten

Make proposed page easy to customize using configuration options. and possibility of adding icons

pvgenuchten:
The templating system seems a good option to change the layout according to user needs
https://github.com/geopython/pygeoapi/blob/master/pygeoapi/templates/base.html

Some comments to improve the generic base template:
- <head>title/keyword/abstract</head> should not be filled using generic config, in stead it should use properties from the collection or feature exposed on that page
- if the css were based on less, it would be nice to allow to define a color-scheme & font.
- add an organisation logo (upload logo.png from admin)
- add option to configure links in the header or footer (contact, disclaimer, help etc)

thijsbrentjens:
@pvgenuchten this is a really good point: 

> head-title/keyword/abstract should not be filled using generic config, in stead it should use properties from the collection or feature exposed on that page (title will be displayed on google search result), so would be interesting to configure for each Collection which property of Feature can be used as title (aka schema.org/name)

I just tested the results in Google Search. Google indexes up to the individual item, which is really nice (and *could* be useful for some datasets). But the HTML representation could be improved. See for example [https://www.google.com/search?q=site:demo.pygeoapi.io&start=70](https://www.google.com/search?q=site:demo.pygeoapi.io&start=70)

Improvements like: add a specific H1 header (and title element) for an item (there is no H1 currently), preferably a human readable title, but that may require some more work (e.g. an optional ```title_field``` per provider for example). It may not be today, but I could maybe implement something for this

tomkralidis:
Closing given we probably have enough given our current HTML representation.  Feel free to reopen if we missed anything.  Having said this, I will open a new ticket for being able to modify look and feel.

pvgenuchten:
I agree with Thijs (sorry missed this discussion) we should make sure to list at least the id of the feature in the title, now google shows same title for each feature

![image](https://user-images.githubusercontent.com/299829/68399784-3929d400-0177-11ea-8999-d04ba112d478.png)

I'll work on a PR

pvgenuchten:
or maybe better to move this to a new issue...

[...]

PR #92, opened 2019-04-12
introduce mini.css

pvgenuchten:
This is a proposal (as a start for #26) to show what a basic css framework can do
[mini.css](https://minicss.org) is one of the common minimal css frameworks out there (License MIT)

image


tomkralidis:
cc @justb4 (we have a similar situation in [GeoHealthCheck](https://github.com/geopython/GeoHealthCheck))

Thanks @pvgenuchten.  What is the lifecycle of mini.css?  How often should we update it?  This is true of any dependency really but at least our Python deps are one step closer with pip / debian requirements.  How do you see mini.css being maintained over time in pygeoapi?

pvgenuchten:
I'm also new to mini.css also, and neutral to a choice of framework, bootstrap or bulma also fine. mini.css has a bunch of releases, see https://cdnjs.com/libraries/mini.css (47kb). Since i'm not using the advanced features (yet), i guess updates would be limited. Are there any preferences within geopython team? Geonetwork, pycsw.org and ldproxy are bootstrap based (97kb). Both systems use similar naming conventions, eg col-md-6 etc

tomkralidis:
Thanks @pvgenuchten. I'm thinking about upkeep of mini.css in the long term against software (vs. a website).  In GHC part of our setup process includes pulling in JavaScript libs etc.  Should/do we consider this for pygeoapi or keep a static minified copy in version control?

pvgenuchten:
It brings a level of complexity to the build process, but sure can be introduced (at anytime). 

Then include also the leaflet css/javascript (now pulled from unpkg cdn).

tomkralidis:
Good point @pvgenuchten.  Is mini.css available from an unpkg cdn?  If yes, I would suggest we use that approach to be consistent to other js/css deps right now and address these deps as a whole in a separate issue.

pvgenuchten:
now using unpkg for mini.css

Issue #102, opened 2019-04-29
HTML UI issues from #92 (mini.css)


justb4:
Several issues found, some were there already before #92:

1. Stale links when navigating with Next|Prev through Collections

* navigate to http://localhost:5000/collections/lakes/items
* click in table on id single Item, Ok, shows single Item
* back to http://localhost:5000/collections/lakes/items
* click Next: ok, navigates to http://localhost:5000/collections/lakes/items/?startindex=10
* click single Item Id in table. e.g. id 10
* navigates to stale link: http://localhost:5000/collections/lakes/items/items/10

NB this was already seen before #92 

2. Ditto for single Item

* navigate to http://localhost:5000/collections/obs/items/238
* HTML+Map single Item shows
* click Next or Prev button
* Tries to show: http://localhost:5000/collections/obs/items/238&f=html which fails

3. JSON OpenAPI doc link

* go to http://localhost:5000/
* click the link `the OpenAPI definition as JSON` 
* opens http://localhost:5000/api showing HTML i.s.o. JSON (when #101 as well integrated)

Not sure what the standard requires, or if http://localhost:5000/api?f=json should be explicitly rendered?

justb4:
Problems 1) and 2): some html links have trailing slash '/' in page URLs. In those cases relative links within the page get expanded to URLs with 'items/items..' i.s.o. 'items/..'. . Will solve.

justb4:
Item 3) : depends on content negotiation, default would be JSON. But in browser would be HTML, so need explicit `f=json`.

justb4:
Can close IMO, pls re-open if still nav errors as listed above found.

justb4:
Still stale links in map markers, caused by 
https://github.com/geopython/pygeoapi/blob/master/pygeoapi/templates/items.html#L85 . One-liner fix needed similar to regular HTML ft id links in table view. Will fix.

justb4:
Closing, pls re-open if still nav errors as listed above found.

PR #103, opened 2019-05-01
#102 several fixes for (dead) links in navigating HTML

justb4:
Appears this issue was not really due to #92. 
At least no dead links anymore and JSON content when accessed in browser. 

Question remains that on single Item the `Next` and `Prev` links should really walk through Collection but I suggest to do that in other issue, as this is quite tricky to achieve (as one does not know the prev/next id's in isolation of single Item).  

Hold on..one failing test now...

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -102,7 +102,7 @@ class API(object):
         fcm['links'] = [{
               'rel': 'self',
               'type': 'application/json',
-              'title': 'this document',
+              'title': 'this document as JSON',
               'href': self.config['server']['url']
             }, {
               'rel': 'self',
@@ -125,6 +125,10 @@ class API(object):
         ]
 
         if format_ == 'html':  # render
+            for link in fcm['links']:
+                if 'json' in link['type']:
+                    link['href'] += '?f=json'
+
             headers_['Content-Type'] = 'text/html'
             content = _render_j2_template(self.config, 'root.html', fcm)
             return headers_, 200, content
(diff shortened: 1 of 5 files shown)


pygeoapi/flask_app.py
```diff
@@ -166,3 +166,7 @@ def serve(ctx, debug=False):
 #    setup_logger(CONFIG['logging'])
     APP.run(debug=True, host=api_.config['server']['bind']['host'],
             port=api_.config['server']['bind']['port'])
+
+
+if __name__ == '__main__':  # run locally, for testing
```
tomkralidis: Do we need this?  `pygeoapi serve` accomplishes this already.

pygeoapi/api.py
```diff
@@ -125,6 +125,10 @@ def root(self, headers, args):
         ]
 
         if format_ == 'html':  # render
+            for link in fcm['links']:
+                if 'json' in link['type']:
+                    link['href'] += '?f=json'
```
tomkralidis: is this better written as `link['href'] = ''.join((link['href'], '?f=json'))`?  I'm not sure what the current verdict is around 'adding' strings in Python.

pygeoapi/flask_app.py
justb4: Main reason for `main()` here is for (stepwise) debugging within an IDE like IntelliJ IDEA or WebStorm and probably PyCharm: needed easy entry point.

pygeoapi/api.py
justb4: From what I read both `+` and `join` can be used, with `join` being marginally faster with many strings to concatenate. For readablity `+` is preferred.

justb4:
Checks failed since one test need change in `headers()` fixture to emulate real HTTP (`Werkzeug`) Headers structure with embedded `environ`. Will add. This would also allow for more advanced tests simulating many variations of client requests (think of proxied requests, using SCRIPT_NAME and returning full URLs within responses (i.s.o http://localhost:5000).

justb4:
Also needed to fix the breadcrumbs relative URL paths: in cases where the page URL ends with `/` the breadcrumbs paths gave stale links.

Now all HTML links are based on Flask/Werkzeug `PATH_INFO`. This is more robust, as this also would work behind a proxy (when SCRIPT_NAME included).

PR #117, opened 2019-05-13
add HTML pages for processes, add raw mode option for process execution

tomkralidis:


tomkralidis:
cc @pvgenuchten for review

pygeoapi/api.py
```diff
@@ -586,6 +587,16 @@ def describe_processes(self, headers, args, process=None):
 
         headers_ = HEADERS.copy()
 
+        format_ = check_format(args, headers)
```
justb4: AFAIKS `check_format()` may return `None` in cases (e.g. when format other than `json` or `html` is requested, try e.g. `?f=gml`).
The code below must handle that case, returning exception, otherwise `json` is assumed.

tomkralidis:
@justb4 processing support in pygeoapi is ad hoc at the moment.  However it follows the WPS REST binding work that will come out of OGC. In the long run, in pygeoapi we are taking into account features, processing, coverages and search.  You are correct in that our processing support is decoupled from features/WFS3 per se.

justb4:
@tomkralidis thanks, clear, found: https://github.com/opengeospatial/wps-rest-binding as well.

pygeoapi/api.py
justb4: Hmm, rethinking, makes sense: by default JSON is returned, even if no format specified (e.g. when using curl), so forget above comment.

justb4:
Was missing, also in standard a specific example. Running the `wps-fixes` branch locally, this works (returns `outputs` with same content) as `inputs` in POST body for process execution:

```
{
    "inputs": [
        {
            "id": "name",
            "value": "hi there",
	    "type" : "text/plain"
        }
    ]
}
```
Maybe add as example somewhere.

tomkralidis:
Fixed.  FYI the Swagger error was based on the fact that we were referencing a YAML schema that didn't resolve.

justb4:
Ok, merged, we'll deal with any problems later.

PR #138, opened 2019-05-24
map url config

tomkralidis:
Provide the ability to set a specific tile server for HTML output displaying maps.

Not sure if I missed any parts of the codebase.

### DIFF
--- a/aws-lambda/pygeoapi-serverless-config.yml
+++ b/aws-lambda/pygeoapi-serverless-config.yml
@@ -10,6 +10,9 @@ server:
     pretty_print: true
     limit: 10
     # templates: /path/to/templates
+    map:
+        url: https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png
+        attribution: 'Wikimedia maps | Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
 
 logging:
     level: ERROR
--- a/aws-lambda/pygeoapi-serverless-config.yml
+++ b/aws-lambda/pygeoapi-serverless-config.yml
@@ -12,7 +12,7 @@ server:
     # templates: /path/to/templates
     map:
         url: https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png
-        attribution: 'Wikimedia maps | Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
+        attribution: '<a href="https://wikimediafoundation.org/wiki/Maps_Terms_of_Use">Wikimedia maps</a> | Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
 
 logging:
     level: ERROR
--- a/docker/default.config.yml
+++ b/docker/default.config.yml
@@ -14,7 +14,7 @@ server:
     # templates: /path/to/templates
     map:
         url: https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png
-        attribution: 'Wikimedia maps | Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
+        attribution: '<a href="https://wikimediafoundation.org/wiki/Maps_Terms_of_Use">Wikimedia maps</a> | Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
 
 logging:
     level: ERROR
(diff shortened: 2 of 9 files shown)


tomkralidis:
@justb4 thanks for the info.

> this change will break existing configurations, I suggest to provide a sensible default

Are we at the point of guaranteeing backwards compatability?  Given we are pre-1.0 IMHO anything goes

> using tile.osm.org in apps that are distributed is considered "not done" (see OSM policies, maybe some other provider?

Any suggestions?  We can then change in [GHC](https://github.com/geopython/GeoHealthCheck/blob/master/GeoHealthCheck/config_main.py#L85) as well.

justb4:
> @justb4 thanks for the info.
> 
> > this change will break existing configurations, I suggest to provide a sensible default
> 
> Are we at the point of guaranteeing backwards compatability? Given we are pre-1.0 IMHO anything goes
also true, need to at least  fix demo server.
> 
> > using tile.osm.org in apps that are distributed is considered "not done" (see OSM policies, maybe some other provider?
 
> Any suggestions? We can then change in [GHC](https://github.com/geopython/GeoHealthCheck/blob/master/GeoHealthCheck/config_main.py#L85) as well.

a list here https://wiki.openstreetmap.org/wiki/Tile_servers with TOUs. I use https://www.4umaps.com/mountain-bike-hiking-bicycle-outdoor-topographic-map.aspx but seems to have no global coverage. True for GHC as well.

justb4:
WIkiMedia Maps https://maps.wikimedia.org may be a candidate. 

Another item (config attr) that should be added in any case: "Map Attribution": fixed to Leaflet and OpenStreetMap contributors, plus the config-attr-provided Map Provider.

pvgenuchten:
Would be nice to also allow various types of background, eg wmts, tms, wms or even age-api-tiles :-)

An advanced configuration option could be to add a leaflet layer object as background configuration.

tomkralidis:
Thanks @justb4 and @pvgenuchten for comments.

- I've added a `server.map.attribution` directive in config
- in terms of defaults, different map configurations, perhaps we can have a `server.map.config` object which would point to a JavaScript snippet of some sort?  In either case, I would push this into another ticket as the goal of the PR here was to switch map providers based on the same design pattern

justb4:
All example configs point to `tile.osm.org` but `attribution` refers to Wikimedia. So either the config should be;

For tile.osm.org:

```
map:
        url: https://tile.osm.org/{z}/{x}/{y}.png
        attribution: 'Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
```

For WikiMedia Maps:

```
map:
        url: https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png
        attribution: '<a href="https://wikimediafoundation.org/wiki/Maps_Terms_of_Use">Wikimedia maps</a> | Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
```

Possibly each could be added in the various configs. Later in documentation these options (OSM or WikiMedia) can be addressed.

justb4:
Also check if the Leaflet attribution is still there or if we need to do: `map.attributionControl.addAttribution(myAttributions);`

See https://geoexamples.com/other/2018/01/08/leaflet-attribution.html

tomkralidis:
@justb4 thanks, URLs updated.  I'd prefer to leave the Leaflet attribution as part of the attributions added in the caller.

justb4:
Ok, only the full Attribution for WikiMedia should point to their TOU URL:

```
attribution: '<a href="https://wikimediafoundation.org/wiki/Maps_Terms_of_Use">Wikimedia maps</a> | Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
```

tomkralidis:
@justb4 once again ready for review.

justb4:
Now see one missed, e.g.:

https://demo.pygeoapi.io/master/collections/lakes/items/0 shows no attribution, missing in template:

https://github.com/geopython/pygeoapi/blob/master/pygeoapi/templates/item.html#L81

Also `subdomains` is not really valid, only goes for URLs of the form `https://${s}.domain.xy/bla//{z}/{x}/{y}.png`. But will be probably ignored.

tomkralidis:
Fixed in master, thanks.

justb4:
Note that Leaflet attribution is automatically added.

tomkralidis:
Right, as mentioned above:

>  I'd prefer to leave the Leaflet attribution as part of the attributions added in the caller.

PR #139, opened 2019-05-24
allow no processes to be defined (#136)

### LINKED ISSUE #136: pygeoapi openapi docgen fails when no processes configured in .yml
When leaving out the `processes` section within the `pygeoapi` config YAML or even making it empty, like:

```
    countries:
        title: Countries in the world
        description: Countries of the world
        keywords:
.
.
.            id_field: ogc_fid
            table: ne_110m_admin_0_countries

processes:
#    hello-world:
#        processor:
#            name: HelloWorld

```

generating the `openapi `doc fails with this exception:

```
  File "/usr/local/bin/pygeoapi", line 11, in <module>
    load_entry_point('pygeoapi', 'console_scripts', 'pygeoapi')()
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/pygeoapi/pygeoapi/openapi.py", line 481, in generate_openapi_document
    click.echo(yaml.safe_dump(get_oas(s), default_flow_style=False))
  File "/pygeoapi/pygeoapi/openapi.py", line 466, in get_oas
    return get_oas_30(cfg)
  File "/pygeoapi/pygeoapi/openapi.py", line 278, in get_oas_30
    for k, v in cfg['processes'].items():
AttributeError: 'NoneType' object has no attribute 'items'
```

So always at least one process needs to be configured. IMHO we should be able to leave `processes` out or empty.

tomkralidis:
Fixes #136

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -752,7 +752,7 @@ class API(object):
 
         processes = self.config.get('processes', {})
 
-        if not processes or process not in processes.keys():
+        if process not in processes:
             exception = {
                 'code': 'NotFound',
                 'description': 'identifier not found'
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -680,31 +680,37 @@ class API(object):
             LOGGER.error(exception)
             return headers_, 400, json.dumps(exception)
 
-        processes_config = self.config['processes']
+        processes_config = self.config.get('processes', {})
 
-        if process is not None:
-            if process not in processes_config.keys():
-                exception = {
-                    'code': 'NotFound',
-                    'description': 'identifier not found'
-                }
-                LOGGER.error(exception)
-                return headers_, 404, json.dumps(exception)
+        if processes_config:
+            if process is not None:
+                if process not in processes_config.keys():
+                    exception = {
+                        'code': 'NotFound',
+                        'description': 'identifier not found'
+                    }
+                    LOGGER.error(exception)
+                    return headers_, 404, json.dumps(exception)
 
-            p = load_plugin('process', processes_config[process]['processor'])
-            p.metadata['jobControlOptions'] = ['sync-execute']
-            p.metadata['outputTransmission'] = ['value']
-            response = p.metadata
-        else:
-            processes = []
-            for k, v in processes_config.items():
-                p = load_plugin('process', processes_config[k]['processor'])
+                p = load_plugin('process',
+                                processes_config[process]['processor'])
                 p.metadata['jobControlOptions'] = ['sync-execute']
                 p.metadata['outputTransmission'] = ['value']
-                processes.append(p.metadata)
-            response = {
-                'processes': processes
-            }
+                response = p.metadata
+            else:
+                processes = []
+                for k, v in processes_config.items():
+                    p = load_plugin('process',
+                                    processes_config[k]['processor'])
+                    p.metadata['jobControlOptions'] = ['sync-execute']
+                    p.metadata['outputTransmission'] = ['value']
+                    processes.append(p.metadata)
+                response = {
+                    'processes': processes
+                }
+        else:
+            processes = []
+            response = {'processes': processes}
 
         if format_ == 'html':  # render
             headers_['Content-Type'] = 'text/html'
@@ -744,7 +750,9 @@ class API(object):
             LOGGER.error(exception)
             return headers_, 400, json.dumps(exception)
 
-        if process not in self.config['processes'].keys():
+        processes = self.config.get('processes', {})
+
+        if not processes or process not in processes.keys():
             exception = {
                 'code': 'NotFound',
                 'description': 'identifier not found'
@@ -753,7 +761,7 @@ class API(object):
             return headers_, 404, json.dumps(exception)
 
         p = load_plugin('process',
-                        self.config['processes'][process]['processor'])
+                        processes[process]['processor'])
 
         data_ = json.loads(data)
         for input_ in data_['inputs']:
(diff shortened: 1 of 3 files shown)


justb4:
Suggestion: add some unit tests (cases: processes defined, no processes defined or empty)

justb4:
One issue/question: can `processes` be completely left out of the config (I would be in favour)?

Some of the code seems to assume that, but other (unchanged code) does not:, e.g. 

https://github.com/geopython/pygeoapi/blob/8495d8ea1073b5c5c03052d3673751428a004329/pygeoapi/api.py#L754

Optional: The general pattern I usually use in these kind of cases (`dict` item empty or not present)  to save code lines is:

i.s.o.

```
        if ('processes' not in self.config or
                len(self.config['processes']) == 0):
            processes = []
            response = {'processes': processes}
        else:
            processes_config = self.config['processes']

```

Use `dict` `get()` with default empty `dict` (assuming `processes_config` is of type `dict`)
```
processes_config = self.config.get('processes', {})
if len(processes_config) == 0:
   response = {'processes': []} 
else:
  
```

tomkralidis:
@justb4 updated.

tests/test_api.py
```diff
@@ -321,6 +329,14 @@ def test_execute_process(config, api_):
 
     assert response['outputs'][0]['value'] == 'test'
 
+    api_.config['processes'] = []
```
justb4: shouldn't empty value be `{}`?

pygeoapi/api.py
```diff
@@ -744,7 +750,9 @@ def execute_process(self, headers, args, data, process):
             LOGGER.error(exception)
             return headers_, 400, json.dumps(exception)
 
-        if process not in self.config['processes'].keys():
+        processes = self.config.get('processes', {})
+
+        if not processes or process not in processes.keys():
```
justb4: same here: `processes` will always have `dict` value so check could be `if process not in processes.keys():`..

pygeoapi/api.py
```diff
@@ -680,31 +680,37 @@ def describe_processes(self, headers, args, process=None):
...
 
-        if process is not None:
-            if process not in processes_config.keys():
-                exception = {
-                    'code': 'NotFound',
-                    'description': 'identifier not found'
-                }
-                LOGGER.error(exception)
-                return headers_, 404, json.dumps(exception)
+        if processes_config:
```
justb4: Notice that `processes_config` will always have a (`dict`) value now. First thought  the `else` case would never be reached. But then again line 685 will evaluate to `True` if `processes_config` is not `None` AND non-empty [as per Python rules](https://docs.python.org/2/library/stdtypes.html#truth-value-testing). So no change required.

tests/test_api.py
```diff
@@ -294,6 +294,14 @@ def test_describe_processes(config, api_):
     assert len(process['outputTransmission']) == 1
     assert len(process['jobControlOptions']) == 1
 
+    api_.config['processes'] = []
```
justb4: should be 
`api_.config['processes'] = {}`

pygeoapi/openapi.py
```diff
@@ -275,85 +275,88 @@ def get_oas_30(cfg):
     }
 
     LOGGER.debug('setting up processes')
-    for k, v in cfg['processes'].items():
-        p = load_plugin('process', v['processor'])
+    if 'processes' in cfg:
```
justb4: Why not the same convention as in `api.py`: `processes = cfg.get('processes', {})` and `if processes:` etc ?

tests/test_api.py
```diff
@@ -321,6 +329,14 @@ def test_execute_process(config, api_):
...
     assert response['outputs'][0]['value'] == 'test'
 
+    api_.config['processes'] = {}
+
+    req_headers = make_req_headers()
+    rsp_headers, code, response = api_.execute_process(
+        req_headers, {}, json.dumps(req_body), 'hello-world')
+    response = json.loads(response)
+    assert response['code'] == 'NotFound'
+
```
justb4: Add test case for processes not in config? i.e.

```
api_.config['processes'] = None
rsp_headers, code, response = api_.describe_processes(...
.
.
assert len(processes['processes']) == 0
```

tests/test_api.py
```diff
@@ -294,6 +294,14 @@ def test_describe_processes(config, api_):
...
     assert len(process['jobControlOptions']) == 1
 
+    api_.config['processes'] = {}
+
+    req_headers = make_req_headers()
+    rsp_headers, code, response = api_.describe_processes(
+        req_headers, {}, 'foo')
+    processes = json.loads(response)
+    assert len(processes['processes']) == 0
+
```
justb4: have you seen my suggestion for additional test? The essence of this issue is to allow `processes` to be left out of config (right?) but that case is never tested. Could even be done with a loop:

```
empty_vals = [None, {}]
for empty_val in empty_vals:
    api_.config['processes'] = empty_val

    req_headers = make_req_headers()
    rsp_headers, code, response = api_.describe_processes(
        req_headers, {}, 'foo')
    processes = json.loads(response)
    assert len(processes['processes']) == 0
```

justb4:
thanks @tomkralidis also fo patience!

Issue #134, opened 2019-05-22
display link or image if feature contains link


pvgenuchten:
if a feature property contains a link (starts with 'http') it should be displayed as a clickable hyperlink, i hope it can be implemented as filter in the template language:

```
{{ft.prop | link}}
```
This can be extended to convert links that end in .gif/.svg/.jpg/.png to 
```
<a href="{{ft.prop}}"><img src="{{ft.prop}}" class="thumbnail"/></a>
```
A test service (as wfs2 source) could be this one http://services.rce.geovoorziening.nl/landschapsatlas/wfs?request=GetCapabilities&service=WFS (various featuretypes have image references; kastelen, hunebedden, etc)

pvgenuchten:
[jinja2](http://jinja.pocoo.org/docs/2.10/templates/#list-of-builtin-filters) indeed has a filter for this
```
{{ft.prop|urlize}}
```

justb4:
You folks still want to add thumbnail images. Think there is a simple way and an advanced method.

Simple, just look at the extension of the attr URL, if it ends with `.jpeg|.jpg|.png|.gif`etc. The [RCE WFS ](http://services.rce.geovoorziening.nl/landschapsatlas/wfs?request=GetCapabilities&service=WFS) @pvgenuchten mentioned above has an interesting layer "castles" (Dutch `kastelen`). E.g. Castle Kernhem in Ede:

![image](https://user-images.githubusercontent.com/582630/58484597-a7974e80-8162-11e9-9678-239f5cc20818.png)

![image](https://user-images.githubusercontent.com/582630/58484656-c7c70d80-8162-11e9-9de0-ad4c2c616b61.png)

(hmm more of a Mansion I would say). 

The more advanced way would be to sniff Content-Type for each URL attribute and based on that display the image (or maybe other content like Audio or Video). 

Would go for the simple way first using a Python `regexp` (something like `\.(?i:jpg|gif|png|bmp)`)  for all valid Image extensions. Would make a great demo!

justb4:
I've added this `castles` layer to the demo services, but maybe also a nice layer for the [default config](https://github.com/geopython/pygeoapi/blob/master/pygeoapi-config.yml). Shall I add there as well? As it contains links to HTML and images.

tomkralidis:
Sure.  Having said this, should we break out our sample configs?  i.e.:

[...]

PR #140, opened 2019-05-26
adds links to feature properties as suggested in #134

pvgenuchten:
for test purposes adds links to great lakes dataset

image


justb4:
Great guys! This also percolated automagically to the demo app:

https://demo.pygeoapi.io/master/collections/lakes/items

PR #143, opened 2019-05-27
unpin dev requirements, fix flake8

tomkralidis:


justb4:
Reason why to unpin? At least the Docker build needs to be reproducible. It is failing now on `psycopg2`:
https://cloud.docker.com/u/geopython/repository/docker/geopython/pygeoapi/builds/58b5c369-5541-4df1-b670-f51a57d87237

Ok, now see, the discussion on Gitter, but that was about `gunicorn` (?). If we add `gunicorn` to the core (in which I am neutral since it is a deploy option) then it should be pinned as well. I can imagine that we need a base version (`requirements.txt`) without all Providers. And some Providers bring their own requirements. In particular GDAL can be a pain.

tomkralidis:
I unpinned mainly because new flake8 errors were going unnoticed given we were pinned to a specific version.  I have since opened #144.

Is there a way to test a Docker build as part of a pull request (like Travis)?

justb4:
Can't see the relation between `flake8` and `requirements` immediately....

Yes, Travis has Docker Service support. (under `services`). We can even run the tests with Docker Image.

tomkralidis:
The version of flake8 that was pinned in `requirements-dev.txt` did not catch flake8 errors (in this PR) that a later version of flake8 caught.

justb4:
Aha, but in that case I would always pin to the highest version and when there is a new version pin to that. Goes for all components. I remember we had a similar issue with GHC I think. Non-reproducible builds are even considered bad practice for security reasons. We can not fully tests everything. If all would use either `virtualenv` or Docker we can always go from one stable set of versions to the next.

tomkralidis:
We had a new version of flake8 and this went unnoticed given the pin of flake8 in master at the time. @kalxas had a good idea to have pinned versions only against stable branches.

PR #144, opened 2019-05-27
fix dev requirements

tomkralidis:


justb4:
Ok, this should get Docker built again (fingers crossed). IMO "to pin or not to pin" requires a broader discussion.

justb4:
Yes, Docker build ok (for now). Still the pinning. Also think that we have actually three sets of requirements:

* core (`requirements.txt`)
* providers (e.g. `requirements-providers.txt`)
* dev (`requirements-dev.txt`) for development and testing

tomkralidis:
Fair enough.  Having said this, any serious deployment will not use pip anyways.

PR #147, opened 2019-05-28
Update requirements.txt

bradh:
Require 5.1 to ensure `yaml.FullLoader` is available.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 click
 Flask
-PyYAML
+PyYAML>=5.1
 unicodecsv


tomkralidis:
cc @justb4 

Thanks @bradh.  I'd prefer merging #145 instead which covers pyyaml before and after 5.1, which also works for both pip and non-pip installations.  Does this make sense?

bradh:
Understood - sorry I didn't see the other PR.

justb4:
Great, thanks @bradh ! (wanted to make same suggestion).

PR #153, opened 2019-05-30
split out provider requirements

tomkralidis:


tomkralidis:
Oops! [Wrong project](https://github.com/geopython/pycsw/blob/master/requirements-standalone.txt) :) Fixed

Issue #142, opened 2019-05-27
Make better use of screen real estate in Collections HTML page


justb4:
The `Collections` HTML page is sort of squeezed-in, not using available browser-screen optimally. Best to show this with a picture, for example when opening
https://demo.pygeoapi.io/master/collections?f=html
we see:

![image](https://user-images.githubusercontent.com/582630/58421450-41da9200-8090-11e9-8315-b1f68827adab.png)

I.e.

* vertically: a scroll-bar which IMO is not required
* horizontally: the table could be broader 

Seen on Google Chrome and Safari Mac OSX.

justb4:
I can take this one up, PR upcoming. Below after some CSS and other mods:

![image](https://user-images.githubusercontent.com/582630/58707775-220bdc80-83b6-11e9-8577-5737ebb11777.png)

* excess text removed
* no vertical scrollbar table
* keep footer positioned at bottom

justb4:
Should be fixed.

tomkralidis:
The fixed footer results in some pages' content not being fully visible.  Example: https://demo.pygeoapi.io/master

However pages like https://demo.pygeoapi.io/stable/collections?f=html now render much better.

Can we not have the fixed footer?  Pull request for review in #158

justb4:
Think this issue is fixed and can be closed, also it is more about formatting than content as in #361.

PR #158, opened 2019-06-01
unfix footer (#142)

tomkralidis:


justb4:
I struggled quite some time to get the footer "right":

* `<footer>` element from `mini.css` was not positioned ok
* explicit `div` + CSS as now looked better but still not perfect

I tested leaving out `position: fixed` as suggested here gave garbled result for `collections.html` page:

image

Tried some other variants but no good results. But then: the nesting of various CSS `<div> row/col` and `<table>` elements from `mini.css` within `pygeoapi` templates needs review and could be causing this.

MichelGabriel:
Hi,

@pvgenuchten asked me to have a quick look at the code and I noticed that some of the problems are caused by the `max-height` in this style:
```css
table:not(.horizontal) {
    overflow: auto;
    max-height: 400px;
}
```
Although a bigger table doesn't 'listen' to this max-height, the container it is in does. So as a result the `<div class="container">` and `<main>` are also 400px high and the footer is positioned directly after these and therefore 'overwriting' the table.

tomkralidis:
Thanks @MichelGabriel .  What would be a suggested way forward on fixing?

MichelGabriel:
I changed some small things to make it work. Basically I made the footer fixed again by adding the class `sticky`. This PR was just removing that (sorry).

And I added some styles to `default.css`:
```css
html, body {
    background-color: #fff;
}
footer.sticky {
    position: fixed;
    width: 100%;
}
main {
    padding-bottom: 65px; /* prevent from falling under the footer */
}
table:not(.horizontal) {
    max-height: none;
}
```
And I removed the style `overflow:visible` from the table in `collections.html`. The table already had the style `overflow:auto`.

MichelGabriel:
But if you don't want a sticky footer I can have another look.

tomkralidis:
No problems removing a sticky footer.  Are you able to add to this PR (or feel free to open a new one)?

tomkralidis:
@MichelGabriel sorry I meant to keep the sticky footer.  I've updated this PR to implement your comments. I think this works now?  @justb4 can you check as well?

justb4:
Tested the branch locally: looks good now!

justb4:
Thanks @tomkralidis and @MichelGabriel !

justb4:
You can see the result at:
https://demo.pygeoapi.io/master/collections?f=html
(had to shift-reload for new CSS).
vs the 0.6.0 (quite old now) version:
https://demo.pygeoapi.io/stable/collections?f=html

PR #161, opened 2019-06-03
fix: avoid repeating self links in collections response

### LINKED ISSUE #160: collections repeats self links
collections?f=json repeats the self links at the end multiple times. If you have 5 collection items, you'll get 5 JSON self links, and 5 HTML self links:

```
"links": [{"type": "application/json", "rel": "self", "title": "This document as JSON", "href": "http://192.168.40.22:5000/collections?f=json"}, {"type": "text/html", "rel": "alternate", "title": "This document as HTML", "href": "http://192.168.40.22:5000/collections?f=html"}, {"type": "application/json", "rel": "self", "title": "This document as JSON", "href": "http://192.168.40.22:5000/collections?f=json"}, {"type": "text/html", "rel": "alternate", "title": "This document as HTML", "href": "http://192.168.40.22:5000/collections?f=html"}, {"type": "application/json", "rel": "self", "title": "This document as JSON", "href": "http://192.168.40.22:5000/collections?f=json"}, {"type": "text/html", "rel": "alternate", "title": "This document as HTML", "href": "http://192.168.40.22:5000/collections?f=html"}, {"type": "application/json", "rel": "self", "title": "This document as JSON", "href": "http://192.168.40.22:5000/collections?f=json"}, {"type": "text/html", "rel": "alternate", "title": "This document as HTML", "href": "http://192.168.40.22:5000/collections?f=html"}, {"type": "application/json", "rel": "self", "title": "This document as JSON", "href": "http://192.168.40.22:5000/collections?f=json"}, {"type": "text/html", "rel": "alternate", "title": "This document as HTML", "href": "http://192.168.40.22:5000/collections?f=html"}]}
```

bradh:
Resolves #160.

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -310,20 +310,20 @@ class API(object):
 
             fcm['collections'].append(collection)
 
-            fcm['links'].append({
-                'type': 'application/json',
-                'rel': 'self',
-                'title': 'This document as JSON',
-                'href': '{}/collections?f=json'.format(
-                    self.config['server']['url'])
-            })
-            fcm['links'].append({
-                'type': 'text/html',
-                'rel': 'alternate',
-                'title': 'This document as HTML',
-                'href': '{}/collections?f=html'.format(
-                    self.config['server']['url'])
-            })
+        fcm['links'].append({
+            'type': 'application/json',
+            'rel': 'self',
+            'title': 'This document as JSON',
+            'href': '{}/collections?f=json'.format(
+                self.config['server']['url'])
+        })
+        fcm['links'].append({
+            'type': 'text/html',
+            'rel': 'alternate',
+            'title': 'This document as HTML',
+            'href': '{}/collections?f=html'.format(
+                self.config['server']['url'])
+        })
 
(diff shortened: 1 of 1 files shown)


bradh:
Bad fix.

justb4:
I see one api test failing, probably counting the (wrong) number of links, maybe it is a matter fixing the test. Think removing adding the global links section out of the for-loop is a good direction. You can still re-open, add commits, we'll squash them into single during merge.

bradh:
If the number had reduced, I would have been happy to just adjust the test. However it adds duplicate entries (links goes from 6 -> 8), so I'm missing something on the implementation side. I'll work it out and reopen when I have it sorted.

justb4:
Ok, think in `def test_describe_collections(config, api_):` on line 148 len should be 2. A bit misleading as there is only one collection in config ('obs') thus the test is wrong:

Should be something like:

    assert len(collections['collections']) == 2
    assert len(collections['links']) == 2 # tests your fix

and on line 163: 

      assert len(collection['links']) == 8 # think is right

PR #163, opened 2019-06-03
fixes #134 render item attr image URL as image thumb element

### LINKED ISSUE #134: display link or image if feature contains link

justb4:
* started a `macros.jinja2` for possibly more rendering specifics
* optionally render image thumb if URL ends with image extension


justb4:
@pvgenuchten good idea, only something like:

            <div style="background-image: url({{ val }}); border: 1px solid black;"> </div>

does not play nice with the `<table>` (CSS). Will take quite some CSS-fiddling. Shall we just merge the above, after you can then improve?

pvgenuchten:
+1 to merge now, and improve later

PR #164, opened 2019-06-03
Adding contribution guidelines (#162)

kalxas:

PR #165, opened 2019-06-04
address OSGeo Incubation Committee comments

tomkralidis:
Addresses comments in https://lists.osgeo.org/pipermail/pygeoapi/2019-June/000009.html (note contribution guidelines are covered in #164).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,34 @@
+# =================================================================
+#
+# Authors: Tom Kralidis <tomkralidis@gmail.com>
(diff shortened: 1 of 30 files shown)


jodygarnett:
Just a quick comment, my email only covered spot checks (the application process is an excuse to talk to a team about this sort of thing and ask them to check their files and sample data to ensure they know where it came from).

kalxas:
@jodygarnett I think we addressed the issues in the e-mail thread. Can we move forward with a motion?

PR #187, opened 2019-06-19
More content in Presentation for Spec and pygeoapi Chapters

justb4:
Mainly in those two chapters:

* notice name change WFS3 to "OGC APIs - Features"
* more pictures, e.g. pygeoapi architecture

Preview url of this version at [raw cdn](https://rawcdn.githack.com/justb4/pygeoapi/0f6e82a6ec2a5bc22609ce2b33615e6cea3bc6da/presentation/index.html)


presentation/index.html
```diff
@@ -203,6 +206,7 @@ <h3>WFS 3.0 Hackathon</h3>
 					<li>Develop WFS 3.0 based on Swagger/OpenAPI (v3) </li>
 					<li>Write specification using GitHub</li>
 					<li>First implementations</li>
+					<li><strong>pygeoapi was born!</strong></li>
```
tomkralidis: First commit was 14 February 2018, perhaps we can articulate?

presentation/index.html
```diff
@@ -358,12 +363,17 @@ <h2>VI. pygeoapi</h2>
 				<section>
 					<!--  First slide -->
 					<h2>pygeoapi</h2>
-					<img src="static/geopython.png">
-					<p>Part of the amazing Geopython community in GitHub</p>
+					<img src="static/geopython.png" alt="">
```
tomkralidis: should we provide text instead of the empty `alt` ?

presentation/index.html
```diff
@@ -358,12 +363,17 @@ <h2>VI. pygeoapi</h2>
...
+					<p>Part of the amazing GeoPython community in GitHub</p>
```
tomkralidis: I think we use `Geopython` as a rule?

presentation/index.html
```diff
@@ -372,7 +382,7 @@ <h3>pygeoapi</h3>
 						<li>OSGeo Community Project (in Motion) </li>
 						<li>Powered by:</li>
 					</ul>
-					<img src="static/flask.png" style="background: white; border-radius: 10px"/>
+					<img src="static/flask.png" alt="" style="background: white; border-radius: 10px"/>
```
tomkralidis: here too, should we provide text instead of the empty `alt` ?

justb4:
> Nice work here, see some comments. In addition, for the architecture diagram we should articulate something like:
> 
> providers <-> core (i.e. `pygeoapi.api`) <-> Flask/Django / WSGI <-> HTTP
> 
> The key point being that while we provide a flask app the core allows for any Python microframework to be built downstream

Ha, that was the image I had earlier, but for simplicity left Core/API class out, yes makes sense as to not be tied to Flask.

justb4:
Ok, made edits, in particular new arch image. [New raw CDN link](https://rawcdn.githack.com/justb4/pygeoapi/d5067b70e44377de80c5049d57c73be5a2abbc3c/presentation/index.html).

PR #190, opened 2019-06-20
Update demo server example URL

jodygarnett:


justb4:
thanks @jodygarnett !

PR #194, opened 2019-06-22
Create CODE_OF_CONDUCT.md

tomkralidis:


tomkralidis:
Note I set the pygeoapi mailing list as the email for enforcement (https://github.com/geopython/pygeoapi/pull/194/files#diff-a1ee87dafebc22cbd96979f1b2b7e837R57) if this works for everyone.

PR #196, opened 2019-06-23
add canonical link to each html representation

### LINKED ISSUE #150: html encoding should provide canonical url for collections to help search engines
I suggest to add a [link tag with canonical url](https://www.metatags.org/rel_canonical) to root and each collection & item page

```
<link rel="canonical" href="www.yourdomainnam.com/collections/123/items/type987" />
```

search engines use this property to identify the most relevant url of an object

pvgenuchten:
fixes #150 

todo: make sure all methods return max (and at least) 1 "self" link of type "text/html"

toverify: for /items, if a filter is applied, does it make sense to use /items as canonical, or should it include the filter


pvgenuchten:
idea here is that /collections/obs and /collections/obs?f=html share the same canonical url, so machines considers both url's equal.

PR #207, opened 2019-08-02
Issue 43

jorgejesus:
- Code and text for read the docs
- Alterations on travis to better support read the docs e.g ` PYGEOAPI_CONFIG="pwd/pygeoapi-config.yml"`
- new requirements for RTD on its own folder
- Update some code files to have better description


.travis.yml
```diff
@@ -46,10 +47,12 @@ before_script:
   - gunzip < tests/data/hotosm_bdi_waterways.sql.gz | psql -U postgres test
 
 script:
+  - make -C ./docs html
```
tomkralidis: Can we build the docs after pytest and before the flake8'ing?

requirements.txt
```diff
@@ -1,4 +1,5 @@
 click
 Flask
+flask_cors
```
tomkralidis: Why is flask_cors now a required dependency?

pygeoapi/process/base.py
```diff
@@ -33,7 +33,7 @@
 
 
 class BaseProcessor(object):
-    """generic Processor ABC"""
+    """generic Processor ABC. Processes are inherid from this class"""
```
tomkralidis: *inherited

pygeoapi/log.py
```diff
@@ -27,6 +27,8 @@
 #
 # =================================================================
 
+""" Logging system """
```
tomkralidis: `"""Logging system"""`

pygeoapi/flask_app.py
```diff
@@ -144,6 +179,12 @@ def describe_processes(name=None):
 
 @APP.route('/processes/<name>/jobs', methods=['GET', 'POST'])
 def execute_process(name=None):
+    """
+    OGC open api jobs from processes access point (experimental)
+
```
tomkralidis: `:param name: identifier of process to execute`

pygeoapi/flask_app.py
```diff
@@ -162,7 +203,10 @@ def execute_process(name=None):
 @click.pass_context
 @click.option('--debug', '-d', default=False, is_flag=True, help='debug')
 def serve(ctx, debug=False):
-    """Serve pygeoapi via Flask"""
+    """
+    Serve pygeoapi via Flask. Runs pygeoapi
+    as a flask server. Not recommend for production.
```
tomkralidis:
```
:param debug: `bool` of whether to run in debug mode

:returns void
```

pygeoapi/flask_app.py
```diff
@@ -131,6 +161,11 @@ def dataset(feature_collection, feature=None):
 @APP.route('/processes')
 @APP.route('/processes/<name>')
 def describe_processes(name=None):
+    """
+    OGC open api processes access point (experimental)
+
```
tomkralidis: `:param name: identifier of process to describe`

pygeoapi/flask_app.py
```diff
@@ -131,6 +161,11 @@ def dataset(feature_collection, feature=None):
...
+    :returns: HTTP response
```
tomkralidis: all `:returns:` should describe the data returned (`dict` of ..., `list` of ...), etc.

pygeoapi/api.py
```diff
@@ -835,7 +840,7 @@ def to_json(dict_):
     """
     serialize dict to json
 
-    :param dict_: dict_
+    :param dict_: dict
```
tomkralidis:
```python
:param dict_: `dict` of JSON representation
```

pygeoapi/api.py
```diff
@@ -45,11 +48,13 @@
 TEMPLATES = '{}{}templates'.format(os.path.dirname(
     os.path.realpath(__file__)), os.sep)
 
+#: Return headers for requests (e.g:X-Powered-By)
```
tomkralidis: is `#:` intended (as opposed to `#`) ?

docs/requirements.txt
```diff
@@ -0,0 +1,2 @@
+sphinx_rtd_theme
```
tomkralidis: Should we put in `/requirements-dev.txt` instead?

requirements.txt
justb4: Think this is very useful for many client-apps, though in practice CORS would be handled by a front-end proxy like Apache, nginx, Traefik. NB shouldn't the name be `flask-cors`? (no underscore). It is already implemented [see here](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/flask_app.py#L52).

.travis.yml
justb4: Suggestion: more "bash-like" would be: `PYGEOAPI_CONFIG=$(pwd)/pygeoapi-config.yml`

requirements.txt
tomkralidis: This complicates the debian install which now needs to provide python-flask-cors (or python3-flask-cors?) given `setup.py` is tied to `requirements.txt`.

The implementation in https://github.com/geopython/pygeoapi/blob/master/pygeoapi/flask_app.py#L52 is conditional, so if a user specifies cors in pygeoapi config, and the install doesn't have it, then it will fail on startup and a simple `pip install flask-cors` will do the trick.  I think this is a cleaner and safer approach than promoting flask-cors to the core of the implementation requirements.

tomkralidis:
Yup we can have RTD setup against master and then release tags.

requirements.txt
justb4: Ok, no strong opinion, this is typically something to document, also that the Docker image supports it (via `requirements-dev.txt`).

jorgejesus: Some clarification, from the back of my head (will not put my hands on fire for it) we had the `PYGEOAPI_CONFIG="pwd/pygeoapi-config.yml` with `cors:true` and I think that was the case when I was initially working with RTD. 

And this caused a problem when loading pygeoapi on the RTD platform. Now I see that the coors has been deactivated on the configuration file. So I will removed it from the dependencies

jorgejesus: Put Flask_cors back to requirements.txt as it was before

.travis.yml
jorgejesus: Changes done

pygeoapi/process/base.py
jorgejesus: done

pygeoapi/log.py
jorgejesus: done

pygeoapi/flask_app.py
jorgejesus: Fixed

pygeoapi/flask_app.py
jorgejesus: Done

pygeoapi/flask_app.py
jorgejesus: done

pygeoapi/api.py
jorgejesus: done

pygeoapi/api.py
jorgejesus: The `:#` is necessary to indicate RTD that it should that string to describe the content below in this case the headers (it needs to stay)

requirements.txt
tomkralidis: Should we keep flask_cors in `requirements-dev.txt`?

pygeoapi/plugin.py
```diff
@@ -26,12 +26,15 @@
...
 # =================================================================
+"""Plugin loader"""
 
 import importlib
 import logging
 
 LOGGER = logging.getLogger(__name__)
 
+#: Loads provider plugins to be used by pygeoapi,\
+#: formatters and processes avaiable
```
tomkralidis: avaiable

pygeoapi/flask_app.py
```diff
@@ -144,6 +181,13 @@ def describe_processes(name=None):
 
 @APP.route('/processes/<name>/jobs', methods=['GET', 'POST'])
 def execute_process(name=None):
+    """
+    OGC open api jobs from processes access point (experimental)
+
+    :param name: identifier of process to execute
+    :returns: HTTP response
```
tomkralidis: should be `:returns: `dict` of response, or something similar.  In other words, provide the Python type of the response (should be done for all functions).

docs/source/code.rst
```diff
@@ -0,0 +1,219 @@
...
+
+CSV provider
+^^^^^^^^^^^^
+
+.. automodule:: pygeoapi.provider.csv_
+   :show-inheritance:
+   :members:
+   :private-members:
+   
+Elastic Search provider
```
tomkralidis: Elasticsearch provider

justb4:
Where are we now? Has the flask-cors point resolved. Think we should merge then and fix document-related content from then+have dedicated url like docs.pygeoapi.io . How is this published to RTD? Whose account?

tomkralidis:
@justb4 we can probably merge once @jorgejesus looks at my latest review.  flask-cors still needs to be removed from `requirements.txt`

pygeoapi/flask_app.py
jorgejesus: Can we keep it as it is, since I am not certain of what is the nature of the flask HTTP responde ?

jorgejesus:
@justb4 @tomkralidis  Mijn excuses for Flask_cors sill on the `requirements.txt` it has been fixed and implemented the spelling corrections. I have not implemented the `HTTP response` since I it is not clear in my head what is the nature of the object

PR #217, opened 2019-08-26
update presentation for FOSS4G 2019

tomkralidis:
- moved existing presentation to `presentations/2018`
- default presentation now `presentations/default`


justb4:
great work!
minor comment: the numbering in top-URL and bottom-right does not align, see screenshot:

image

tomkralidis:
Ref: https://github.com/hakimel/reveal.js/issues/2067

I've since removed slide numbering to the URL stays true.

PR #220, opened 2019-08-29
Proposal for enabling property filtering with ogr provider

ynte:
Being able to filter on the properties of ogr data sources is very important for me. So i've developed the following code to add this ability to the provider in question. I can improve upon this pull request and add tests if this feature is something you guys would consider adding.

I did wonder why `get_fields` was not used yet, perhaps there was a problem that prevented the loading of fields?

Using this project to consume legacy api's and expose them through a modern easy to implement protocol sounds very appealing to me, so i would love to help :+1:

### DIFF
--- a/pygeoapi/provider/ogr.py
+++ b/pygeoapi/provider/ogr.py
@@ -176,6 +176,9 @@ class OGRProvider(BaseProvider):
         self.driver = None
         self.conn = None
 
+        LOGGER.debug('Grabbing field information')
+        self.fields = self.get_fields()
+    
     def _list_open_options(self):
         return [
             f"{key}={str(value)}" for key, value in self.open_options.items()]
@@ -284,6 +287,20 @@ class OGRProvider(BaseProvider):
 
                 # layer.SetSpatialFilterRect(
                 # float(minx), float(miny), float(maxx), float(maxy))
+
+            if properties:
+                LOGGER.debug('processing properties')
+
+                attribute_filter = ' and '.join(
+                    map(
+                        lambda x: '{} = \'{}\''.format(x[0], x[1]),
+                        properties
+                    )
+                )
+
+                LOGGER.debug(attribute_filter)
+
+                layer.SetAttributeFilter(attribute_filter)
 
             # Make response based on resulttype specified
             if resulttype == 'hits':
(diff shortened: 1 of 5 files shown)


justb4:
Ok, thanks @ynte ! Yes, a very useful feature, give me some time to review (now travelling back from FOSS4G). Yes, preferably add unit tests and try to test with different OGR backends (services like WFS v1/2, GeoJSON etc). `GetFields`, maybe not all OGR backends gave correct results? Have to look into. Property filtering may also be a very expensive operation for big e.g. DB tables where the properties are not indexed...

ynte:
I added tests in the currently defined ogr suites (gpkg, sqlite, wfs, shapefile). I could try to find an ogr format that does not support `SetAttributeFilter`, I think the AttributeFilter would simply be ignored in that case.

It's true property filtering might be very expensive. Perhaps the fields that are available for filtering could be constrained in the configuration of the dataset.

justb4:
Hmm, new insights, see [my comment here, last sentences](https://github.com/geopython/pygeoapi/pull/283#discussion_r337930549). @ynte "your `get_fields()`" made sense after all in your context, but we should solve this in core:

* attempting to populate `self.fields` for any Provider
* checks for allowed `properties` ([already implemented](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/api.py#L574))

ynte:
That is how I remembered it. I wanted to double check it before answering, but hadn't found any time for that yet.

I can imagine the current situation isn't perfect yet. I think an ability to confine the fields that can be filtered on would be very useful for instance. That way admins can prevent filtering on non-indexed fields. But as I understand from your comments that sort of functionality does not belong to the data sources themselves.

ynte:
I tried to merge the upstream master, but I think i have failed 😅. I'll have to come back to this later unfortunately.

ynte:
I'm sorry for the mess :stuck_out_tongue_closed_eyes: , the PR should be clean and up to date now.

justb4:
@ynte no sweat! We're all busy bees, and always happy with contributions! Yes, I tripped in the `flake8` error (one space on empty line is enough) many a time!

ynte:
In reference to #398: what was blocking the inclusion of this PR?

justb4:
> In reference to #398: what was blocking the inclusion of this PR?

Not sure, maybe the API was not ready, concerns with large datasets like WFS, or I overlooked something. I tested just now locally and merged locally without conflicts and tests run ok! See [Gitter](https://gitter.im/geopython/pygeoapi?at=5e9452afcc370f0b07df2ce1). So will merge, thanks for patience @ynte .

PR #223, opened 2019-09-03
Add support for Starlette

### LINKED ISSUE #222: Support starlette
I want to support the no-blocking async/await pattern with asyncio 

Add a new server type capability for [starlette.io](https://www.starlette.io)

Other asyncio implementations can be considered from some benchmarks [here](https://github.com/the-benchmarker/web-frameworks) and [here](https://www.techempower.com/benchmarks/#section=data-r18&hw=ph&test=fortune&l=zijzen-1) but also `starlette` has a good reputation and some nice features like the support of `GraphQL`, `WebSockets`, `API Schema` generator,  `Background Tasks`, HTTP/2 and HTTP/3 server push, etc

francbartoli:
closes #222


francbartoli:
> Checked and run the starlette server and everything seems ok. I find that using ASGI is important.
> 
> Only problem is that we have new dependencies on `requirements.txt` and from previous discussions the base requirements should be keept minimal, maybe passing the startlette/uvloop to `requirements-dev.txt` and raise a warning when they are not avaialble for the server the run on the command line ???? @tomkralidis what do you think ??

One option might be to handle starlette deps as separate requirements file and also as a plugin require in `setup.py`, something like:

`pip install pygeoapi[starlette] `

At the moment the setup file has not been touched. Also, I'd expect to deploy it in production with a different built from the default dockerfile. I would come up with a new PR for the docker stuff.

pygeoapi/__init__.py
```diff
@@ -30,15 +30,30 @@
 __version__ = '0.6.0'
 
 import click
-from pygeoapi.flask_app import serve
+from pygeoapi.flask_app import serve as serve_flask
```
tomkralidis: Can we move these to:

```python
    if server == 'flask':
        from pygeoapi.flask_app import serve as serve_flask
        ctx.forward(serve_flask)
        ctx.invoke(serve_flask)
    elif server == 'starlette':
        from pygeoapi.starlette_app import serve as serve_starlette
        ctx.forward(serve_starlette)
        ctx.invoke(serve_starlette)
    else:
        raise click.ClickException('--flask/--starlette is required')

Comment (francbartoli) [reply pygeoapi/__init__.py]: +1

File: pygeoapi/__init__.py [hunk 2]
```diff
@@ -30,15 +30,30 @@
...
-@click.group()
-@click.version_option(version=__version__)
-def cli():
-    pass
+cli = click.Group()
+cli.version = __version__
+
+
+@cli.command()
+@click.option('--flask', 'server', flag_value="flask", default=True)
```
tomkralidis: does `default=True` mean that the `server` param is `flask` by default ?

pygeoapi/flask_app.py
```diff
@@ -205,7 +205,7 @@ def execute_process(name=None):
 @click.command()
 @click.pass_context
 @click.option('--debug', '-d', default=False, is_flag=True, help='debug')
-def serve(ctx, debug=False):
+def serve(ctx, server, debug=False):
```
tomkralidis: does passing `server` here do anything?  Should we document what it is in the docstring?

docs/source/install.rst
```diff
@@ -44,9 +44,9 @@ It is advisable to run pygeoapi inside an isolated enviroment either *virtualenv
 	# generate OpenAPI Document
 	pygeoapi generate-openapi-document -c local.config.yml > openapi.yml
 	export PYGEOAPI_OPENAPI=/path/to/openapi.yml
-	
+
 	#Run pygeoapi
-	pygeoapi serve
+	pygeoapi serve --flask
```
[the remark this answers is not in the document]
francbartoli: Right that is true

pygeoapi/__init__.py
francbartoli: Yes it does, see answer above

tomkralidis:
@francbartoli it looks like deploying via starlette results in 404s for the `static` assets (`pygeoapi/static`).  Is there a way to include static file serving / setup here for the starlette front end?

PR #224, opened 2019-09-04
move address into contact details

tomkralidis:
Moves address into contact section of HTML page


tomkralidis:
@pvgenuchten note the Travis errors are due to a badly behaving WFS and unrelated to this PR.

pvgenuchten:
hi @tomkralidis seems this PR broke the schema.org annotations

an item of type address should be under organisation, not within contactpoint

image

image

tomkralidis:
@pvgenuchten good catch.  PR to fix in #251

Issue #213, opened 2019-08-23
collection detail page has wrong link in schema.org/includedInDataCatalog


pvgenuchten:
It should reference the pygeoapi main page (or collections page)

![image](https://user-images.githubusercontent.com/299829/63591729-e432e480-c5af-11e9-9ea7-aa08c6898597.png)

jorgejesus:
Fixed #210  also looking at this
The collections page is considered a catalog:
 ```<section id="collections" itemscope itemtype="//schema.org/DataCatalog">```

now collection detail page points to the collection page (includedInDataCatalog)

@pvgenuchten Shouldnt the collection page also have a "includedInDataCatalog" to the pygeopai main page ??

PR #225, opened 2019-09-06
Issue 210 missing  HTTPS for schema.org

### LINKED ISSUE #210: Schema.org links should use HTTPS
**Description**
On template folder files we have schema.org annotations that use links with HTTP. If we run pygeoapi on HTTPS the browser will report mixed content. In case we run pygeopai as HTTP it will not be a problem to use links with HTTPS

**Steps to Reproduce**
For examples: Look at `templates/items.html` file an find:
 `itemtype="http://schema.org/Dataset`

**Expected behavior**
it should be: `itemtype="https://schema.org/Dataset`
**Screenshots/Tracebacks**

**Environment**
 - OS: All
 - Python version: 3.7.4 
 - pygeoapi version:  master - 0.6.0

**Additional context**

jorgejesus:
closes #210 and #213  

It was implemmented a generic schema (//) on schema.org links. Code was then tested using a pygeoapi/Gunicorn using HTTPS local certificates (https://github.com/FiloSottile/mkcert)

Browser didn't reported any mixed content problem

I requesting for @pvgenuchten  to comment on ticket #213  and check if there is changes needed

### DIFF
--- a/pygeoapi/templates/collection.html
+++ b/pygeoapi/templates/collection.html
@@ -5,8 +5,8 @@
 / <a href="./{{ data['name'] }}">{{ data['title'] }}</a>
 {% endblock %}
 {% block body %}
-    <section id="collection" itemscope itemtype="//schema.org/Dataset">
-      <span itemprop="includedInDataCatalog" itemscope itemtype="//schema.org/DataCatalog">
+    <section id="collection" itemscope itemtype="https://schema.org/Dataset">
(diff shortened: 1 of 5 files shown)


tomkralidis:
Should we use `https://` ?  From https://www.paulirish.com/2010/the-protocol-relative-url/

`
Now that SSL is encouraged for everyone and doesn’t have performance concerns, this technique is now an anti-pattern. If the asset you need is available on SSL, then always use the https:// asset. 
`

PR #228, opened 2019-09-07
fix starlette static file support (#222)

tomkralidis:
Attempts to address https://github.com/geopython/pygeoapi/pull/223#issuecomment-529053032

There may be better ways to accomplish this functionality, which @francbartoli would know better :)


tomkralidis:
@francbartoli sounds good/agree.  We can cover this in #229 as part of a larger testing update. Does this make sense?

PR #232, opened 2019-09-09
replace title for logo in html encoding

pvgenuchten:
- add logo to html encoding, replacing title (or do we need that at each page, on home it was duplicated)
- add logo to footer
- add 2 images (1 as logo, to be replaced, 1 as pygeoapi attribution)

image


pvgenuchten:
ai, the logo ended up really small, i must have pushed wrong image, somehow it has a huge top-bottom margin

image

tomkralidis:
@pvgenuchten can you update based on latest master.  icons are in `static/img` now

PR #234, opened 2019-09-12
fix GeoJSON response for PostgreSQL, SQLite and GeoPackage (#95)

### LINKED ISSUE #95: Several Providers return FeatureCollection on get single Feature
This applies to the Providers `GeoPackage`,  `PostgreSQL` and `SQLite`.

The WFS3 operation to request a single Feature, like e.g. http://localhost:5000/collections/obs/items/371?f=json 
should return a single Feature GeoJSON construct like: 

    {"type": "Feature", 
     "ID": "371", 
     "geometry": {"type": "Point", "coordinates": [-75.0, 45.0]}, 
     "properties": {
         "stn_id": "35", 
         "datetime": "2001-10-30T14:24:55Z", "value": "89.9"
            .
       }, 
      "links": [...
     }

Instead these Providers return a GeoJSON FeatureCollection with a single Feature.
Other providers like CSV, ES and GeoJSON return a single Feature as expected.

Applies to Provider implementations: `geopackage.py`, `postgresql.py` and `sqlite.py` and their tests.

tomkralidis:
cc @kalxas 

The PostgreSQL provider always provides a full FeatureCollection `dict` (via `get` and `query`).

The `query` and `get` methods should return per the abstract base class docs:

https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/base.py#L63

https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/base.py#L72

This PR attempts to fix #95

### DIFF
--- a/pygeoapi/provider/geopackage.py
+++ b/pygeoapi/provider/geopackage.py
@@ -1,8 +1,10 @@
 # =================================================================
 #
 # Authors: Jorge Samuel Mendes de Jesus <jorge.dejesus@geocat.net>
+#          Tom Kralidis <tomkralidis@gmail.com>
 #
 # Copyright (c) 2018 Jorge Samuel Mendes de Jesus
+# Copyright (c) 2019 Tom Kralidis
 #
 # Permission is hereby granted, free of charge, to any person
 # obtaining a copy of this software and associated documentation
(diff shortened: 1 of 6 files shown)


justb4:
issue concerns several Providers, see #95

PR #243, opened 2019-09-21
add operationId to OpenAPI

tomkralidis:


pygeoapi/openapi.py
```diff
@@ -270,6 +274,7 @@ def get_oas_30(cfg):
 
         paths[collection_name_path] = {
             'get': {
+                'operationId': 'collections_{}'.format(k),
```
francbartoli: More appropriate here a singular form

pygeoapi/openapi.py
```diff
@@ -292,6 +297,7 @@ def get_oas_30(cfg):
 
         paths['{}/items'.format(collection_name_path)] = {
             'get': {
+                'operationId': 'collections_{}_items'.format(k),
```
francbartoli: same above with the singular form

francbartoli:
To add more context in the meantime, let's see what are the pros:

From the openapi spec for [operations](https://swagger.io/docs/specification/paths-and-operations/) `operationId` would be relevant for client code generation and the traverse of `links`. [Link object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#link-object) seems to be powerful but I don't see such advanced use from OGC yet.

francbartoli:
@tomkralidis after some internal discussions I can say this might be considered optional so if we don't want to support client code generation then we can close it for now

PR #270, opened 2019-10-05
Fix service-doc and service-desc inversion, and update media type of openapi+json

rouault:


justb4:
@rouault good to see you contributing! FYI: the Travis build failed because of a Python formatting error. `flake8` is quite strict here:

```
./pygeoapi/api.py:211:80: E501 line too long (81 > 79 characters)
```

This also hits us now and then :-(. Best is to install flake8 and run it from topdir before commits/push.

rouault:
>  If you only fix the Python formatting (line length), and have successful Travis build we can merge.

just force pushed with a fix

justb4:
Ok, thanks @rouault !

tomkralidis:
Thanks @rouault !

PR #271, opened 2019-10-05
api: fix to report an array of bbox in collection description

rouault:
The latest version of the OAPI-F spec mandates the bbox to be
advertized as an array of bbox.
This commits accept the service configuration to have just
a single bbox or an array of bbox, and adapt accordingly to
generate the collection description.

### DIFF
--- a/tests/pygeoapi-test-config.yml
+++ b/tests/pygeoapi-test-config.yml
@@ -83,7 +83,7 @@ metadata:
 datasets:
     obs:
         title: Observations
-        description: Observations
+        description: Observations_description
         keywords:
             - observations
             - monitoring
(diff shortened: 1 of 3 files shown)


tests/pygeoapi-test-config.yml
```diff
@@ -83,7 +83,7 @@ metadata:
 datasets:
     obs:
         title: Observations
-        description: Observations
+        description: Observations_description
```
tomkralidis: Does this update fix any breakage?

rouault: no, unrelated to the purpose of this commit. This is just that it better validates that the content of the 'description' of the config goes to 'description' of the output

tomkralidis:
Thanks @rouault

PR #272, opened 2019-10-06
api: fix paging links

rouault:
- Add minimal validation of startindex and limit parameters
- Include limit parameter coming from initial request in next and prevl links
- Only include prev link if startindex > 0
- Only include next link if we got as many features at the limit

Note: there is a broken test for sortby that was unnoticed. I have marked
it as such.


tomkralidis:
Thanks @rouault

PR #274, opened 2019-10-11
Added serialization for Decimal to util.py

mbucknell:
In my testing of the postgresql provider, I have a table which contains "number" columns. The postgres provider converts this column to a `decimal.Decimal` which can not be serialized. I added the fix to util.py. The same Stack Overflow issue also references added an additional handler for decimal.Decimal.

I have not updated the tests. This would require adding a column to the test data which I would be happy to do.


tomkralidis:
thanks @mbucknell !

Issue #269, opened 2019-10-04
Issue with postgres driver schema configuration option and PostGIS functions


alpha-beta-soup:
**Description**
The PostgreSQL driver configuration currently has a `schema` property allowing you to add a schema to the search path (so your table can be found). However if you set this to anything but `public` (which is the default) then the search path seems to be *constrained to* that schema. This is a problem if your data is under schema `x` but the PostGIS functions (e.g. `ST_AsGeoJSON`) that are called in pygeoapi are still defined in the `public` schema. 

**Steps to Reproduce**
- Have data in a non-`public` schema, but leave the PostGIS functions in the default, `public` schema.
- Create a provider configuration like:

```yaml
       provider:
           name: PostgreSQL
           data:
               host: mycoolhost.co.nz
               dbname: my-db
               user: my-read-user
               password: my-read-user-pw
               port: 1234
               schema: x
           id_field: "@id"
           table: my-table
           context: query
```

- Try and request data

**Expected behavior**

There should be no error.

**Screenshots/Tracebacks**

```
[2019-10-04T01:20:45Z] {/pygeoapi/pygeoapi/provider/postgresql.py:221} ERROR - function st_asgeojson(public.geometry) does not exist
LINE 1: ... "hasSampleRelationship", "isFeatureOfInterestOf",ST_AsGeoJS...
                                                             ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

I can hotfix by appending `public` before the calls to `ST_AsGeoJSON`, e.g.

```python
sql_query = SQL("DECLARE \"geo_cursor\" CURSOR FOR \
 SELECT {0},public.ST_AsGeoJSON({1}) FROM {2}").\
    format(db.columns,
    Identifier('geom'),
    Identifier(self.table))
```

But that's not a proper fix since the PostGIS functions aren't constrained to be in that schema.

Another, slightly cleaner workaround is to declare the provider config like:

```yaml
       provider:
           name: PostgreSQL
           data:
               host: mycoolhost.co.nz
               dbname: my-db
               user: my-read-user
               password: my-read-user-pw
               port: 1234
               schema: public,x
           id_field: "@id"
           table: my-table
           context: query
```

(Note the  `schema: public,x`). This avoids the bug but I presume could have other issues further down the track.

jorgejesus:
[...]

PR #281, opened 2019-10-20
Fixes Issue 269

jorgejesus:
Based on the discussion on #269  Have implemented the following configuration on the postgresql driver:

```
        provider:
            name: PostgreSQL
            data:
                host: 127.0.0.1
                dbname: test
                user: postgres
                password: postgres
                port: 5432
                search_path: [osm,public]
            id_field: osm_id
            table: hotosm_bdi_waterways
            geom_field: foo_geom
```

There is no `schema` parameters since using the proposed search path `[x, public] ` will ensure that tables are found and functions in PostGIS are properly called

The dataset  `hotosm_bdi_waterways` has been updated to use a specific schema and defined geometry column. 

If geometry column is not defined it will be set to 'geom' and if no search path is set then it will be public

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,7 +111,3 @@ openapi.yml
 
 #Eclipse project
 .project
-docker-compose.yml
-.gitignore
-notes.txt
-pygeoapi-config-local.yml
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,7 @@ openapi.yml
 
 #Eclipse project
 .project
+docker-compose.yml
+.gitignore
+notes.txt
+pygeoapi-config-local.yml
--- a/pygeoapi/provider/postgresql.py
+++ b/pygeoapi/provider/postgresql.py
@@ -72,8 +72,11 @@ class DatabaseConnection(object):
              (defaults to UNIX socket if not provided)
             port – connection port number
              (defaults to 5432 if not provided)
-            schema – schema to use as search path, normally
-             data is in the public schema
+            search_path – search path to be used (by order) , normally
+             data is in the public schema, [public],
+             or in a specific schema ["osm", "public"].
+             Note: First we should have the schema
+             being used and then public
 
         :param table: table name containing the data. This variable is used to
                 assemble column information
@@ -87,18 +90,18 @@ class DatabaseConnection(object):
         self.context = context
         self.columns = None
         self.conn = None
-        self.schema = None
 
     def __enter__(self):
         try:
-            self.schema = self.conn_dic.pop('schema', None)
-            if self.schema == 'public' or self.schema is None:
+            self.search_path = self.conn_dic.pop('search_path', None)
+            if self.search_path is ['public'] or self.search_path is None:
                 pass
             else:
-                self.conn_dic["options"] = '-c search_path={}'.format(
-                    self.schema)
-                LOGGER.debug('Using schema {} as search path'.format(
-                    self.schema))
+                if "public" not in self.search_path:
+                    self.search_path = self.search_path + ["publc"]
+                self.search_path = ",".join(self.search_path)
+                self.conn_dic["options"] = f'-c search_path={self.search_path}'
+                LOGGER.debug(f'Using search path: {self.search_path} ')
             self.conn = psycopg2.connect(**self.conn_dic)
 
         except psycopg2.OperationalError:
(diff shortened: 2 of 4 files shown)


jorgejesus:
Also made a local test with the following configuration
```
        provider:
            name: PostgreSQL
            data:
                host: 127.0.0.1
                dbname: test
                user: postgres
                password: postgres
                port: 5432
            id_field: osm_id
            table: hotosm_bdi_waterways

```

pygeoapi/provider/postgresql.py
```diff
@@ -87,18 +90,18 @@ def __init__(self, conn_dic, table, context="query"):
...
         self.columns = None
         self.conn = None
-        self.schema = None
 
     def __enter__(self):
         try:
-            self.schema = self.conn_dic.pop('schema', None)
-            if self.schema == 'public' or self.schema is None:
+            self.search_path = self.conn_dic.pop('search_path', None)
+            if self.search_path is ['public'] or self.search_path is None:
```
alpha-beta-soup:
```
>>> ['public'] is ['public']
False
```

pygeoapi/provider/postgresql.py
```diff
@@ -87,18 +90,18 @@ def __init__(self, conn_dic, table, context="query"):
...
                 pass
             else:
-                self.conn_dic["options"] = '-c search_path={}'.format(
-                    self.schema)
-                LOGGER.debug('Using schema {} as search path'.format(
-                    self.schema))
+                if "public" not in self.search_path:
+                    self.search_path = self.search_path + ["publc"]
```
alpha-beta-soup: Typo: `publc`; and do we always want to add this schema? Why?

pygeoapi/provider/postgresql.py
```diff
@@ -149,6 +152,8 @@ def __init__(self, provider_def):
         self.table = provider_def['table']
         self.id_field = provider_def['id_field']
         self.conn_dic = provider_def['data']
+        self.geom = provider_def.get('geom_field') \
+            if provider_def.get('geom_field') else "geom"
```
alpha-beta-soup: Why not `self.geom = provider_def.get('geom_field', 'geom')` ?

pygeoapi/provider/postgresql.py
alpha-beta-soup: I think `['public']` should be the default if the property is completely unspecified, but I don't think it needs to be enforced if it's explicitly left out from the configuration.

.gitignore
```diff
@@ -111,3 +111,7 @@ openapi.yml
 
 #Eclipse project
 .project
+docker-compose.yml
```
tomkralidis: Are the `.gitignore` updates intentional? Files listed seem random-ish>

jorgejesus: Mijn fout it was not supposed to be pushed

pygeoapi/provider/postgresql.py
jorgejesus: @alpha-beta-soup  Thank you, now I know why this is wrong (http://net-informations.com/python/iq/is.htm)

pygeoapi/provider/postgresql.py
justb4: Agree (do not add `public` if missing in `search_path`) .

pygeoapi/provider/postgresql.py
justb4: Yes `is` and `==` can be tricky in Python 😈. Also I tend to never have `else` clauses, many refs on web like recently on [Medium](https://medium.com/better-programming/why-you-need-to-stop-using-else-statements-5b1fd09dea9e). Just a suggestion, halves number of lines, but consider this as syntactic sugar: 
```
search_path = self.conn_dic.pop('search_path', ['public'])
if search_path != ['public']:
    self.conn_dic["options"] = f'-c search_path={",".join(search_path)}'
    LOGGER.debug(f'Using explicit search path: {search_path}')
self.conn = psycopg2.connect(**self.conn_dic)
```
(IMO no need for `search_path` to be member var, in that case would also need initialization in `__init__()`))

jorgejesus: @justb4  Thank you for the fresh eyes, I have implemented your suggestions

pygeoapi/provider/postgresql.py
jorgejesus: @alpha-beta-soup  Implemented recommendation, thank you

.gitignore
jorgejesus: `.gitignore`has is now identical to master

jorgejesus:
@justb4  Well spotted !!! There is already the log information inside `__enter__` so better to remove the extra logging

Issue #279, opened 2019-10-17
Adding the ability to filter on properties in the postgres provider


mbucknell:
**Is your feature request related to a problem? Please describe.**
We are hoping to implement a OGC API service with a Postgres background and will need the ability to filter collections on property values in order to make it useful for our users. I saw the discussion in https://github.com/geopython/pygeoapi/issues/269 regarding this but wanted to call this out as a unique feature.

**Describe the solution you'd like**
At the very least, we should have the ability to query on all property values other than the id and the geom column. An additional features might be to specify the exact properties to use in the config file.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

mbucknell:
I plan on working on adding this feature and welcome any suggestions.

justb4:
Quick scan: is this also related to https://github.com/geopython/pygeoapi/pull/220 ?
Implementation strategy: this feature will be generic, but not all Providers will be able to handle property filtering natively. Similar to a re-projection capability, see [this discussion on Gitter](https://gitter.im/geopython/pygeoapi?at=5da5fe2b8e2e9a7c6b0299da).  

Not sure if standard already provides API convention, but IMHO we should apply a common strategy here as well: Provider should indicate if it can handle property filtering, could be a simple `boolean`.

mbucknell:
I believe it is related. There seems to be hooks in the providers for this but they aren't implemented at least in the case of postgres. I believe it is part of the standard: http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_parameters_for_filtering_on_feature_properties

mbucknell:
This PR contains an example implementation: https://github.com/geopython/pygeoapi/pull/283

tomkralidis:
Should we close this issue given #283 has been merged?

mbucknell:
https://github.com/geopython/pygeoapi/pull/283 implements this feature. Closing issue.

PR #283, opened 2019-10-21
Add ability to filter by properties to postgresql provider.

mbucknell:
This addresses this issue: https://github.com/geopython/pygeoapi/issues/279. I added the config variable allow_property_filtering to the dataset provider definition. If this is accepted, I would update the example config file with this value.


jorgejesus:
Tested locally and check the code functionality, everything seems to be ok :+1:

mbucknell:
Added my name. Do you think it would be good idea to update docker/default.config with the new config parameter? If so, I can do that too.

pygeoapi/api.py
```diff
@@ -569,10 +569,13 @@ def get_collection_items(self, headers, args, dataset, pathinfo=None):
...
-        LOGGER.debug('processing property parameters')
-        for k, v in args.items():
-            if k not in reserved_fieldnames and k in p.fields.keys():
-                properties.append((k, v))
+        if self.config['datasets'][dataset]['provider']\
+                .get('allow_property_filtering', False):
+            LOGGER.debug('processing property parameters if enabled')
+            for k, v in args.items():
+                if k not in reserved_fieldnames and k in p.fields.keys():
+                    LOGGER.debug('Add property filter {0}={1}'.format(k, v))
```
tomkralidis: For consistency, would prefer:     `LOGGER.debug('Add property filter {}={}'.format(k, v))`

pygeoapi/api.py
```diff
@@ -569,10 +569,13 @@ def get_collection_items(self, headers, args, dataset, pathinfo=None):
             LOGGER.error(exception)
             return headers_, 500, json.dumps(exception)
 
-        LOGGER.debug('processing property parameters')
-        for k, v in args.items():
-            if k not in reserved_fieldnames and k in p.fields.keys():
-                properties.append((k, v))
+        if self.config['datasets'][dataset]['provider']\
+                .get('allow_property_filtering', False):
```
tomkralidis: Why do we need `allow_property_filtering`?

mbucknell: I may have misinterpreted the comment in [Issue 270](https://github.com/geopython/pygeoapi/issues/279) from @justb4 

> Not sure if standard already provides API convention, but IMHO we should apply a common strategy here as well: Provider should indicate if it can handle property filtering, could be a simple boolean.

Personally, I would be okay removing it. I believe this feature is part of the OGC API specification

jorgejesus: @tomkralidis  Will give my 2cent, IMHO providers should have a way to control property filtering, since some providers may have problems with or it may consume to many resources.  Maybe  it would be better to implement things in a different way, assuming filtering by default and only if `disable_property_filtering` to disable it 

@justb4  What do you think ??

tomkralidis: Is the question whether a provider implements filtering or not?  I would think that the provider `.fields()` would relay this (advertising through an OpenAPI document), and keeping it empty essentially achieves the same thing?  Backends should then implement property as they wish (no support, or throw an exception, or process accordingly).  Thoughts?

jorgejesus: I will not comment on the OpenAPI since I am a bit unfamiliar,and yes the question is : "Whether a provider implements filtering or not" (at least on my mind)

justb4: My [comment in issue 279](https://github.com/geopython/pygeoapi/issues/279#issuecomment-543371254) expanded (AFAICR):

* a Provider may or may not provide Property filtering, mainly for technical reasons, e.g. a CSV or GeoJSON Provider has no native query/filtering function.
* in those cases the `pygeoapi` core can apply or help with explicit filtering while Provider is building a resultset.

My comment was more directed to avoid every Provider without native property filtering to implement their own. We could e.g. filter during result generation. Probably we don't need a boolean property even, only helper functions the Provider can call.

This was intended to be different from *allowing* property filtering.  This is IMHO more *data collection related*: e.g. some may not allow filtering on any property or only selected properties for very big datasets (millions) or for other reasons. This kind of stuff (complex Filters) has killed many a WFS v1/2, and scared off many users ("you put your database online"!). WFS2 stored queries were a safety net hardly used. This could be a general API capability from 'no filtering' to 'selected fields filtering' to 'any field filtering'. For now I am considering `key=value` filtering. I realize there is a whole range of other filters and -languages like CQL and spatial filters. The OpenAPI spec: https://swagger.io/docs/specification/describing-parameters/#query-parameters . 

(I see now that `pygeoapi` accepts properties as `key=value` in query params but these not seem to be passed to the Provider, unless Provider initializes `self.fields` explicitly. See 

* https://github.com/geopython/pygeoapi/blob/master/pygeoapi/api.py#L574 and
* https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/base.py#L52

looks like if Provider cannot populate `self.fields` no filtering can be applied. We should somehow improve here, but can be a nasty problem for Providers without metadata).

tomkralidis:
Thanks for the comments.  Thoughts:

- I'd prefer to leave filtering (if any) to the backend as opposed to pygeoapi by-hand filtering a result set to keep `pygeoapi.api` straightforward
- pygeoapi only passes `key=value` params to a given provider only if a provider contains said fields.  The idea here is that on a provider init, a provider should be able to identify its fields, else `.fields` is empty.  The point here is that perhaps `.fields` can accomplish the same thing as a specific config property?
- I guess we should also ask whether we should throw a 500 if a given backend cannot do filtering against a relevant request, or whether filtering is ignored and the provider returns everything anyway

mbucknell:
I just want to verify that I should remove the config parameter, allow_property_filtering? I'm getting a little lost in all of the discussion.

pygeoapi/api.py
```diff
@@ -569,9 +569,10 @@ def get_collection_items(self, headers, args, dataset, pathinfo=None):
             LOGGER.error(exception)
             return headers_, 500, json.dumps(exception)
 
-        LOGGER.debug('processing property parameters')
+        LOGGER.debug('processing property parameters if enabled')
```
tomkralidis: Can we revert this line?  Then +1 to merge

mbucknell: Done

geekdenz:
Hi, sorry a bit of a newbie to PyGeoAPI. From what I'm reading here it seems like property filtering should work with

`curl 'host/collections/collection_name/items?property_name=property_value'`

However, this didn't work in my case.

I'm on commit 8388148 of https://github.com/mbucknell/pygeoapi .

I thought it was supposed to be like here:
http://docs.pygeoapi.io/en/latest/code.html#module-pygeoapi.provider.postgresql

`curl 'host/collections/collection_name/items?properties=[(name1,value2),(name2,value2)]`

I believe it is not documented correctly or not working at the moment. Are there decisions to be made?

I would be happy to document the agreed way because I don't find it immediately apparent from the code.

Also, I think filtering on CSV and JSON data is possible, even though it might require some more code in the providers.

tomkralidis:
@geekdenz thanks for the info.  I will let @mbucknell, @jorgejesus and others chime in on this PR.  With regard to CSV/JSON, correct we would some code in the relevant providers, or one could use the OGR Provider.

mbucknell:
@geekdenz, @tomkralidis, I was following the guidance in http://docs.opengeospatial.org/DRAFTS/17-069r1.html#_parameters_for_filtering_on_feature_properties, which I am interpreting as name1=value2&name2=value2. The example that is given shows name1=value1.

geekdenz:
@mbucknell Thanks for your reply. Is there a configuration setting to enable this or is it enabled by default. Last time I tried it didn't work for me, but if you say it works or give any more hints I will try again.

mbucknell:
There is no longe a configuration setting for this so it's always enabled. The property name has to match the key in the returned list of properties for a Feature in a FeatureCollection. In the postgres table I am hitting there is a field named site_type. So to get a FeatureCollection containing only site_type equal to "Stream", I use the following query:
http://localhost:5000/collections/stewards_monitoring_locations/items?f=json&site_type=Stream

I hope that helps.

tomkralidis:
Thanks @mbucknell !

jorgejesus:
@geekdenz  I have tested master with the burundi data set (local computer) and it works as @mbucknell  described:

```
http://localhost:5000/collections/hotosm_bdi_waterways/items?&name=Muha
```

yaml conf:
```
        provider:
            name: PostgreSQL
            data:
                host: 127.0.0.1
                dbname: test
                user: postgres
                password: postgres
                port: 5432
                search_path: [osm,public]
            id_field: osm_id
            table: hotosm_bdi_waterways
            geom_field: geom

```

jorgejesus:
@mbucknell  Could you please #279 the merge has been accepted (find it nicer when the ticket creator closes it)

Issue #285, opened 2019-10-23
move OpenAPI endpoint from /api to /openapi


tomkralidis:
As per https://github.com/opengeospatial/ogcapi-features/issues/236, we are no longer required to support the API endpoint name with the specific naming of `/api`, which means we can rename our OpenAPI endpoint to anything we wish as long as it is identified and linked from the landing page (and elsewhere) accordingly.

For organizations deploying pygeoapi to a non root URL, `/api` might be a useful root endpoint, given pygeoapi will support multiple OGC API standards.

The update is simply to change the route name for `/api` to `/openapi`; this will allow one to deploy a production server to `https://example.org/foo/api` as the landing page thereby making `https://example.org/foo/api/openapi` the OpenAPI document (JSON or Swagger), and then `https://example.org/foo/api/collections`, `https://example.org/foo/api/processes` and so on.

Moving to `/openapi` also opens up to supporting various API definitions (like Capabilities, etc.) as part of API evolution over time.

PR #286, opened 2019-10-23
 move OpenAPI endpoint from /api to /openapi (#285)

tomkralidis:
As per #285


justb4:
This file/line needs changes as well:

* https://github.com/geopython/pygeoapi/blob/master/tests/pygeoapi-test-openapi.yml#L115

justb4:
And luckily on demo server the openapi doc is generated during Docker Image (re-)deploy.

tomkralidis:
@justb4 this is already done in https://github.com/geopython/pygeoapi/pull/286/files#diff-d918a41d8d149b949f58d10c445a3785R115

Issue #291, opened 2019-11-05
Provide a way to allow an implementer to map provider column names to provider names


mbucknell:
**Is your feature request related to a problem? Please describe.**
In our use case we have legacy tables where the column names are not human readable. We also have columns that we don't want to expose. 

**Describe the solution you'd like**
I'd like to come up with a way of adding configuration parameters to define a mapping of field names to property names. Initial implementation will focus on postgres provider but a general solution would be best.

pvgenuchten:
@mbucknell, should we in stead consider to allow a sql query as a base for a datastore, in such a sql query you could limit and rename fields, but also create joins and filters. Geoserver knows this concept as [sql views](https://docs.geoserver.org/stable/en/user/data/database/sqlview.html)

But for sure at some point I guess we'll need some configuration options on the attribute level, and renaming the property could be one of them. I haven't checked #246 in detail yet, but I guess a json-ld encoding will require for example ontology-information for each attribute.

mbucknell:
@pvgenuchten , I absolutely want to be able to allow a sql query as a base for a datastore. In fact, I'm going to explore that as part of the OGC/STAC hackathon. Trying to set up a postgres database with somewhat really data from our use case.

mbucknell:
I have submitted a PR for an example implementation: https://github.com/geopython/pygeoapi/pull/302

justb4:
As this is PostgreSQL, can't this be simply solved by creating a PG VIEW? Unless offcourse you are not allowed to modify the DB...And even then possibly could be solved at core API level? i.s.o  a variant for each driver?

mbucknell:
It could be solved by a view but it would be a database modification. Forgive my ignorance, but I'm not sure what you mean by solving at the core API level. Also I have implemented as a variant but I think that we can easily merge this code with the existing postgres provider. I wanted to get feedback before doing that.

jkreft-usgs:
Yeah, the goal here is that there are many organizations that may have an existing database table with a geospatial index, but the processes for making any actual database schema changes (including a materialized view) may be on a vastly slower timeline than the ability to deploy a web application.  Generalizing from postgres to being able to support any database should probably be addressed in a separate issue, and probably only if someone actually expresses an interest in an RBDMS other than postgres/postgis

jorgejesus:
@jkreft-usgs From your last comment, then the idea would be for a generic RBDMS dataprovider based on SQLAlchmey (or another ORM).

[...]

PR #302, opened 2019-11-15
Arbitrary query used for base of collection.

mbucknell:
I have created an enhanced postgres provider (postgres_query) which takes two new properties to replace the "table" config property. The property, column_property_mapping, is a dictionary mapping column names to property names. This is used in conjunction with the query_table parameter which can be a table name or a join statement.

This work is related to issue #291. I elected to create a new provider. Much of the code is duplicated in the postgresql provider. With additional work, I could merge these two but wanted to get feedback before that happened.

In order to test this well, I would also need to enhance the postgres data that is loaded for testing with probably some sort of relationship table that can be used for joining. I have tested this locally using my own postgres data, joining two tables.


mbucknell:
Overidden by PR #303

PR #303, opened 2019-11-20
Modified the postgres provider to use column_property_mapping and handling a complex query for table

mbucknell:
I modified the postgres provider. It now takes the configuration parameter column_property_mapping which defines the columns to fetch and the property names to use in the geojson response. In addition if column_property_mapping is provided, table does not need to be a simple table name but rather any query which produces a table (for example a join). 

This merges the work that I did in https://github.com/geopython/pygeoapi/pull/302. I will close that PR, in favor of this one.


mbucknell:
This overrides #302 and implements https://github.com/geopython/pygeoapi/issues/291

jorgejesus:
@mbucknell Your PR is awesome, there is just 1 small thing that IMHO need attention:

-  Could we have all the unittesting in a single file, integrating `test_postgresql_column_mapping.py` into `test_postgresql_provider.py` and having a different config dictionary for your unittests :)

Love that we can use this PR to filter columns, from the DB. IMHO this PR is a big improvement to this data provider :+1:

jorgejesus:
@tomkralidis  Can I approve the PR or you want to check it ?

jorgejesus:
closes #303

justb4:
@jorgejesus Bit puzzled: this PR overrides #302 (closed). Why is this PR #303 closed? Was merge intended? Or is there another PR/issue I am not aware of?

jorgejesus:
@justb4  Thank you for checking, merge was intended, the PR implements the `column_property_mapping`  and based on discussion on https://github.com/geopython/pygeoapi/issues/291#issuecomment-558841982 assumed that it would be OK to accept the PR.

This was my interpretation, hope that is OK,  for me is hard to follow all the discussion on different issues and PR, if it was a mistake please let me know

justb4:
@jorgejesus no sweat! My question was if you intended to merge this PR but instead closed it without merge. At least I don't see recent merges on `master`.

Issue #306, opened 2019-11-26
Display Data in Other Projections on Map in Viewer


geekdenz:
It's quite useful to have data in other projections. For us that is EPSG:2193. We currently only need this in the PostgreSQL provider.

PR #307, opened 2019-11-26
allow other projections than WGS84 (EPSG:4326) such as NZTM (EPSG:219…

geekdenz:
…3) in viewer #306

So, we can now have source data in another projection and the OSM Leaflet map should still show the coordinates projected into WGS84.


geekdenz:
Sorry forgot about the tests... Will have a look into them.

jorgejesus:
@geekdenz Do you have some dataset for testing ?

Issue #310, opened 2019-11-26
Previous and Next Links do not work


geekdenz:
**Description**
Previous and next links on the viewer of a single item do not work.

**Steps to Reproduce**

1. Go to a specific item.
2. Click on previous or next link.

**Expected behavior**
Previous or next item shows when link is clicked, but same item is displayed.

**Environment**
 - OS: Ubuntu 18.04
 - Python version: 3.7.3
 - pygeoapi version: b61404d

alpha-beta-soup:
>  **Expected behavior**
> Previous or next item shows when link is clicked, but same item is displayed.

My expected behaviour would be either:
- There are no prev/next links when looking at a single feature
- They work but navigate to different features (e.g. prev/next with default sorting).

tomkralidis:
This may or may not be trivial for a given provider backend to come up with.  Having said this, +1 for removing prev/next links from `pygeoapi/templates/item.html` in #311

geekdenz:
@tomkralidis thanks for your feedback.

The previous/next links are broken before this fix for all providers. This fixes it for the pgsql provider and I think it's definitely possible to fix it, at least somewhat, for other providers. Also, I tested this with other providers and it didn't break anything further because the link then just does not work.

Maybe the previous/next links should be removed only for providers where it's not implemented. I'm happy to do this change.

Having said all that it might not be trivial to merge this fix in. If you like I can have a go at merging it into current master. Please let me know.

tomkralidis:
Thanks @geekdenz.  I think it's safe to not assume all providers can implement this functionality and we can test for this in HTML mode, which you're implemented (thanks!).  Minor comments remaining in #311.

PR #311, opened 2019-11-26
Issue 310

geekdenz:


tomkralidis:
@geekdenz thanks.  See comments in https://github.com/geopython/pygeoapi/issues/310#issuecomment-565622894 .  Can we remove prex/next on single item HTML?

pygeoapi/provider/postgresql.py
```diff
@@ -265,6 +265,44 @@ def query(self, startindex=0, limit=10, resulttype='results',
...
+        """
+        sql = 'SELECT {} AS id FROM {} WHERE {}<%s ORDER BY {} DESC LIMIT 1'
+        cursor.execute(SQL(sql).format(
+            Identifier(self.id_field),
+            Identifier(self.table),
+            Identifier(self.id_field),
+            Identifier(self.id_field),
+        ), (identifier,))
+        item = cursor.fetchall()
+        id = item[0]['id']
```
tomkralidis: `id` is a Python built-in function.  Suggest using `id_`.

pygeoapi/provider/postgresql.py
```diff
@@ -298,6 +336,8 @@ def get(self, identifier):
             row_data = cursor.fetchall()[0]
             feature = self.__response_feature(row_data)
 
+            feature['previous'] = self.get_previous(cursor, identifier)
```
tomkralidis: Suggest changing to `feature['prev']`.

pygeoapi/provider/postgresql.py
geekdenz: Would you like me to change that? I think it's more guaranteed to work the way it is. Also, it might be easier to read for someone like me who didn't know this. So, it is just an estaetic improvement, no?

pygeoapi/provider/postgresql.py
geekdenz: We could and in the interest of brevity I agree. However, in the interest of readability it might be better to stay 'previous'. I don't really mind though. Feel free to change it if you prefer or let me know if you want me to do it.

pygeoapi/provider/postgresql.py
tomkralidis: Given `id` is a Python [function](https://docs.python.org/3/library/functions.html#id) it would be safer not to override function names in the case that they be used elsewhere in the codebase.

pygeoapi/provider/postgresql.py
tomkralidis: If you can change to `prev` that would be consistent with OGC API - Features proper. Thanks.

geekdenz: Thanks @tomkralidis! Your points absolutely make sense. It shall be done. :-)

pygeoapi/provider/postgresql.py
geekdenz: Agree to this @tomkralidis.

tomkralidis:
@geekdenz somehow I was able to edit this PR as per the suggested changes.  If this works for you, suggest we merge this PR and close #328 ?

pygeoapi/templates/item.html
```diff
@@ -37,7 +37,28 @@ <h2>Item {{ data['id'] }}</h2>
...
+            </div>
+          </div>
+          <div class="row">
+            <div class="col-sm-12">
+              {% if data['previous'] or data['next'] %}
+              <div class="row">
+                <div class="col-sm-12">
+                  {% for link in data['links'] %}
+                  {% if link['rel'] == 'prev' %}
+                  <a role="button" href="./{{ data['previous'] }}">Prev</a>
```
geekdenz: @tomkralidis Please ensure these lines have "prev" instead of "previous" to not break things. Otherwise all good.

PR #319, opened 2019-12-11
Update README.md

KoalaGeo:
Struggled with install documentation so updated here. Happy to chip in with readthedocs also.


tomkralidis:
Can you move this as an intro in https://github.com/geopython/pygeoapi/tree/master/docker instead?  This is very valuable info for getting up and running on installing Docker -- thanks!

KoalaGeo:
Sure, done.

tomkralidis:
Thanks!

PR #321, opened 2019-12-16
Initial version of mongo provider

timtuun:


timtuun:
CI fails due to missing mongodb.
We should install mongo to CI environment and I don't know how to do this...
sudo apt-get install -y mongodb

tomkralidis:
Thanks for this functionality! Comments:

- try adding `mongodb` after https://github.com/geopython/pygeoapi/blob/master/.travis.yml#L18 to see if that works?
- can you provide a scripts in `tests` which loads the data in your tests to mongo?  Similar test loader for ES as an example: https://github.com/geopython/pygeoapi/blob/master/tests/load_es_data.py

timtuun:
I need to rewrite the unit tests to use the data loaded by load_mongo_data.py. Probably this functionality should not be merged before that.

tomkralidis:
Thanks @timtuun.  Let us know when this PR is ready for review again.

timtuun:
Hi @tomkralidis. I think this PR can be reviewed now.

requirements.txt
```diff
@@ -3,3 +3,4 @@ python-dateutil
 Flask
 PyYAML
 unicodecsv
+pymongo
```
tomkralidis: Can we move the pymongo dependency to `requirements-provider.txt` instead?

pygeoapi/plugin.py
```diff
@@ -43,6 +43,7 @@
         'GeoPackage': 'pygeoapi.provider.geopackage.GeoPackageProvider',
         'OGR': 'pygeoapi.provider.ogr.OGRProvider',
         'PostgreSQL': 'pygeoapi.provider.postgresql.PostgreSQLProvider',
+        'Mongo': 'pygeoapi.provider.mongo.MongoProvider',
```
tomkralidis: For consistency, can / should this be called `'MongoDB': 'pygeoapi.provider.mongo.MongoProvider'`

tests/test_mongo_provider.py
```diff
@@ -0,0 +1,219 @@
...
+monogourl = 'mongodb://localhost:27017/'
+mongodb = 'testdb'
+mongocollection = 'testplaces'
+
+
+@pytest.fixture()
+def config():
+    return {
+        'name': 'Mongo',
+        'data': {
```
tomkralidis: 1. For consistency (see Elasticsearch provider), can `data` be in the Connection String URI format (see https://docs.mongodb.com/manual/reference/connection-string/) ?
2. Can the `fields` property be populated dynamically (instead of hard coding in config)?

pygeoapi/plugin.py
timtuun: Changed.

requirements.txt
timtuun: I guess it could. :) Moved it.

tests/test_mongo_provider.py
timtuun: 1. Updated as requested.

2. Updated get_fields to be dynamic.

timtuun:
I don't get it. Suddenly api -test are failing and I don't think I am responsible of that... I guess some packages in test environment have changed, so that they produce different results. @tomkralidis are you able to successfully run tests in master?

tomkralidis:
@timtuun I tested locally with success.  Looks like the test failures are unrelated.  Merging.  Thanks!

PR #322, opened 2019-12-16
Issue 290

jorgejesus:
This PR implements:

- Single dataprovider for sqlite and geopackage
- Data provider can use label filter and bbox filder
- pytest update for single sqlite/geopackage provider
- data test poi_portugal.gpkg was reduced to around 10%
- pytest for ogr_geopackage was update to new data test
- Code clean up on data provider


jorgejesus:
Code was updated to use format() instead of f string. Sqlite and postgresl data provider was updated

PR #331, opened 2020-01-07
xarray coverage provider

gpetrochenkov-usgs:
This is the xarray implementation.


tomkralidis:
Thanks @gpetrochenkov-usgs.  As discussed can you add a small zarr file which we can include in tests once we start moving this into master? Thanks

gpetrochenkov-usgs:
Hey Tom, I went ahead and added a zarr file to the tests/data directory.

tomkralidis:
Thanks Greg.  I see the zarr file now adds ~120 or so files.  Can zarr files be bundled/zipped somehow?

gpetrochenkov-usgs:
Let me look in to that, as far as I know they are supposed to be stored in that directory format.  Maybe we can setup the test to unzip the file, test, and then remove so the repository is cleaner.

gpetrochenkov-usgs:
So I've been looking in to it and the store directory is the only type that is supported for use in xarray.  I can write another provider using just the zarr package and then utilize zip files, or help write a test case that would unzip the directory and manipulate the data.  Let me know how you'd like to proceed.

gpetrochenkov-usgs:
Also, after pulling down the latest coverages branch, the query calls fail because of a non existing 'bands' argument.

tomkralidis:
Thanks @gpetrochenkov-usgs.  Can you zip the zarr data into a file and store as `tests/data/sst.zarr.zip`.  We can work on tests and unzipping/etc. as the branch hardens.

tomkralidis:
RE: failing tests, in your provider .py file, change, in your `query()` function, change`range_subset` to `bands`.

gpetrochenkov-usgs:
I went ahead and removed the old zarr file and added the zipped version to the tests/data directory.

Issue #342, opened 2020-01-16
dataset>spatial has wrong json-ld content


pvgenuchten:
for the collection page google structured data testing tool reports an error on 

```
"spatial": [{"Place": {"@type": "GeoShape", "box”:
```
shouldn’t it be?

```
"spatial": [{“@type”:”Place”, “geo”: {“@type”:”GeoShape”, "box":
```

error:

```
The property Place is not recognised by Google for an object of type Place.
```

tomkralidis:
Thanks @pvgenuchten

PR #343, opened 2020-01-16
Place is a type not a property

pvgenuchten:
fix for #342


pvgenuchten:
mmm, i think someone has to explain me the assert syntax for these json-ld blobs...

pvgenuchten:
sorry about all commits to fix travis, can you squash commits on merge?

Issue #312, opened 2019-11-26
http://schemas.opengis.net/ is not running in HTTPs


jkreft-usgs:
**Description**
http uris are making swagger break due to mixed-content issues.  The TODO is indeed something that needs to be done

[Offending Line](https://github.com/geopython/pygeoapi/blob/61b1617d562c0a5573d21759fe6fdefb59cac2fe/pygeoapi/openapi.py#L44): 

```
# TODO: handle this better once schemas are public/final
# allow also for schema caching
OPENAPI_YAML = {
    'oapif': 'http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml',  # noqa
    'oapip': 'https://raw.githubusercontent.com/opengeospatial/wps-rest-binding/master/core/openapi'  # noqa
}
```

**Steps to Reproduce**
load the swagger doc on a page that runs over https.  The browser will block the call to `http://schemas.opengis.net/` because of mixed content, and then the swagger will throw errors when you click on various endpoints, including the `collections/{collection-id}/items endpoint

If you force your browser to accept mixed content, the error goes away.

**Expected behavior**
Swagger works on https, including all of the collection query parameters

**Screenshots/Tracebacks**
![image](https://user-images.githubusercontent.com/8128936/69676603-b786e100-1066-11ea-9e46-2b01e38623dd.png)

**Environment**
 - OS: Docker (Alpine?)
 - Python version: 3.6 
 - pygeoapi version: Latest

**Additional context**
The easiest solution would probably be for schemas.opengis.net to just get a certificate and properly set HSTS and forwarding as well.

jkreft-usgs:
You can see this issue on the demo site as well: https://demo.pygeoapi.io/master/openapi?f=html#/obs/get_collections_obs_items

jkreft-usgs:
A path forward while OGC gets its ducks in a row would be to make the definition of  `OPENAPI_YAML` configurable so that I can easily point to a copy of the YAML file somewhere that does support `https`.

tomkralidis:
Implemented in #347

pvgenuchten:
Referencing duplicated schemas on a https location does indeed solve the case. 
Is the documentation updated to require this configuration step for https environments?
Has there been any contact with OGC to discuss if they can also host those schema's on https?

tomkralidis:
@pvgenuchten 

[...]

PR #347, opened 2020-01-20
Implement configurable OGC schema locations

tomkralidis:
cc @jkreft-usgs 

Implements #312 

# Overview
In the spirit of https://mapserver.org/ogc/wms_server.html#index-21, this PR implements configurable OGC schema location for deployments wishing to serve local copies of the OGC schemas.

# Approach
- pygeoapi administrator downloads .zip or .tgz of http://schemas.opengis.net
- pygeoapi configuration sets `server.ogc_schemas_location` to one of:
  - local path (resulting URL is `server.url` + `server.ogc_schemas_location`)
  - base URL 
- if `server.ogc_schemas_location` is not set, default is `http://schemas.opengis.net`

### DIFF
--- a/pygeoapi/flask_app.py
+++ b/pygeoapi/flask_app.py
@@ -3,7 +3,7 @@
 # Authors: Tom Kralidis <tomkralidis@gmail.com>
 #          Norman Barker <norman.barker@gmail.com>
 #
-# Copyright (c) 2018 Tom Kralidis
+# Copyright (c) 2020 Tom Kralidis
 #
 # Permission is hereby granted, free of charge, to any person
 # obtaining a copy of this software and associated documentation
@@ -34,10 +34,10 @@ import os
 
 import click
 
-from flask import Flask, make_response, request
+from flask import Flask, make_response, request, send_from_directory
 
 from pygeoapi.api import API
-from pygeoapi.util import yaml_load
+from pygeoapi.util import get_mimetype, yaml_load
 
 APP = Flask(__name__)
 APP.url_map.strict_slashes = False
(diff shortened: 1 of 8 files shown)


pygeoapi/flask_app.py
```diff
@@ -55,11 +56,42 @@
...
+        """
+
+        mimetypes.add_type('text/plain', '.yaml')
+        mimetypes.add_type('text/plain', '.yml')
+
+        full_filepath = os.path.join(OGC_SCHEMAS_LOCATION, path)
+        dirname_ = os.path.dirname(full_filepath)
+        basename_ = os.path.basename(full_filepath)
+
+        print('FULL_FILEPATH', full_filepath)
```
francbartoli: Can we log here rather than printing?

pygeoapi/flask_app.py
```diff
@@ -55,11 +56,42 @@
...
+
+        :param path: path of the OGC schema document
+
+        :returns: HTTP response
+        """
+
+        mimetypes.add_type('text/plain', '.yaml')
+        mimetypes.add_type('text/plain', '.yml')
+
+        full_filepath = os.path.join(OGC_SCHEMAS_LOCATION, path)
```
francbartoli: Should we catch an exception somewhere if the local file doesn't exist?

tomkralidis: Isn't this covered by starlette/flask accordingly?

francbartoli: Yes, it is but does it make sense to give the users a `404` for a file that should have been placed manually by the admin? Maybe a fall back to the default?

francbartoli: For `starlette_app.py` what do you think of this below?:

```python
if (OGC_SCHEMAS_LOCATION is not None and
        not OGC_SCHEMAS_LOCATION.startswith('http')):
    try:
        os.path.exists(OGC_SCHEMAS_LOCATION)
        # serve the OGC schemas locally
        app.mount('/schemas', StaticFiles(directory=OGC_SCHEMAS_LOCATION))
    except FileNotFoundError:
        raise RuntimeError(“Wrong local path location”)
```

tests/test_util.py
```diff
@@ -96,3 +96,9 @@ def test_json_serial():
 
     with pytest.raises(TypeError):
         util.json_serial('foo')
+
+
+def test_mimetyhpe():
```
francbartoli: typo in the function name?

tomkralidis: Fixed.

pygeoapi/flask_app.py
tomkralidis: Good idea.  Updated.

PR #348, opened 2020-01-23
allow other projections than WGS84 (EPSG:4326) such as NZTM (EPSG:219…

geekdenz:
…3) in viewer #306

Note that I squashed my commits and added a unit test here. If accepting this one, the other one can be closed as well.


alpha-beta-soup:
> Also there is an issue with outputing gpgk when content negociation is only sqlite3

Yes, there is a distinction there, and it may require more care and precision. For instance, GeoPackage and Spatialite geometries are distinct., as I encountered once: https://gis.stackexchange.com/questions/340179/spatialite-spatial-queries-do-not-function-on-geopackage

PR #357, opened 2020-02-01
change irc notifications to #pygeoapi-activity channel

jmckenna:
- some think notifications flood the discussion channel
- so this method is used by a few other projects: send notifications to a separate channel #pygeoapi-activity
  - the new channel has been registered


justb4:
👍

PR #358, opened 2020-02-03
Issue 329: Add GeoPackage Support

geekdenz:
This PR allows GeoPackage download with a standard Accept headers of
application/geopackage+sqlite3
or
application/x-sqlite3
and with the ?f=geopackage URL parameter.

Currently only normal filters work. One cannot download the whole datasource, at least not efficiently. However, for us it is more important that all filters work with GeoPackage.

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -39,6 +39,9 @@ from dateutil.parser import parse as dateparse
 import pytz
 
 from pygeoapi import __version__
+from pygeoapi.gpkg import (
+    format_handler as gpkg_format_handler,
+)  # could use dictionary for format handlers
 from pygeoapi.linked_data import (geojson2geojsonld, jsonldify,
                                   jsonldify_collection)
 from pygeoapi.log import setup_logger
@@ -56,7 +59,7 @@ HEADERS = {
 }
 
 #: Formats allowed for ?f= requests
-FORMATS = ['json', 'html', 'jsonld']
+FORMATS = ['json', 'html', 'jsonld', "geopackage"]
 
 CONFORMANCE = [
     'http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core',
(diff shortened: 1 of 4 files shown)


pygeoapi/api.py
```diff
@@ -176,10 +179,16 @@ def root(self, headers_, format_):
               'type': 'application/json',
               'title': 'Collections',
               'href': '{}/collections'.format(self.config['server']['url'])
-            }
+            }, {
+                "rel": "self",
```
alpha-beta-soup: Shouldn't this be `alternate`?

pygeoapi/api.py
```diff
@@ -176,10 +179,16 @@ def root(self, headers_, format_):
...
+                "type": "text/html",
+                "title": "This document as HTML",
+                "href": "{}?f=html".format(self.config["server"]["url"]),
+                "hreflang": self.config["server"]["language"],
+            },
         ]
 
-        if format_ == 'html':  # render
+        if format_ == 'html' or format_ == "geopackage":  # render
```
alpha-beta-soup: Why render HTML if a GeoPackage is requested?

pygeoapi/templates/base.html
```diff
@@ -40,6 +40,7 @@
         <a href="{{ config['server']['url'] }}">Home</a>
         {% endblock %}
         <span style="float:right">
+          <a href="{{ data['links'] | map(rel='self') | attr('href') }}?f=geopackage">GEOPACKAGE</a>
```
alpha-beta-soup: I think this should be in the list of links in the main part of the page, rather than this "navbar" part. The JSON and JSON-LD representations can provide metadata about a collection, without any feature-level data. But the GeoPackage is intended for features, not metadata.

geekdenz: Yes, agree. I can fix this in the PR. Thanks @alpha-beta-soup . I may throw in the ability to download the complete dataset as GeoPackage as well.

tomkralidis:
Looks good at first glance. This needs to be implemented as a formatter plugin (see the CSV formatter plugin for design), and then loaded similarly to https://github.com/geopython/pygeoapi/blob/master/pygeoapi/api.py#L797 

pygeoapi.api does need some refactoring so as not to have all the `if..elif` around download formats that we currently have.  However, I would suggest for this PR to make GPKG a plugin as the first step and then we can refactor proper in another PR.

geekdenz:
Should we close this?

tomkralidis:
@geekdenz +1, closed.  Once we refactor `pygeoapi/api.py` some, output formats will be more pluggable.

PR #359, opened 2020-02-03
GDAL3 support

### LINKED ISSUE #318: Move GDAL dependency for OGR provider on version 3
Upgrade GDAL library to version 3

It would open up new scenarios for implementing ad-hoc providers by wrapping the [new python driver](https://github.com/OSGeo/gdal/blob/7b7ecbab1e2b10ecab24e5126d94469d2d4efdc9/gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst#example) when available.

None

**Additional context**

francbartoli:
Trying to fix #318

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,5 @@ ENV/
 #Eclipse project
 .project
 
+
+.DS_Store
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,9 +75,8 @@ script:
   # run docker image with cite configuration
   - docker run -d -p 5001:5001 --network host --add-host="localhost:127.0.0.1" --rm -it -v $(pwd)/tests/cite/ogcapi-features/cite.config.yml:/pygeoapi/local.config.yml --name pygeoapi-travis-master geopython/pygeoapi:latest run
   - docker ps | grep -wq "pygeoapi-travis-master"
-  - docker logs pygeoapi-travis-master 2>&1 |grep -wq "Listening at:"
+  - docker logs pygeoapi-travis-master |grep -wq "Listening at:"
   - docker run --rm --name pygeoapi-travis-runtests-master geopython/pygeoapi:latest test
-
 
 after_success:
   - python3 setup.py sdist bdist_wheel --universal
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 
-dist: xenial
+dist: bionic
 sudo: false
 
 env:
@@ -15,19 +15,23 @@ python:
 
 services:
   - postgresql
-  - mongodb
 
 addons:
-  postgresql: 9.6
+  apt:
+    sources:
+      - sourceline: 'ppa:ubuntugis/ubuntugis-unstable'
+      - sourceline: 'deb [ arch=amd64 ] https://repo.mongodb.org/apt/ubuntu bionic/mongodb-org/4.0 multiverse'
+    packages:
+      - gdal-bin
+      - libgdal-dev
+      - libsqlite3-mod-spatialite
+      - python-pyspatialite
+      - postgresql-10-postgis-2.4
+      - mongodb-org
 
 before_install:
-  - sudo apt-add-repository ppa:ubuntugis/ubuntugis-unstable -y
-  - sudo apt-get -qq update
-  - sudo apt-get install -qq build-essential python3-dev python3-setuptools python-all python3-all python-setuptools
-  - sudo apt-get install -y libsqlite3-mod-spatialite devscripts fakeroot debhelper
-  - sudo apt-get install -y postgresql-9.6-postgis-2.4
-  - sudo apt-get install -y libgdal-dev gdal-bin
-  - sudo apt-get install -y mongodb
+  - sudo apt-get install -qq build-essential python3-dev python3-setuptools python-all python3-all python-setuptools python3-pip
+  - sudo apt-get install -y devscripts fakeroot debhelper
 
   # install Elasticsearch 7
   - curl -s -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-7.5.1-amd64.deb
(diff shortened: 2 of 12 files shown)


francbartoli:
For reference to the new proj6 SRS transformations let's have a look at this [post](http://osgeo-org.1560.x6.nabble.com/gdal-dev-Possible-bug-in-gdal-TransformPoint-gdal-3-0-1-td5426659.html)

justb4:
Travis is IMO failing as it sets up for Ubuntu Xenial (16.4) which does not provide GDAL v3 via Ubuntugis unstable. .Travis file needs to upgrade to Bionic (18.4) as to enable GDAL v3, see also https://docs.travis-ci.com/user/reference/bionic/ and Ubuntugis PPA info: https://launchpad.net/~ubuntugis/+archive/ubuntu/ubuntugis-unstable .

francbartoli:
Thanks @justb4, I will try to see what will happen once Travis will be upgraded. The PR wasn't intended to pass Travis though cause they are still failing locally. I wanted to share it if someone else could have hints about the current failure

justb4:
Looks like the tests are failing because of changes coming with Proj.6, right? Axis Ordering after all? Did you manage to get the failing tests working locally with the migration hints from https://github.com/OSGeo/gdal/blob/release/3.0/gdal/MIGRATION_GUIDE.TXT#L7 ?

francbartoli:
@justb4 No I didn't even after I've applied the hints, see [c242040ea49](https://github.com/geopython/pygeoapi/pull/359/commits/c242040ea4908f38fea4ff3749e763b1aaf271ba)

francbartoli:
I fixed the installation on Travis, now only tests remaining stuck. They are failing exactly the same as I got locally

francbartoli:
@justb4 I solved the issue, it was a very very stupid error on my side :(. Now, tests should pass so the question is if we want to support both GDAL versions. Maybe we can achieve this as an improvement in a different PR? What is the feeling of others? @tomkralidis @jorgejesus @kalxas @pvgenuchten

tomkralidis:
+1 for GDAL 3 only support.  We can investigate as part of another PR if required.  Nice work here @francbartoli

kalxas:
I also agree that we should move to GDAL3 only right now. Thanks @francbartoli !

.gitignore
```diff
@@ -110,3 +110,5 @@ ENV/
 #Eclipse project
 .project
 
+
+.DS_Store
```
tomkralidis: Is this change intentional ?

francbartoli: It is @tomkralidis, despite just in favour of OSX users

francbartoli:
@justb4 thanks for the feedback and comments. Since the build of docker images is not in the travis pipeline is it still fine to merge it on master and open another issue for docker images with GDAL 3? This would also make the merge of https://github.com/geopython/pygeoapi/pull/368 easier

justb4:
> @justb4 thanks for the feedback and comments. Since the build of docker images is not in the travis pipeline is it still fine to merge it on master and open another issue for docker images with GDAL 3? This would also make the merge of #368 easier

Hmm, the Docker Image is rebuilt and deployed on demo.pygeoapi.io automatically on commit/push of the GitHub repo. The GDAL-based code in the `OGRProvider` will fail (hence I was thinking of fallback to GDAL2 there). We have a [similar  issue in the Stetl project](https://github.com/geopython/stetl/issues/93) so we can apply a similar solution for the Docker file.

justb4:
One thing we could do is upgrade the Docker file from `buster-slim` (Debian 10) to `bullseye-slim` (debian 11) although `bullseye` is the current `Debian testing` release as Debian 11 support GDAL (3.0.4) currently: https://packages.debian.org/testing/science/gdal-bin .Thoughts? @kalxas ?

Dockerfile
```diff
@@ -61,7 +63,7 @@ ARG ADD_PIP_PACKAGES=""
 ENV TZ=${TIMEZONE} \
 	DEBIAN_FRONTEND="noninteractive" \
 	DEB_BUILD_DEPS="tzdata build-essential python3-setuptools python3-pip apt-utils git" \
-	DEB_PACKAGES="locales libgdal20 python3-gdal libsqlite3-mod-spatialite curl ${ADD_DEB_PACKAGES}" \
+	DEB_PACKAGES="locales libgdal-dev python3-gdal libsqlite3-mod-spatialite curl ${ADD_DEB_PACKAGES}" \
```
justb4: IMO the -dev package is not required (maybe that also adds to Image size?). As it looks this should be the GDAL package: https://packages.debian.org/bullseye/libgdal26 so `libgdal26` which is strangely GDAL 3.0.4. Have you tried already?

francbartoli:
@justb4 I thought about to run the tests within the image but I didn't realize that could be done without backend servers. However, it might be likely possible to reach the services on the host with the option I set in the run and execute `pytest tests`. I will try as the last attempt and in case of negative result I will then force the reset of that commit

.travis.yml
```diff
@@ -54,16 +65,23 @@ before_script:
...
+  # load cite data
+  - python3 tests/load_es_data.py tests/cite/ogcapi-features/canada-hydat-daily-mean-02HC003.geojson IDENTIFIER
 
 script:
   - pytest --cov=pygeoapi
   - make -C ./docs html
   - find . -type f -name "*.py" | xargs flake8
+  # run docker image with cite configuration
+  - docker run -d -p 5001:5001 --network host --add-host="localhost:127.0.0.1" --rm -it -v $(pwd)/tests/cite/ogcapi-features/cite.config.yml:/pygeoapi/local.config.yml --name pygeoapi-travis-master geopython/pygeoapi:latest run
+  - docker ps | grep -wq "pygeoapi-travis-master"
```
justb4: OK, this fails now. Rethinking, this is a bit complex to setup as services on host are needed from within the Docker Container, always very complex and error-prone. Better via a Docker compose-file like the [demo one](https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_cite/docker-compose.yml). Also the `docker logs | grep` is not a good test I realize. Possibly better to just 

1) start the Docker image, wait til ready
2) do sanity check with some `curl `requests
2) stop the Docker image

But the Unit tests all run ok with Docker. That already gives confidence. So I suggest to skip for now and open a new issue to implement robust Docker tests for Travis.  For now, if you remove the Cite Docker commands, all Travis should be ok, and we can merge!

francbartoli: Well I understand that your plan is good but more inclined to an e2e strategy which can be opened as a different issue. Everything is now passing so I would merge.

Issue #299, opened 2019-11-07
Postgres provider doesn't handle query parameters correctly when resulttype=hits


mbucknell:
**Description**
When the postgres provider is used and a query for items using the parameter resulttyepe=hits, it always returns the count of the total items in the collection. If other query parameters are provided, they are ignored and the result still contains the total number of items in the collection

**Steps to Reproduce**
For a postgres provider, use the propertyname=propertyvalue query parameter along with resulttype=hits. The returned count will be the total number if the collection rather than the number of items in the filtered collection.

**Expected behavior**
I expect the count to reflect the number of items returns when resulttype=hits is not included on the query.

**Screenshots/Tracebacks**

**Environment**
 - OS:
 - Python version: 
 - pygeoapi version: 

**Additional context**
This is for the postgres provider

jorgejesus:
Just noticed on the postgresql provider, we have `self.conn_dic = provider_def['data']` which is already created by the parent class (BaseProvider) but called `self.data = provider_def['data']`

jorgejesus:
``http://localhost:5000/collections/hotosm_bdi_waterways/items?name=Muha&resulttype=hits&f=json`` should return 2

Issue #364, opened 2020-02-19
Missing SQLite plugin with Docker Image


jvanulde:
Trying to launch pygeoapi container using the following:

    docker run -p 5000:80 -it geopython/pygeoapi

Get the following error:

    START /entrypoint.sh
    Trying to generate openapi.yml
    Plugin SQLite not found
    NoneType: None
    Traceback (most recent call last):
      File "/usr/local/bin/pygeoapi", line 11, in <module>
    load_entry_point('pygeoapi', 'console_scripts', 'pygeoapi')()
      File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
      File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
      File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
      File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
      File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
      File "/usr/local/lib/python3.7/dist-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
      File "/pygeoapi/pygeoapi/openapi.py", line 519, in generate_openapi_document
    click.echo(yaml.safe_dump(get_oas(s), default_flow_style=False))
      File "/pygeoapi/pygeoapi/openapi.py", line 504, in get_oas
    return get_oas_30(cfg)
      File "/pygeoapi/pygeoapi/openapi.py", line 336, in get_oas_30
    p = load_plugin('provider', cfg['datasets'][k]['provider'])
      File "/pygeoapi/pygeoapi/plugin.py", line 82, in load_plugin
    raise InvalidPluginError(msg)
    pygeoapi.plugin.InvalidPluginError: Plugin SQLite not found
    ERROR: openapi.yml could not be generated ERROR

**Environment**
 - OS: MacOS
 - Python version: 3.7
 - pygeoapi version: latest

justb4:
The current Docker Image does not work with the SQLIte Provider (Plugin). See also [here](https://github.com/geopython/geopython-workshop/issues/40#issuecomment-586559565).
[...]

PR #368, opened 2020-02-23
where implementation on resulttype=hits

jorgejesus:
Implements `where` clauses for postgresql dataprovider as generic private method that is the used accounting the `resulttype` request 

Closes: #299

### DIFF
--- a/pygeoapi/provider/postgresql.py
+++ b/pygeoapi/provider/postgresql.py
@@ -174,6 +174,35 @@ class PostgreSQLProvider(BaseProvider):
                 self.fields = db.fields
         return self.fields
 
+    def __get_where_clauses(self, properties=[], bbox=[]):
+        """
+        Generarates WHERE conditions to be implemented in query.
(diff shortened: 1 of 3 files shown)


jorgejesus:
@francbartoli Made a copy from commit https://github.com/geopython/pygeoapi/commit/0ae20848f3ae52dd8b8ae73bd4a57b44fbd8671d to test it, pytest works local but no on travis, only difference is on one SQL and the usage of psycopg2-binary

jorgejesus:
Current commit implements `DISTINCT` 
https://github.com/jorgejesus/pygeoapi/blob/3caf4a92a4b888af0313d94dcef623a198bc1c1f/pygeoapi/provider/postgresql.py#L255

Locally pytest on ubuntu 18.04 is  with postgresql9.6-postgis2.4 (docker from katoza) it works fijne, but on travis (ubuntu 16.04)  the pytest is broken reporting different features

jorgejesus:
closes #364

PR #372, opened 2020-02-28
Fix issue #371

### LINKED ISSUE #371: OGR Provider fails to output geojson encoded feature with size bigger than 200MB
**Description**
Working with this dataset:

```yaml
countries:
        title: Countries
        description: Admin0 official countries from WFP
        keywords:
            - admin0
            - countries
            - WFP
        links:
            - type: text/html
              rel: canonical
              title: information
              href: https://geonode.wfp.org/layers/geonode%3Aglb_bnd_adm0_1
              hreflang: en-EN
        extents:
            spatial:
                bbox: [-180,-90,180,90]
                crs: http://www.opengis.net/def/crs/OGC/1.3/CRS84
            temporal:
                begin: null
                end: now  # or empty
        provider:
            name: OGR
            data:
                source_type: WFS
                source: WFS:https://geonode.wfp.org/geoserver/wfs
                source_srs: EPSG:4326
                target_srs: EPSG:4326
                source_capabilities:
                    paging: True
                source_options:
                    OGR_WFS_LOAD_MULTIPLE_LAYER_DEFN: NO
                open_options:
                    EXPOSE_GML_ID: NO
                    URL: https://geonode.wfp.org/geoserver/wfs?outputformat=json
                gdal_ogr_options:
                    EMPTY_AS_NULL: NO
                    GDAL_CACHEMAX: 64
                    CPL_DEBUG: NO
                    GDAL_HTTP_UNSAFESSL: YES
            id_field: adm0_id
            layer: geonode:glb_bnd_adm0_1
```

some features that have geojson output bigger than the default `OGR_GEOJSON_MAX_OBJ_SIZE=200MB` fail with several error messages like:

```shell
/pygeoapi/pygeoapi/provider/ogr.py:122} ERROR - Error Number: 1, Type: Failure, Msg: GeoJSON object too complex, please see the OGR_GEOJSON_MAX_OBJ_SIZE environment option
```

**Steps to Reproduce**

For example _Indonesia_ with `adm0_id=272`

`curl -XGET http://localhost:5000/collections/countries/items/272`

**Expected behavior**
It is expected the correct geojson payload of the response. This can be mitigated in GDAL version 3.0.2 and above by setting this environment variable:

```shell
OGR_GEOJSON_MAX_OBJ_SIZE=500MB
```

However the error happening in the code is not handled properly

**Screenshots/Tracebacks**

**Environment**
 - OS: Darwin
 - Python version: 3.7.3
 - pygeoapi version: master

**Additional context**

francbartoli:
Working in progress to fix #371

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -855,8 +855,16 @@ class API(object):
         p = load_plugin('provider',
                         self.config['datasets'][dataset]['provider'])
 
-        LOGGER.debug('Fetching id {}'.format(identifier))
-        content = p.get(identifier)
+        try:
+            LOGGER.debug('Fetching id {}'.format(identifier))
+            content = p.get(identifier)
+        except ProviderQueryError:
+            exception = {
+                'code': 'NoApplicableCode',
+                'description': 'query error (check logs)'
+            }
+            LOGGER.error(exception)
+            return headers_, 500, json.dumps(exception)
 
         if content is None:
             exception = {
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -858,12 +858,12 @@ class API(object):
         try:
             LOGGER.debug('Fetching id {}'.format(identifier))
             content = p.get(identifier)
-        except ProviderQueryError:
+        except ProviderQueryError as err:
             exception = {
                 'code': 'NoApplicableCode',
                 'description': 'query error (check logs)'
             }
-            LOGGER.error(exception)
+            LOGGER.error(err)
             return headers_, 500, json.dumps(exception)
 
         if content is None:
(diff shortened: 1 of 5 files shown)


pygeoapi/api.py
```diff
@@ -855,8 +855,16 @@ def get_collection_item(self, headers_, format_, dataset, identifier):
         p = load_plugin('provider',
                         self.config['datasets'][dataset]['provider'])
 
-        LOGGER.debug('Fetching id {}'.format(identifier))
-        content = p.get(identifier)
+        try:
+            LOGGER.debug('Fetching id {}'.format(identifier))
+            content = p.get(identifier)
+        except ProviderQueryError:
```
tomkralidis: Perhaps we can log the actual error message instead?  See below:

```python
        except ProviderQueryError as err:
            exception = {
                'code': 'NoApplicableCode',
                'description': 'query error (check logs)'
            }
            LOGGER.error(err)
            return headers_, 500, json.dumps(exception)
```

francbartoli: Good point!

PR #378, opened 2020-03-12
expanded locales; locale and tz environment variables

### LINKED ISSUE #377: Support a wider range of locales, and set LANG/TZ environment variables
For the WPS implmentation (#365), it is important to display datetimes of when processing started and ended. The current Docker image only supports one locale, baked in at build, and does not include any others. This can be confusing for users and providers alike, especially when there's a confusion about month/date order.

It may be possible eventually to display feature-level datetime data in a particular locale, rather than an ISO8601 datetime string - for this there needs to be a way to set and inspect the locale information.

The Docker image should support a large variety of locales, that will allow the Web UI to display dates in whatever locale is specificed in the `LANG` environment variable. If not running from a Docker image, the host locale should be respected rather than ignored.

I have a small PR ready to support this general idea. The resulting image is larger than before, but it does not add much to the build time since it uses the pre-compiled package [`locales-all`](https://packages.debian.org/sid/locales-all).

**Additional context**
- Possibly related to #100.
- Parallel to this is also allowing the `TZ` to be set in an environment variable for the locale-appropriate display of time to end users. My PR also addresses this as a tightly-related locale/environment issue.

**Caveat**
I don't have a Windows machine to test on, and I'm not sure if/how Windows manages locales and timezones as environment variables. I would appreciate a Windows perspective on the `__init__.py` change.

alpha-beta-soup:
Fixes #377


tomkralidis:
@alpha-beta-soup are you able to rebase this PR?  Is the PR still valid?  We should also derive default local values from configuration (`server.language` + `server.encoding`)

alpha-beta-soup:
Yes, I still think this is valid. I'll rebase and add the ability to preferentially set it from configuration.

alpha-beta-soup:
Uhh. messed this one up! Will do this properly and submit a new PR

PR #382, opened 2020-03-16
Fix travis with latest postgis version

francbartoli:
Fix to create postgis extension

Fix postgis 3 package

Add the start of postgresql

Remove standard postgresql from pgdg

Remove all postgresql packages and update

Add autoremove

Declare postgresql version


francbartoli:
Unfortunately it is currently failing because the backend of https://geonode.wfp.org is down

Issue #221, opened 2019-09-02
support STAC specification


tomkralidis:
Implement the [STAC](https://stacspec.org/) API specification to support search/discovery of geospatial assets.  Notes for implementation based on initial discussions with @matthewhanson :

- add front end routes (`/stac/search`)
- update `pygeoapi.api` to address request handling
- implement STAC provider backend via plugin mechanism which will interrogate backend and return results as a Python dictionary for marshalling to JSON proper to the client
- add stubs for transactional capability
- populating a STAC backend could be via workflow beyond pygeoapi (i.e. implement a CLI within the backend, which is not hooked into pygeoapi tooling proper but can be run offline just the same)

pvgenuchten:
I'm curious to see how stac, ogcapi-coverage and ogcapi-records operate together on a single endpoint, what aspects they can share and where are the challenges, would be good input for the upcoming sprint

francbartoli:
> Implement the [STAC](https://stacspec.org/) API specification to support search/discovery of geospatial assets. Notes for implementation based on initial discussions with @matthewhanson :
> 
> * add front end routes (`/stac/search`)
> * update `pygeoapi.api` to address request handling
> * implement STAC provider backend via plugin mechanism which will interrogate backend and return results as a Python dictionary for marshalling to JSON proper to the client

Are you considering external libraries for marshalling? Or we need to implement our own?

> * add stubs for transactional capability
> * populating a STAC backend could be via workflow beyond pygeoapi (i.e. implement a CLI within the backend, which is not hooked into pygeoapi tooling proper but can be run offline just the same)

tomkralidis:
> Are you considering external libraries for marshalling? Or we need to implement our own?

Good point.  I'm guessing a STAC backend could provided via one of the [sat-utils](https://github.com/sat-utils/sat-stac-landsat) tools (for example) and a STAC backend's mission would be to provide Python `dict`'s of JSON objects back to `pygeoapi.api`, but this remains to be seen/needs to be further tested.

francbartoli:
> > Are you considering external libraries for marshalling? Or we need to implement our own?
> 
> Good point. I'm guessing a STAC backend could provided via one of the [sat-utils](https://github.com/sat-utils/sat-stac-landsat) tools (for example) and a STAC backend's mission would be to provide Python `dict`'s of JSON objects back to `pygeoapi.api`, but this remains to be seen/needs to be further tested.

Ok that makes sense, thanks @tomkralidis.

[...]

PR #389, opened 2020-04-07
add STAC catalog functionality

tomkralidis:
## Overview

This PR implements STAC Catalog functionality (#221) into pygeoapi.  Basic setup:

- this implements STAC **catalog** (STAC API is ongoing in the [stac branch](https://github.com/geopython/pygeoapi/tree/stac))
- STAC catalog is implemented through a pygeoapi FileSystem provider
- configuration connects to a directory tree of geospatial data files (raster or vector) on disk.  Example provider configuration snippet:

```yaml
        provider:
            name: FileSystem
            data: /Users/tomkralidis/Dev/data/gdps
            file_types:
                - .grib2
```
- JSON and HTML responses are supported
- STAC functionality gets triggered via the `/stac` endpoint and binds to any `FileSystemProvider` datasets in the config
- tests in `tests/test_filesystem_provider.py`
- docs will be added to the [docs branch](https://github.com/geopython/pygeoapi/tree/issue-277-documentation)
- a generic breadcrumbs function was added in `pygeoapi.util:create_breadcrumbs` and used as a jinja2 filter.  This was implemented because STAC pathing is arbitirary (vs. OGC API's more fixed `/collections/{collectionId}/items/{itemID}` structure).  Might be useful for other templates
- raster based support is limited to the entire file's properties.  Multiband support could be added as future work
- the provider uses rasterio and fiona (and thus adds dependencies (see `pygeoapi.provider.filesystem:_describe_file`).  Comments/thoughts?  Is it worth it at this point to drop down to the GDAL Python API to remove the dependencies on rasterio/fiona?


jorgejesus:
For what I understood on the yml config `pygeoapi-config.yml` 
Should have out of the box support for a small dataset/example:
```
        provider:
            name: FileSystem
            data: tests/data
            file_types:
                - .geojson
```

tomkralidis:
@jorgejesus good idea.  I added a sample setup to our default config to expose the test data.  In addition:

- fix asset URLs to be absolute
- fix tests
- moved fiona and rasterio into `requirements-provider.txt` so that STAC can still work (albeit without geospatial data description) as required

tomkralidis:
FYI test failures are unrelated.

Issue #277, opened 2019-10-15
update documentation


tomkralidis:
Given the awesome work by @jorgejesus to implement project documentation, it's a good time to update docs to provide further user support.

- [x] establish table of contents
- [x] develop first draft of updates for review and comment/refinement

tomkralidis:
Suggest ToC for review/comment:

- project overview
  - features
- installation
   - pip
    - git
    - docker
- configuration
- administration/publishing
  - OpenAPI
- running
  - wsgi
  - asgi
- plugins
  - overview
  - formatters
  - providers
  - processes
- Python API documentation
- support
- contributing
- license

jorgejesus:
The TOC content sounds good

justb4:
ToC looks good, possibly need a config reference section/annex, in time with a formal config schema? Or fine in config chapter itself.
At least in configuration chapter we should point to more examples, e.g. from unit testing and demo.

justb4:
* installation: also `conda`?
* chapter on https://demo.pygeoapi.io setup as example (Docker proxy-ing etc), instances and CITE?

tomkralidis:
Thanks @justb4 I've added a deployment section to the above, as well as an OGC CITE compliance section, which should cover our CITE support/story/workflow (including how to test).

justb4:
section suggestion: "Further Reading/Info": with links to presentations (slides, video) and..

tomkralidis:
Implemented in #392.

PR #392, opened 2020-04-10
Issue 277 documentation

tomkralidis:
Addresses #277 

To test/view:
```bash
git clone https://github.com/geopython/pygeoapi.git -b issue-277-documentation
cd issue-277-documentation
pip3 install sphinx
cd docs
make html
cd build/html
python3 -m http.server
```

Notes:
- areas marked with ``.. todo::`` need help


docs/source/installation.rst
```diff
@@ -0,0 +1,90 @@
...
+Install
+=======
+
+pygeoapi is easy to install on numerous environments.  Whether you are a user, administrator or developer, below
+are multiple approaches to getting pygeoapi up and running depending on your requirements.
+
+Requirements and dependencies
+-----------------------------
+
+pygeoapi runs on Python 3.
```
doublebyte1, line 12: pygeoapi won't build on python < 3.11 . We should indicate this on these instructions or else `python3 setup.py install` will fail.

docs/source/installation.rst
```diff
@@ -0,0 +1,90 @@
...
+   cd pygeoapi
+   . bin/activate
+   git clone https://github.com/geopython/pygeoapi.git
+   cd pygeoapi
+   pip install -r requirements.txt
+   python setup.py install
+   cp pygeoapi-config.yml example-config.yml
+   vi example-config.yml
+   export PYGEOAPI_CONFIG=example-config.yml
+   export PYGEOAPI_OPENAPI=example-openapi.yml
```
doublebyte1, line 33: `example-openapi.yml` is not provided which will cause `pygeoapi serve` to fail. In order to get pygeoapi running, we should create at least an empty file : `echo > " " > example-openapi.yml`

Issue #393, opened 2020-04-11
update configuration to be resource specific


tomkralidis:
As discussed on [Gitter](https://gitter.im/geopython/pygeoapi?at=5e8db6df16f84f0461649cb4), currently we have the following setup for data/process definitions:

```yaml
datasets:
    obs:
       ….
processes:
   hello-world:
      ….
```

We should, instead, consider everything as a resource and introduce a type. Example:

```yaml
resources:
    obs:
        type: vector
        ….
    hello-world:
        type: process
        ….
```

This might be better in the long run as we add support for coverages and metadata, etc.

This would be a breaking change, but pre 1.0 is a great time to iron out things like this.
this is also in line with the ‘resource based architecture’ efforts of APIs in general.

tomkralidis:
Note #397 implements this functionality with the following valid resource types:
- `collection`
- `process`
- `stac-collection`

PR #397, opened 2020-04-13
update configuration to be resource specific (#393)

tomkralidis:
Implements #393.

### DIFF
--- a/aws-lambda/pygeoapi-serverless-config.yml
+++ b/aws-lambda/pygeoapi-serverless-config.yml
@@ -81,8 +81,9 @@ metadata:
         instructions: During hours of service.  Off on weekends.
         role: pointOfContact
 
-datasets:
+resources:
     obs:
+        type: collection
         title: Observations
         description: My cool observations
         keywords:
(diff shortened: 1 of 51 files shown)


tomkralidis:
Note: Travis failure is unrelated.

@francbartoli / @kalxas / @justb4 / @jorgejesus / @pvgenuchten given this is a breaking change can folks approve this PR (either by approving in GitHub or with a +1 here).

aws-lambda/pygeoapi-serverless-config.yml
```diff
@@ -81,8 +81,9 @@ metadata:
         instructions: During hours of service.  Off on weekends.
         role: pointOfContact
 
-datasets:
+resources:
     obs:
+        type: collection
```
pvgenuchten, line 86: Isn't the term collection too generic here? Don't we need a term like "collection-of-vector-features", "collection-of-records", "collection-of-coverages", etc? 
Do we actually require a collection to contain similar things, or can a collection contain both vector-features, records and coverages?

pvgenuchten: as discussed in [gitter](https://gitter.im/geopython/pygeoapi?at=5ea34a52e5ed621d4ddb292c) we can delegate to identify the type of assets in the collection to the backend, and if at some point we need to know the type-of-things in the collection upfront, we can add a another parameter for it. @tomkralidis: "we could also think about adding a type key to the provider. Like a Vector tile provider, OAPIF provider, etc."

Issue #399, opened 2020-04-13
Errors in OpenAPI Document


justb4:
**Description**
The OpenAPI HTML service pages gives errors on page and in JS console

**Steps to Reproduce**
* go to https://demo.pygeoapi.io/master (latest master version)
* select OpenAPI doc as HTML: https://demo.pygeoapi.io/master/openapi?f=html
* select items from any collection, e.g. https://demo.pygeoapi.io/master/openapi?f=html#/obs/get_collections_obs_items
* observe error-emojis in page and errors in JS console (see screenshot below)
 
**Expected behavior**
No errors and texts for I think `properties` (attr filtering) and `bbox` filtering

**Screenshots/Tracebacks**
![image](https://user-images.githubusercontent.com/582630/79117695-1c2fa780-7d8c-11ea-8863-9721f069a4b1.png)

**Environment**
 - OS: any
 - Python version: 3.7
 - pygeoapi version: master as on april 13, 2020

**Additional context**
N.A.

tomkralidis:
This appears to be the same problem as #312 (root cause https://github.com/opengeospatial/ogcapi-features/issues/276).  I have reached out again on the OGC ticket.

Until then, anyone running pygeoapi via HTTP will have a broken Swagger UI; options:

1. Use `server.ogc_schemas_location` as per https://docs.pygeoapi.io/en/latest/configuration.html#server and reference locally
2. copy YAML schemas tree into codebase and reference locally (we do this in pycsw, repo size obviously increases but stable/locked down)
3. other options?

tomkralidis:
For option 1:

```bash
mkdir /tmp/schemas.opengis.net && cd $_
curl -O http://schemas.opengis.net/SCHEMAS_OPENGIS_NET.zip  && unzip SCHEMAS_OPENGIS_NET.zip
# edit pygeoapi config to point `server.ogc_schemas_location` to `/tmp/schemas.opengis.net`
# regenerate OpenAPI doc.
```

justb4:
Q: for option 1, which (http) server actually serves the files? pygeoapi itself?  saw, think api.py,  that the config url is prepended to the unpacked file location..is an external apache server assumed?

Other option is a schemas.pygeoapi.io server with yaml files served from dedicated GitHub repo with that name using a CNAME. That includes https support. Or am I missing something completely?

tomkralidis:
[...]

PR #405, opened 2020-04-14
include local setup of OGC schemas for Docker

tomkralidis:
This pull request implements local OGC schemas as part of pygeoapi's default Docker image.

The Docker setup installs the http://schemas.opengis.net schemas .zip file into `/schemas.opengis.net` in the Docker image and updates the configuration to use local schemas.

Note: any downstream Docker image that uses `pygeoapi:latest` and wants to leverage the local schemas needs to update their config as per `docker/default.config.yml` in this PR.
Fixes: #399

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 #          Just van den Broecke <justb4@gmail.com>
 #          Francesco Bartoli <xbartolone@gmail.com>
 #
-# Copyright (c) 2019 Tom Kralidis
+# Copyright (c) 2020 Tom Kralidis
 # Copyright (c) 2019 Just van den Broecke
 # Copyright (c) 2020 Francesco Bartoli
 #
@@ -62,8 +62,8 @@ ARG ADD_PIP_PACKAGES=""
 # ENV settings
 ENV TZ=${TIMEZONE} \
 	DEBIAN_FRONTEND="noninteractive" \
-	DEB_BUILD_DEPS="tzdata build-essential python3-setuptools python3-pip apt-utils git" \
-	DEB_PACKAGES="locales libgdal26 python3-gdal libsqlite3-mod-spatialite curl ${ADD_DEB_PACKAGES}" \
+	DEB_BUILD_DEPS="tzdata build-essential python3-setuptools python3-pip apt-utils curl git unzip" \
+	DEB_PACKAGES="locales libgdal26 python3-gdal libsqlite3-mod-spatialite ${ADD_DEB_PACKAGES}" \
 	PIP_PACKAGES="gunicorn==19.9.0 gevent==1.5a4 wheel==0.33.4 ${ADD_PIP_PACKAGES}"
 
 ADD . /pygeoapi
@@ -90,6 +90,10 @@ RUN \
 	&& pip3 install -r requirements-dev.txt \
 	&& pip3 install -r requirements-provider.txt \
 	&& pip3 install -e . \
+	# OGC schemas local setup
+	&& mkdir /schemas.opengis.net \
+	&& curl -O http://schemas.opengis.net/SCHEMAS_OPENGIS_NET.zip \
+	&& unzip ./SCHEMAS_OPENGIS_NET.zip "ogcapi/*" -d /schemas.opengis.net \
 	# Cleanup TODO: remove unused Locales and TZs
 	&& pip3 uninstall --yes wheel \
 	&& apt-get remove --purge ${DEB_BUILD_DEPS} -y \
@@ -101,4 +105,3 @@ COPY ./docker/entrypoint.sh /entrypoint.sh
 
 WORKDIR /pygeoapi
 ENTRYPOINT ["/entrypoint.sh"]
-
--- a/docker/default.config.yml
+++ b/docker/default.config.yml
@@ -49,6 +49,7 @@ server:
     map:
         url: https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png
         attribution: '<a href="https://wikimediafoundation.org/wiki/Maps_Terms_of_Use">Wikimedia maps</a> | Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
+    ogc_schemas_location: /schemas.opengis.net
 
 logging:
     level: ERROR


justb4:
Testing...

justb4:
Also need to add `unzip` to `DEB_BUILD_DEPS` ENV var above...

justb4:
The Dockerfile still has several errors, at least at these lines:

* `$` var sign gone: https://github.com/geopython/pygeoapi/blob/e065fe07dc8b571eb209109cb8dfe7c9cf335e9b/Dockerfile#L66
* `unzip` removed at this point: https://github.com/geopython/pygeoapi/blob/e065fe07dc8b571eb209109cb8dfe7c9cf335e9b/Dockerfile#L102

Do you have an environment to build? If not this will take too long for trial/error. 

Given the following:

* the schema files add about 150MB to the Docker Image, making it 880MB, while the earlier versions were a few 100 MB, which I would like to go back to, adds also to internet bandwidth-use
* did not realize we get all of OGC's schemas, while we only need a few files
* serving files in Flask has severe security implications

On the other hand, hosting on GitHub Pages:

* hosting on GitHub pages does support CORS `*` I just learned
* simpler setup for users

Just a few simple steps, did the same last week:

* create a CNAME for `schemas.pygeoapi.io` pointing to GH (think: pygeoapi.github.io)  in pygeoapi DNS provider
* create GH repo `geopython/schemas.pygeoapi.io`
* create a `docs` dir in `master` branch with in it OGC Schemas content
* in GH Repo Settings: designate `/docs` as GH Pages there, also define CNAME there and enable https
* https://schemas.pygeoapi.io should now effectively point to the content under `/docs`

That's about it. If you set the CNAME in DNS, I could do the rest (if I have permissions to create GH repo).

tomkralidis:
Once OGC provides an HTTPS equivalent via https://github.com/opengeospatial/ogcapi-features/issues/276 this should solve our requirement at least for linking directly to OGC, in addition to lessening maintenance efforts on our side.

Having said this, this PR covers the use case where a pygeoapi admin wants to deploy the schemas as part of pygeoapi proper (same host/reduced network traffic). I've updated the PR to thin out the OGC schema footprint to only what's required for OGC API.

justb4:
Ok, what about then using http://schemas.opengis.net/ogcapi/ogcapi-features-1-1_0_0.zip ? 

minor: `unzip` should move to DEB_BUILD_DEPS env var [L65](https://github.com/geopython/pygeoapi/blob/8f2e406507bb0b12b622c06909c651d37c840b02/Dockerfile#L65).

tomkralidis:
> Ok, what about then using http://schemas.opengis.net/ogcapi/ogcapi-features-1-1_0_0.zip ?

That would result in only downloading OGC API - Features schemas.  To be future proof we download all the OGC schemas and prune out only `/ogcapi`, so that once the other OGC API schemas are published to http://schemas.opengis.net/ogcapi we are covered as well.

Having said this, I will inquire to OGC about setting up a zip file for all OGC API schemas like http://schemas.opengis.net/ogcapi/ogcapi-schemas.zip 

> minor: unzip should move to DEB_BUILD_DEPS env var L65.

OK, but when I make this change:

```diff
diff --git a/Dockerfile b/Dockerfile
index 6ba50f6..060feb4 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,8 +62,8 @@ ARG ADD_PIP_PACKAGES=""
 # ENV settings
 ENV TZ=${TIMEZONE} \
        DEBIAN_FRONTEND="noninteractive" \
-       DEB_BUILD_DEPS="tzdata build-essential python3-setuptools python3-pip apt-utils git" \
-       DEB_PACKAGES="locales libgdal26 python3-gdal libsqlite3-mod-spatialite curl unzip ${ADD_DEB_PACKAGES}" \
+       DEB_BUILD_DEPS="tzdata build-essential python3-setuptools python3-pip apt-utils git unzip" \
+       DEB_PACKAGES="locales libgdal26 python3-gdal libsqlite3-mod-spatialite curl ${ADD_DEB_PACKAGES}" \
        PIP_PACKAGES="gunicorn==19.9.0 gevent==1.5a4 wheel==0.33.4 ${ADD_PIP_PACKAGES}"

 ADD . /pygeoapi
```

and try building the image, I received the following error:

```bash
$ docker --version
Docker version 19.03.8, build afacb8b
$ docker build -t pygeoapi_test_ogc_schemas:latest .
...
...
/bin/sh: 1: unzip: not found
The command '/bin/sh -c apt-get update 	&& apt-get --no-install-recommends install -y ${DEB_BUILD_DEPS} ${DEB_PACKAGES} 	&& cp /usr/share/zoneinfo/${TZ} /etc/localtime	&& dpkg-reconfigure tzdata && sed -i -e "s/# ${LOCALE} UTF-8/${LOCALE} UTF-8/" /etc/locale.gen 	&& dpkg-reconfigure --frontend=noninteractive locales 	&& update-locale LANG=${LOCALE} 	&& echo "For ${TZ} date=$(date)" && echo "Locale=$(locale)" 	&& python3 -m pip install --upgrade pip 	&& pip3 install ${PIP_PACKAGES} 	&& cd /pygeoapi 	&& pip3 install -r requirements.txt 	&& pip3 install -r requirements-dev.txt 	&& pip3 install -r requirements-provider.txt 	&& pip3 install -e . 	&& pip3 uninstall --yes wheel 	&& apt-get remove --purge ${DEB_BUILD_DEPS} -y 	&& apt autoremove -y  	&& rm -rf /var/lib/apt/lists/* 	&& mkdir /schemas.opengis.net         && mkdir /tmp/schemas 	&& cd /tmp/schemas 	&& curl -O http://schemas.opengis.net/SCHEMAS_OPENGIS_NET.zip 	&& unzip ./SCHEMAS_OPENGIS_NET.zip         && mv ./ogcapi /schemas.opengis.net         && cd /         && rm -fr /tmp/schemas' returned a non-zero code: 127
(pygeoapi) (base) tomkralidis@aruba:~/Dev/pygeoapi/pygeoapi-tomkralidis$ #vi Dockerfile docker --
```

However when I run the same command with unzip on line 66 it works (?)

tomkralidis:
FYI I have squashed/rebased this PR against current master, as well as moving unzip to `DEBIAN_BUILD_DEPS` (and moving the schema unzip **before** the cleanup).

PR #416, opened 2020-04-20
Catch ogr http errors and item not existing

francbartoli:
Catch ogr http errors and item not existing

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -46,7 +46,7 @@ from pygeoapi.log import setup_logger
 from pygeoapi.plugin import load_plugin, PLUGINS
 from pygeoapi.provider.base import (
     ProviderGenericError, ProviderConnectionError, ProviderNotFoundError,
-    ProviderQueryError)
+    ProviderQueryError, ProviderItemNotFoundError)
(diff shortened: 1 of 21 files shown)


tomkralidis:
Looks good @francbartoli . Comments:

- should we rename the class to `ProviderItemNotFound` (will help for support of other standards (i.e. OGC API - Records)?
- should we update the various `get` functions of the various other providers (if they support it)?

francbartoli:
Should it be `ProviderItemNotFoundError` or `ProviderItemNotFound`? The first would be aligned with other names

tomkralidis:
+1

pygeoapi/api.py
```diff
@@ -868,6 +868,13 @@ def get_collection_item(self, headers_, format_, dataset, identifier):
         try:
             LOGGER.debug('Fetching id {}'.format(identifier))
             content = p.get(identifier)
+        except ProviderItemNotFoundError:
```
tomkralidis, line 871: Should this error go **after** the `ProviderConnectionError` below for logical sequence?

justb4:
Running tests in Docker take a long time I have been noticing, could be due to networking/DNS issues...

PR #417, opened 2020-04-21
implement queryables endpoint

tomkralidis:
# Overview
This PR implements `/collections/{collectionId}/queryables` endpoint as per https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/standard/clause_6_filter.adoc#queryables and https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/standard/requirements/filter/REQ_get-queryables-response.adoc

Note this functionality begins the work for OGC API - Records work and CQL (which will be used by OARec and OAFeat).

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -505,10 +505,18 @@ class API(object):
         }
 
         for k, v in p.fields.items():
-            queryables['queryables'].append({
-                'queryable': k,
-                'type': v['type']
-            })
+            show_field = False
+            if p.properties:
+                if k in p.properties:
+                    show_field = True
+            else:
+                show_field = True
+
+            if show_field:
+                queryables['queryables'].append({
+                    'queryable': k,
+                    'type': v['type']
+                })
 
         if format_ == 'html':  # render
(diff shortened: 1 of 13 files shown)


tomkralidis:
@alpha-beta-soup wondering if this needs any JSON-LD treatment?

pygeoapi/api.py
```diff
@@ -448,6 +448,78 @@ def describe_collections(self, headers_, format_, dataset=None):
...
+
+        queryables = {
+            'queryables': []
+        }
+
+        for k, v in p.fields.items():
+            queryables['queryables'].append({
+                'queryable': k,
+                'type': v['type']
+            })
```
francbartoli, line 511: Does it make sense to have a control if the set of `queryables` is a subset of `fields` and return an `InvalidPluginError` in case the check fails?

tomkralidis:
@francbartoli 

> As a general comment, are we supporting only csv, geojson and ES for queryables?

The other providers look like they implement `.fields`, which is what `pygeoapi.api.API.get_collection_queryables` looks for from a provider when handling the request

Issue #451, opened 2020-06-10
html-UI for basic search mechanism inside collections


pvgenuchten:
with OARec incoming, searching for items in a collection from the html-ui becomes more relevant. 

![image](https://user-images.githubusercontent.com/299829/84291932-0baa6c00-ab46-11ea-8f8b-4b57aec174d3.png)

Suggestion is to add a search box next to title.

Interesting aspect is if we need to explicitely reference a column to use in search, or can query over multiple columns with  col1 like '%a%' OR col2 like '%b%', but I gladly delegate that to another issue...

subtask of #404

Issue #403, opened 2020-04-13
clarify the status of each data provider 


jorgejesus:
Clarify reach data provider capabity for pagging, property filter, bbox, datetime, determine maintainer and status. Test status of each data provider

tomkralidis:
also:
- add docs on contributing a data provider plugin
- ensure automated tests cover the requirements/policy to help regression testing over time

jorgejesus:
Situation om  27 june 2020 and commit https://github.com/geopython/pygeoapi/commit/7766e2533ce487dc3cef158bc02ae56ccebf6b72


Data provider | properties | resulttype | bbox | datetime | sortby | maintainer
-- | -- | -- | -- | -- | -- | --
csv | YES | ALL | NO | NO | NO | Tom Kralidis
elasticsearch | YES | ALL | YES | YES | YES | Tom Kralidis
geojson | YES | ALL | NO | NO | NO | Matthew Perry
mongo | YES | RESULTS | YES | YES | YES | Timo Tuunanen
ogr | YES | ALL | YES | NO | NO | Just van den Broecke /Francesco Bartoli
postgresql | YES | ALL | YES | NO | NO | Jorge Samuel Mendes de Jesus, Tom Kralidis, Mary Bucknell
sqlite | YES | ALL | YES | NO | NO | Jorge Samuel Mendes de Jesus, Tom Kralidis, Mary Bucknell

PR #458, opened 2020-06-12
Adding a basic query widget to items html

pvgenuchten:
Initial implementation for #451

image

Current limitations:
- does not use the queryables as advertised in /collections/{col}/queryables
- no option to see/cancel the current filters
- only support field=value for now, when CQL is available we can extend it to support other operators such as like, >, <, etc.

Is it correct that filters are not operational on the geojson-backend ?


pvgenuchten:
Any feedback here? people like the approach as a start, any suggestions for improvement?

tomkralidis:
cc'ing @kngai for any UI/UX comments.

Nice work @pvgenuchten . Correct that the native GeoJSON driver doesn't support filters.

We'll need some sort of advertisement from providers on what they support filter wise (see #403)

pygeoapi/templates/items.html
```diff
@@ -16,12 +16,30 @@
...
+            {% endfor %}
+          </select>
+          <select style="height:40px;width:60px">
+            <option>=</option>
+          </select>
+          <input type="text" id="filterfield" value="" />
+          <input type="submit" value="Filter"/>
+        </FORM>
+      </div>
+    </div>
```
kngai, line 41: Excellent stuff here. My only minor suggestions are:
- Place this `<div class="row">` block under line 43 (`{% if data['features'] %}`) so it renders with the table
- Change `<FORM>` to `<form>`

pvgenuchten:
updated the pr to populate the filters from queryables
@kngai I'd suggest to move/improve visuals after merge

pvgenuchten:
replaced by #609

Issue #457, opened 2020-06-12
Google doesn't dereference linked JSON-LD context.


dblodgett-usgs:
**Description**
Google doesn't resolve the geojson-ld context.

**Steps to Reproduce**
This resource from the latest pyGeoAPI: https://demo.pygeoapi.io/stable/collections/lakes/items/1

In the google test-tool:

https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fdemo.pygeoapi.io%2Fstable%2Fcollections%2Flakes%2Fitems%2F1

note Type error

**Expected behavior**
This resource: https://geoconnex.us/ref/hu02/09 from a hacked pygeoapi that has the geojson-ld context inline.

https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fgeoconnex.us%2Fref%2Fhu02%2F09

note no type error

**Additional context**
Just putting the content of the geojson-ld context in line fixes this issue.

alpha-beta-soup:
That's nice and small, should be fine to inline it. But it's a bit disappointing that it won't dereference remote contexts though... what if they change?

dblodgett-usgs:
I think the "what if they change?" issue goes both ways -- having a context disconnected from the file in question introduces the potential that the remote context could change and break the file in question.

pvgenuchten:
schema.org on collection level is correctly picked up, in the test tool

![image](https://user-images.githubusercontent.com/299829/84549730-06e1e580-ad09-11ea-9cbd-866a7de6df88.png)

also in the search console

![image](https://user-images.githubusercontent.com/299829/84549697-eb76da80-ad08-11ea-856c-9bb590136ec0.png)

I can confirm google test tool fails in parsing https://demo.pygeoapi.io/master/collections/dutch_windmills/items/Molens.1, but i've no idea if that is valid json-ld

pvgenuchten:
mmm, results above may be related to #459, we still have 2 implementations of schema.org on the collections page, and indeed the embedded json-ld seems broken

alpha-beta-soup:
So do we want to remove the microdata and leave all the semantic content up to JSON-LD? It is possible to mix them.

dblodgett-usgs:
Note that because the geojson-ld context isn't resolved, google doesn't recognize any of the json-ld as structured data. If I copy in the geojson-ld context I get:
[...]

Issue #433, opened 2020-05-05
Enable custom uris for feature id for geojson providers


ksonda:
**Is your feature request related to a problem? Please describe.**
I would like to have the id:/ @id: field in the json-ld content for feature collections generated from geojson providers and their constituent items have custom uri (as provided by an external PID service). 

For example: the persistent identifier [https://geoconnex.us/ref/nat_aq/N100BSNRGB](https://geoconnex.us/ref/nat_aq/N100BSNRGB) redirects to [https://info.geoconnex.us/collections/nat_aq/items/N100BSNRGB](https://info.geoconnex.us/collections/nat_aq/items/N100BSNRGB)

In the json-ld representation of this item [https://info.geoconnex.us/collections/nat_aq/items/N100BSNRGB](https://info.geoconnex.us/collections/nat_aq/items/N100BSNRGB?f=jsonld), there is

```
id: "https://info.geoconnex.us/collections/nat_aq/items/N100BSNRGB"
```

I would like this to be the persistent identifier instead.

**Describe the solution you'd like**
One possible solution would be to allow for a ```uri_field:``` parameter in the configuration .yml, which if provided, would override the json-ld generation occuring around line 178 of [linked_data.py](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/linked_data.py) with the entry in the desingated field of the geojson file. 

Of course, this could be generalized to other provider types as well.

Ideally, this would propogate to everywhere in the API and html templates wherever "id" has a hyperlink pointing to its items.html instance.

**Describe alternatives you've considered**
Another solution would be to allow for a ```uri_stem:``` parameter for a resource that would replace ```{}/collections/``` and remove ```items/``` from the ```id:``` generated in linked_data.py. For example, I could specify:

```
resources:
    nat_aq:
        type: collection
        title: National Aquifers
        description: National Aquifers of the United States
        uri_stem: https://geoconnex.us/ref/nat_aq/

```

alpha-beta-soup:
Without commending on implementation details, I'm totally in agreement that:
- the @id should default to being a link to the item (i.e. a direct link to the pygeoapi instance)
- it should be possible to specify a property that can override this (as a literal and maybe as a template?) to support PIDs as well as the semantic meaning of @id

This is what I should have implmented to begin with, I struggled with what to do with @id by default.

tomkralidis:
Can this be closed given #676 ?

ksonda:
Yep!

Issue #384, opened 2020-03-20
Support omitting geometry to enable smaller payloads and faster retrieval


alpha-beta-soup:
**Is your feature request related to a problem? Please describe.**

Sometimes a client doesn't actually care about a geometry, but the properties alone. To reduce the payload in such a use-case, it could be useful to allow clients to express that they want to omit a geometry altogether.

As discussed in https://github.com/geopython/pygeoapi/issues/363, we might also like to have an alternative non-WGS84 geometry representation as a GeoJSON `property` (WKT string or other value), and then the ability to disable the `geometry` allows us to still keep payloads reasonably small, particularly for large and complex geometries.

**Describe the solution you'd like**

Some URL parameter would need to be chosen to allow a client to express this preference, probably a Boolean `geom=false` — this is a very loose suggestion and I eagerly await better ideas. The default behaviour should of course be to include the geometry.

The [GeoJSON spec](https://geojson.org/geojson-spec.html#feature-objects) says:

> A GeoJSON object with the type "Feature" is a feature object.
>
> A feature object must have a member with the name "geometry". The value of the geometry member is a geometry object as defined above or a JSON null value.

So the `geometry` property would still need to be included in any GeoJSON responses, but it would have the value of `null`.

An exploration of how this should be handled in (Geo)JSON-LD should be carried out—we do not want to make a _positive statement_ that a geometry *is* null or empty (unless it actually is), but rather that it is simply absent, not-represented, or undefined in this case.

Other output types would need to be considered too.

**Describe alternatives you've considered**

The alternative is to *not* support this and instead always include the geometry. The drawback is simply a larger payload, slower download, and also potentially wasted server time retrieving/joining/reprojecting geometries that will be ignored.

**Additional context**

Ideally this is implemented more deeply than simply lopping off the geometry at the last moment, but actually results in geometries being totally ignored when requests are made to underlying data stores, unless there's a spatial query (e.g. a bbox filter) being simultaneously applied.

francbartoli:
@alpha-beta-soup the feature sounds reasonable. Regarding the implementation I'm wondering if it's just a matter of the requested `Content-Type`:

```text
/items
/items/{item_id}
application/geo+json --> return GeoJSON
application/json          --> return JSON encode properties 
```

What do you think?

alpha-beta-soup:
[...]

Issue #444, opened 2020-06-04
JSON LD errors in Travis


tomkralidis:
**Description**
For some reason, running tests against `tests/test_api.py` yield errors within the JSON LD tests.

**Steps to Reproduce**
`py.test tests/test_api.py`

**Expected behavior**
Passing tests/no errors

**Screenshots/Tracebacks**
https://travis-ci.org/github/geopython/pygeoapi/jobs/694768368

**Environment**
 - OS: any
 - Python version: 3.6/3.7
 - pygeoapi version: master

**Additional context**
None

alpha-beta-soup:
I suspect it's related to this https://github.com/schemaorg/schemaorg/issues/2578#issuecomment-632227864 (ref. https://github.com/digitalbazaar/pyld/issues/105), i.e. that following a DOS attack, https://schema.org/ was reconfigured and now only returns HTML, rather than the JSON-LD it was once possible to obtain with content negotation at that URI.

Kind of hits at a sore spot for linked data, i.e. the assumption that arbitrary remote URLs will resolve reliably.

> While I regret that this has been inconvenient to various folk, maybe it can help kickstart discussion of robustness in JSON-LD parsing? It is not ideal to have software processes which assume 100% perfect access to remote networked services. At Schema.org we expressed concerns back in 2014 (/cc @rvguha) that
>
> > [...] we do have concerns about the
schema.org website forming an integral part of numerous unknown
software systems and applications. It ought to be possible to do
useful things with schema.org-based json-ld without a dependency on
the Web site.

Fortunately in this case it should just mean updating those links to use https://schema.org/docs/jsonldcontext.jsonld

alpha-beta-soup:
cf https://github.com/digitalbazaar/pyld/issues/128

This keep causing me grief. Tests (with pyld) only work with the full `.jsonld` URL, but the Google Structured Data Testing tool will only work with `https://schema.org`. I think I'd now prefer to fix it upstream than to work around it.

tomkralidis:
Any update on this one?  Should we close?

PR #465, opened 2020-06-24
Id field and geometry json

gpetrochenkov-usgs:
This pull request was made to solve the following issues:

https://github.com/geopython/pygeoapi/issues/433
- If an identifier and URI are available they are used as the @id.
- The the link in the id field used in table on the items page now links to a URI if that is available to address google indexing the URL as opposed to the URI.

https://github.com/geopython/pygeoapi/issues/384#issuecomment-624978617
- Changed the item JSON-LD geometry attribute to link to the GeoJSON representation as a blank node.

https://github.com/geopython/pygeoapi/issues/457
- Add default GeoJSON vocabulary to context in item JSON-LD, (defaults stored in metadata of pygeoapi config)
- Changed the general semantics of the JSON-LD to: "@context": {
    "schema": "https://schema.org/"
    } and prepending "schema" to relevant keys and attributes

All the tests pass.  Let me know if you have any questions or comments.


ksonda:
There are a few commits from my fork that you don't want in here, particularly modifications to logos, and to root.html

alpha-beta-soup:
This looks great. The `@id` changes in particular are really important.

I was actually just working to refactor some of the JSON-LD generation, so that it worked as a single decorator rather than having one decorator function for some endpoints, and another function that transformed generated collection/feature JSON to JSON-LD. I'd like to see if I can reconcile these two and produce an alternative PR that incorporates both sets of changes; I can do this in the next day, although my main source of delay is that I've had failing tests: see #444 - I'm surprised that your change to the `@context` means tests are passing, I suppose pyld expansion algorithm doesn't try to fetch the remote context in that case, which is why the tests are failing otherwise.

A few comments specifically on this PR:
- You've accidentally committed some USGS assets that should probably remain in your fork. This includes both static assets and changes to the UI/templates.
- A `default_vocabulary` is maybe useful, but I think I'd prefer to have the GeoJSON voabulary hardcoded for collections and features, since it's not really something that can be changed, it's a mandatory description of the GeoJSON interface.

pygeoapi/linked_data.py
```diff
@@ -201,5 +215,8 @@ def geojson2geojsonld(config, data, dataset, identifier=None):
                 feature['id'] = featureId
             else:
                 feature['id'] = '{}/{}'.format(data['id'], featureId)
+    else:
+        jsonld_data["geometry"] = {"schema:encodingFormat": "application/geo+json",
+                     "schema:url": data['id'] + '?f=json'}
```
alpha-beta-soup, line 220: Is this really correct? Linking to the JSON representation of itself isn't really the `geometry`, since it would be the entire object, including the properties that were already transported in the JSON-LD response. Perhaps pygeoapi needs an alternate endpoint or query paramter (i.e. `geom=only` or something...) that is *just* geometry for a collection or feature, so this will work as expected.

I think it's also worth discussing whether this should be an option alongside including the geometry in-line, and then which behaviour should be the default.

pygeoapi/linked_data.py
```diff
@@ -174,20 +177,31 @@ def geojson2geojsonld(config, data, dataset, identifier=None):
...
+
     context = config['resources'][dataset].get('context', [])
-    data['id'] = (
-        '{}/collections/{}/items/{}' if identifier
-        else '{}/collections/{}/items'
-    ).format(
-        *[config['server']['url'], dataset, identifier]
-    )
+
+    uri = data['properties'].get('uri') if 'properties' in data else None
```
alpha-beta-soup, line 183: Will need to add documentation that `uri` is a magic property that has this effect; and probably allow configuration to override it.

alpha-beta-soup: Actually, I wonder if we can better hook into the configuration `provider['id_field']`.

pygeoapi/linked_data.py
dblodgett-usgs: This is just a problematic aspect of converting GeoJSON into linked data. It's certainly not valid JSON-LD to include the geometry in line. It's also not really valid to link directly to a non-semantic resource as the object of a triple. This is a compromise that I'm really not sure is worth pursuing. Just dropping the geometry in the JSON-LD representation is probably fine, IMHO.

alpha-beta-soup: OK, well it should probably have some kind of link to the complete GeoJSON representation then. Perhaps later on we can add logic for computing a https://schema.org/GeoShape, or convert to WKT.

For now, would you be happy if we included the geometry in-line if it's a Point, since that is valid JSON-LD, and omitting it otherwise?

dblodgett-usgs: Yes -- if point, I think it's a good idea to leave it in. For other geometry types, I think the best way to do it is to have a blank node with a link to the complete feature collection up at the root of the document. 

There's already the "rel": "alternate" that points to the geoJSON representation, so we are kind of covered. I'm not seeing a context for the links JSON keys witha little googling, maybe that exists?

gpetrochenkov-usgs:
Good Morning Everybody,  thanks for the feedback!  I will make revisions, roll back the commits we don't want, and push up a new build shortly.

pygeoapi/linked_data.py
gpetrochenkov-usgs: I think it's a good idea to have this be a part of the config in some way, perhaps something akin to:
image
However, since you have the intention of doing a PR merging some of this, I probably want to finish sooner than later as the above warrants more discussion.  

Would documentation done as follows suffice or is there a preferred way to document this property:
image

gpetrochenkov-usgs: With better English, ha.
image

ksonda:
To be very rigorous about this, the URI should also populate the links of items. In `api.py`: 
```
  fcm['links'] = [{
              'rel': 'self' if not format_ or
              format_ == 'json' else 'alternate',
              'type': 'application/json',
              'title': 'This document as JSON',
              'href': '{}?f=json'.format(self.config['server']['url'])  ##if an item and has 'uri', should be 'uri'
            }, {
              'rel': 'self' if format_ == 'jsonld' else 'alternate',
              'type': 'application/ld+json',
              'title': 'This document as RDF (JSON-LD)',
              'href': '{}?f=jsonld'.format(self.config['server']['url']) ##if an item and has 'uri', should be 'uri'
            }, {
              'rel': 'self' if format_ == 'html' else 'alternate',
              'type': 'text/html',
              'title': 'This document as HTML',
              'href': '{}?f=html'.format(self.config['server']['url']), ##if an item and has 'uri', should be 'uri'
              'hreflang': self.config['server']['language']
            }, {
              'rel': 'service-desc',
              'type': 'application/vnd.oai.openapi+json;version=3.0',
              'title': 'The OpenAPI definition as JSON',
              'href': '{}/openapi'.format(self.config['server']['url'])
            }, {
              'rel': 'service-doc',
              'type': 'text/html',
              'title': 'The OpenAPI definition as HTML',
              'href': '{}/openapi?f=html'.format(self.config['server']['url']),
              'hreflang': self.config['server']['language']
            }, {
              'rel': 'conformance',
              'type': 'application/json',
              'title': 'Conformance',
              'href': '{}/conformance'.format(self.config['server']['url'])
            }, {
              'rel': 'data',
              'type': 'application/json',
              'title': 'Collections',
              'href': '{}/collections'.format(self.config['server']['url'])
            }
        ]
```

gpetrochenkov-usgs:
> To be very rigorous about this, the URI should also populate the links of items, and the urls in the script header in `base.html` should also be the URI (in this example below, https://geoconnex.us/ref/hu02/07?f=html). This screenshot is from my fork that has this PR merged.
> 
> image

I went ahead and moved the uri link creation to the following spot, now the canonical link is the uri:
image

I'm going to test a little bit and push up a new revision soon.

pygeoapi/linked_data.py
alpha-beta-soup: By documentation, I meant the user-facing configuration documentation in `/docs/source/configuration.rst`. As part of the merge I'm looking at elevating it be part of the configuration, so I can do this part too.

pygeoapi/linked_data.py
```diff
@@ -180,26 +182,59 @@ def geojson2geojsonld(config, data, dataset, identifier=None):
...
+         },
+         "features": {
+             "@container": "@set",
+             "@id": "geojson:features"
+         },
+         "geometry": "geojson:geometry",
+         "id": "@id",
+         "properties": "geojson:properties",
+         "type": "@type",
+         "description": "http://purl.org/dc/terms/description",
```
alpha-beta-soup, line 228: What's the reason these are included?

alpha-beta-soup: these = `description` and `title`

gpetrochenkov-usgs:
Quick general question: Does anyone know why I would be passing this test locally but it would fail on a travis build?

image

pygeoapi/linked_data.py
alpha-beta-soup: Can you please provide an example of what the non-point GeoJSON-LD should look like with the blank node? I was just planning to drop the `geometry` field; not sure if that's what you had in mind.

The `links` will be dropped in the GeoJSON-LD output as well, at least until someone proposes what these should be mapped to in some vocabulary.

pygeoapi/linked_data.py
gpetrochenkov-usgs: Those two properties were a part of the vocabulary resolved by the original link.  We copied the whole vocab, should I omit these two properties?

alpha-beta-soup:
No, and it doesn't seem related to your changes. That test actually makes an API call to a third-party endpoint that is not mocked. Could be some network issue on the travis server, possibly even intermittent. Might be worth raising as a separate issue, as it's generally not a great idea to have a test depend on an external service.

pygeoapi/linked_data.py
gpetrochenkov-usgs: https://geojson.org/geojson-ld/geojson-context.jsonld

ksonda: https://geojson.org/geojson-ld/geojson-context.jsonld

pygeoapi/linked_data.py
gpetrochenkov-usgs: Sure thing, and I agree dropping the geometry field is a valid approach as well.  Here is what it looks like:

image
 
The JSON is not a direct mapping, although the type is Feature it's mapping to geometry which is understandably not ideal.

pygeoapi/linked_data.py
gpetrochenkov-usgs: Ok good to know, going forward I'll make sure to update the rst docs.

pygeoapi/linked_data.py
alpha-beta-soup: I don't see those properties there? That context ends at `type`, the one in your commit continues with `description` and `title`?

gpetrochenkov-usgs:
> No, and it doesn't seem related to your changes. That test actually makes an API call to a third-party endpoint that is not mocked. Could be some network issue on the travis server, possibly even intermittent. Might be worth raising as a separate issue, as it's generally not a great idea to have a test depend on an external service. (Or maybe it doesn't... I'm not really sure of the intention of that test...)

Good idea, I'll go ahead an open an issue.  Thanks for clarifying that.

pygeoapi/linked_data.py
alpha-beta-soup: Ah I see, at https://geojson.org/geojson-ld/ you can see a full exmaple, but the description and title parts are included since those are need for the example's `properties`. No, don't include those. Any additional context is expected to be specified in the configuration YAML.

gpetrochenkov-usgs: So locally this test fails in "test_api.py"  "test_get_collection_items_json_ld" without @type:
image

Should I adjust this assertion?  The rest of the test works fine.

gpetrochenkov-usgs: Never mind my last comment, ha.

Issue #459, opened 2020-06-12
Remove duplicated schema.org


pvgenuchten:
Current master contains 2 approaches for adding schema.org, via an embedded json-ld, and as microdata. Somehow the two are not conflicting, maybe because the embedded json-ld throws an error?

![image](https://user-images.githubusercontent.com/299829/84551197-25e27680-ad0d-11ea-9a98-536ea04c81cf.png)

On the microdata part (maybe not relevant anymore) I noticed some validation warnings on schema.org/dataset

![image](https://user-images.githubusercontent.com/299829/84550563-46113600-ad0b-11ea-9528-a039e7ff0f74.png)

contentUrl is not provided in the proper way here...

and a warning: license information is missing

also tried a collection record on the yandex validator

![image](https://user-images.githubusercontent.com/299829/84551820-0f3d1f00-ad0f-11ea-80a7-4e5c005646a8.png)

which translates to "ERROR: The content property of the meta tag cannot contain a link"

this is an interesting one

![image](https://user-images.githubusercontent.com/299829/84551960-8a063a00-ad0f-11ea-89bb-e218603d1548.png)

translates to "ERROR: the contentURL field is not defined in the https://schema.org/DataDownload specification. Perhaps you meant contentUrl"

alpha-beta-soup:
Ah cool, didn't even know there was a Yandex validator too.

Issue #453, opened 2020-06-10
remove duplicated id


pvgenuchten:
This line displays the (internal) id of the feature

https://github.com/geopython/pygeoapi/blob/917f1c0a959ca9962ceedfcca04a08b9e1298418/pygeoapi/templates/items.html#L87

Which in most/all cases is the same value as the field defined in 

https://github.com/geopython/pygeoapi/blob/917f1c0a959ca9962ceedfcca04a08b9e1298418/pygeoapi-config.yml#L117

Which causes a duplication of the id column.

![image](https://user-images.githubusercontent.com/299829/84317377-f09f2280-ab6c-11ea-8406-e486aaf4f4d7.png)

I'm wondering should we remove this line

https://github.com/geopython/pygeoapi/blob/917f1c0a959ca9962ceedfcca04a08b9e1298418/pygeoapi/templates/items.html#L87

Or only show it conditionally for certain providers or if id_field is not provided in

https://github.com/geopython/pygeoapi/blob/917f1c0a959ca9962ceedfcca04a08b9e1298418/pygeoapi-config.yml#L117

alpha-beta-soup:
After looking at https://info.geoconnex.us/collections/hu02/items/09, @dblodgett-usgs it would be good to get your opinion on how IDs (especially HTTP URI IDs) should be treated.

dblodgett-usgs:
I don't think the id field should be duplicated and agree that the id field should have a hyperlink on it that gets you to the item. That issue should be treated separately from canonical URI IDs but there is some overlap in that you may want to override the hyperlink with a URI.

We have this issue: https://github.com/geopython/pygeoapi/issues/433 to handle the canonical URI use case. Thre's a linked issue discussing the issue a bit more here: https://github.com/internetofwater/geoconnex.us/issues/19

@gpetrochenkov-usgs is working on fixing up some of the hacks we put into this fork of pygeoapi: https://github.com/ksonda/pygeoapi

alpha-beta-soup:
![Screenshot from 2020-06-26 13-36-58](https://user-images.githubusercontent.com/5925045/85811431-42bd8600-b7b2-11ea-86c1-9a5a16d8c169.png)
![Screenshot from 2020-06-26 13-37-12](https://user-images.githubusercontent.com/5925045/85811430-418c5900-b7b2-11ea-940c-2e00652b5169.png)

This issue is related to #433, so I'm trying to hit both birds with one stone. Above are some screenshots of my local development instance.

[...]

PR #469, opened 2020-06-26
Meta review

alpha-beta-soup:
Includes changes from #465 

Attempts to address:
- #384 (only partially - geometry is mandatorily omitted for GeoJSON-LD payloads with non-point geometry)
- #433 (`uri_field` is obtained from configuration, to support custom URIs for features, which determine JSON-LD `@id` at the level of an individual feature)
- #453 (in the UI, there is now a URI property in item tables, that (by default) links to the Pygeoapi item itself, or to a URI from a property (intended to support PIDs) 
- #457 (inlines the GeoJSON-LD context; addresses problems with https://schema.org ignoring content negotiation and not returning JSON-LD `@context` data)
- #459 (removes schema.org microdata; leaves JSON-LD as the only method of describing embedded meta/data)

I appreciate that this is a lot to pick off in a single merge, but these issues were genuinely quite intertwined, and changes to support the custom ID require careful examination of how the feature IDs are treated in the codebase.

I'd really appreciate if @gpetrochenkov-usgs, @ksonda and @dblodgett-usgs could check out this branch and use it with your sample data to see if it still meets all the requirements you have for linked data support in pygeoapi.

I also took the opportunity to do some refactoring. In particular, all JSON-LD responses are now obtained with the use of the `@jsonldify` function decorator, which inspects function names as strings and does appropriate work. This allows the decorator to either do some pre-computation before the decorated function runs (e.g. to generate `API.fcmld`) or to transform the decorated function output (e.g. to transform GeoJSON data, which may have been filtered, to GeoJSON-LD), or to do a no-op for functions where the appropriate JSON-LD mapping has not been determined (e.g. for STAC).

All tests pass.


ksonda:
Thanks much @alpha-beta-soup and @gpetrochenkov-usgs !  The docker image internetofwater/pygeoapi-geoconnex:context mod incorporates this PR and an example deployment is at https://wade-test.geoconnex.us Everything appears to behave as we want it to. @dblodgett-usgs can confirm.

dblodgett-usgs:
I am stoked. Thanks everyone. This looks good enough for government work. 

If there's one thing that ended up a little unfortunate it's the repeated uri/id property as in 
https://wade-test.geoconnex.us/collections/WaDE/items/NM_4488?f=jsonld and https://wade-test.geoconnex.us/collections/WaDE/items/NM_4488?f=html

I'm not sure it matters enough to mess with, but figured I'd mention it.

ksonda:
was some version of this ever merged? I cant get the master branch image to have this functionality.

alpha-beta-soup:
No, this is still open and unmerged, and now needs to be updated to work with the current state of master

dblodgett-usgs:
I've asked @gpetrochenkov-usgs to have a look at getting these conflicts resolved. If and when that happens, we should really get this merged if at all possible.

tomkralidis:
Closing given #676

PR #470, opened 2020-06-26
allows locale to be specified in config or environment variabls

### LINKED ISSUE #377: Support a wider range of locales, and set LANG/TZ environment variables

alpha-beta-soup:
Replaces #378 , fixes #377

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,16 +54,17 @@ LABEL maintainer="Just van den Broecke <justb4@gmail.com>"
 # https://github.com/geopython/demo.pygeoapi.io/tree/master/services/pygeoapi
 
 # ARGS
-ARG TIMEZONE="Europe/London"
-ARG LOCALE="en_US.UTF-8"
+ARG TZ="Etc/UTC"
+ARG LANG="en_US.UTF-8"
 ARG ADD_DEB_PACKAGES=""
 ARG ADD_PIP_PACKAGES=""
 
 # ENV settings
-ENV TZ=${TIMEZONE} \
+ENV TZ=${TZ} \
+	LANG=${LANG} \
 	DEBIAN_FRONTEND="noninteractive" \
 	DEB_BUILD_DEPS="tzdata build-essential python3-setuptools python3-pip apt-utils curl git unzip" \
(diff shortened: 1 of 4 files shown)


alpha-beta-soup:
My head isn't really in this space at present, but it sounds like it's been supplanted? Happy for this to be closed without being merged, if that's appropriate.

tomkralidis:
Thanks @alpha-beta-soup.  Note that the `Dockerfile` in current master has since changed https://github.com/geopython/pygeoapi/blob/master/Dockerfile.  Do we still need the following?

- https://github.com/geopython/pygeoapi/pull/470/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R67
- https://github.com/geopython/pygeoapi/pull/470/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R82


...or do we, at this point, only need to update the default TZ to UTC?

tomkralidis:
@alpha-beta-soup given #711, can we close this PR/remove the branch?

PR #475, opened 2020-06-30
Close #474

francbartoli:
Fix trailing slash in server url

Fix tag not declared in the root mapping key

Fix missing operationId

Add spectral command to travis for linting the openapi document

PR #478, opened 2020-07-01
Openapi cql extension

FarheenB:
Currently pygeoapi supports only selection clauses that are provided by the OGC API Common specification (i.e. bbox and datetime). Enhanced filtering capabilities beyond(bbox and datetime) are not yet implemented in pygeoapi. I as a users would like to retrieve a set of collection based on certain feature query filters made up of either logical or comparison or spatial or temporal expressions. Sometimes I would even want to get a collection satisfying two or more query filters. I would not like to fetch all of the resources everytime. For eg: Get the lakes of the world whose area of geometry is greater than XYZ sq.km. So currently not having a feature filtering functionality on a collection resource is a big problem for pygeoapi, which should be solved.

A fundamental operation performed on a collection of features is that of querying in order to obtain a subset of the collection which contains feature instances that satisfies some filtering criteria. CQL can be used to specify how resource instances in a source collection should be filtered to identify a result set. Typically, CQL is used in query operations to identify the subset of resources that should be included in a response document. Each resource instance in the source collection is evaluated using a filtering expression. The overall filter expression always evaluates to true or false. If the expression evaluates to true, the resource instance satisfies the expression and is marked as being in the result set. If the overall filter expression evaluates to false, the data instance is not in the result set.  

There could be two approaches to solve the above stated problem: 
1. To implement the CQL filter functionality on features at API level of pygeoapi. But it would be difficult to handle the varied output from different providers at one API level.
2. To implement the feature filter extension at provider level. Since some providers support filter extension while some don't, so to write CQL approaches depending on the provider could be an alternative  

**Additional context**
Depending on the feature request raised above I have implemented CQL extension in pygeoapi which follows [OGC API - Features - Part 3: Common Query Language](http://docs.opengeospatial.org/DRAFTS/19-079.html) standards and have generated OpenApi Document for it. The information for the resources' support for CQL filter is read from its config file. For Openapi document generation, I have added certain test cases on resources that supports CQL extension and also for the resources that does not support CQL.  All the test cases are passed when it is tried with different combinations of CQL/ NO CQL resources. I have followed flake8 standards and have added proper docstring for the functions and CQL information along with configuration example in RTD doumentation. The build for CQL branch [openapi-cql-extension](https://github.com/FarheenB/pygeoapi/tree/openapi-cql-extension) is successful with travis.

Following are the screenshots to support my above tasks:
- **Test cases passed:**
image

- **Openapi Document generated:**
JSON doc
image

Collection that supports CQL
image

Collection that does not support CQL
image

CQL Schema
image

- **Travis Build**
image
https://travis-ci.com/github/FarheenB/pygeoapi/jobs/356559046
https://travis-ci.com/github/FarheenB/pygeoapi/jobs/356559047

- **RTD** 
image
https://pygeoapi-cql-filters.readthedocs.io/en/latest/configuration.html

*Updated test result screenshot after restructuring test functions


tests/test_openapi.py
```diff
@@ -45,3 +53,310 @@ def test_str2bool():
...
+    Get the generated OpenAPI 3.0 Document
+
+    :param config: configuration object
+
+    :returns: OpenAPI definition YAML dict
+    """
+    return get_oas_30(config)
+
+
+def test_cql_filters(get_oas_30_, config):
```
francbartoli, line 96: If you look at other tests you would find many functions to assert minimal unit test cases.
I'd rather splitting this big function to several ones in order to follow the principle of single responsibility and group together assertions with the same scope, like `test_cql_filters_logical_expressions` for logical expressions

francbartoli:
I would also wait for the merge of #475 so we can have the linting of the openapi document cc @tomkralidis

docs/source/configuration.rst
```diff
@@ -251,6 +251,40 @@ implementation of JSON-LD structured data is available for any data provider but
 ``@context``.  Relationships between items can be expressed but is dependent on such relationships being expressed
 by the dataset provider, not pygeoapi.
 
+
+CQL Filter
```
francbartoli, line 255: What do you think to start a nested section here in the [openapi](https://pygeoapi-cql-filters.readthedocs.io/en/latest/openapi.html) chapter about CQL filters?

pygeoapi/openapi.py
```diff
@@ -416,6 +426,831 @@ def get_oas_30(cfg):
             }
         }
 
+    # if CQL filter is applicable
+    if cql_filter_exists:
```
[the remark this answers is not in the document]
FarheenB: if the configuration for filters is not set for any of the resources then **"/queryables"** (L431) wont be available(am still waiting for your decision on this) but **"collections/{collectionId}/queryables"** (L354) is available for all the resources irrespective of the CQL configuration(as it was before, I didn't apply any condition for it)

docs/source/configuration.rst
FarheenB: In  [openapi](https://pygeoapi-cql-filters.readthedocs.io/en/latest/openapi.html) chapter like **using OpenAPI section** I can have another section for **using CQL filters** or even inside  **using OpenAPI section** I can put in Swagger screenshots related to CQL with proper descriptions. I can do that. What do you say?

tests/test_openapi.py
FarheenB: I will split this big function into multiple smaller functions related to each type of expression assertion.

FarheenB: restructured test functions with proper flake8 and linting. please check https://github.com/geopython/pygeoapi/pull/478/commits/c53877cd96c83825b43a19a219b3308daeccf4a0

FarheenB: @francbartoli @jorgejesus give a look to this https://github.com/FarheenB/pygeoapi/blob/openapi-cql-extension/tests/test_openapi.py for the above issue

docs/source/configuration.rst
FarheenB: @francbartoli @jorgejesus thoughts on this?

tests/test_openapi.py
```diff
@@ -93,93 +95,206 @@ def get_oas_30_(config):
...
+    for k, _ in collections.items():
+        if 'filters' in collections[k]:
+            return True
+    return False
 
-    assert isinstance(get_oas_30_, dict)
 
-    # assertion for get paths
+@pytest.fixture()
+def cql_components(get_oas_30_):
```
francbartoli, line 116: I don’t think it makes sense to split the configuration into different fixtures. We agreed to split just the assertions into several functions for unit tests

FarheenB: Depending on the structure of unit test functions I tried to combine the configurations as much as I can. https://github.com/geopython/pygeoapi/pull/478/commits/27c7c7794117e558fe5f01a99e23ec9d3f35e628 for your reference.

docs/source/configuration.rst
jorgejesus: @FarheenB @francbartoli  I suggest for the openAPI section to be done as documentation on  GSoC-2, since at the end of GSoC-2 it will be very clear the status of the implementation and we will have (POST for filter-json) and what we can support on filter-text. If you agree I will open a ticket for this, also for  @alex-mathew

tests/test_openapi.py
```diff
@@ -45,3 +56,452 @@ def test_str2bool():
...
+
+    :param config: configuration object
+
+    :returns: OpenAPI definition YAML dict
+    """
+    return get_oas_30(config)
+
+
+@pytest.fixture()
+def is_cql(config):
```
jorgejesus, line 99: @FarheenB  liked very much that you used `is_cql` as fixture

pygeoapi/openapi.py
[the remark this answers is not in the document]
jorgejesus: @FarheenB Leave the higher level resource  "/queryables" (L431), I understand the logic but there are extra componentes (metdata records) that require it.  Leave it for now, this topic should be better checked.

@francbartoli   Trying to find some docs on this section "collections/{collectionId}/queryables" (L354)   even if we only have datetime, bbox,  we should have a list of queriables no ?

tests/pygeoapi-test-openapi-config.yml
```diff
@@ -0,0 +1,136 @@
...
+        phone: +xx-xxx-xxx-xxxx
+        fax: +xx-xxx-xxx-xxxx
+        email: you@example.org
+        url: Contact URL
+        hours: Hours of Service
+        instructions: During hours of service.  Off on weekends.
+        role: pointOfContact
+
+resources:
+    obs:
```
jorgejesus, line 84: @FarheenB  you should have a resource with filter and another without filter.

FarheenB: @jorgejesus I have tested this having one resource with filter and another without filter. Since pygeoapi-test-openapi-config.yml is a copy of pygeoapi-test-config.yml (with cql config extension) which had only one resource for testing, so kept the changes to the minimal. But I have this file in my local having one resource with filter and another without filter if you say so I can push it here.

tests/test_openapi.py
FarheenB: thank you :)

pygeoapi/openapi.py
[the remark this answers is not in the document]
FarheenB: sure @jorgejesus keeping this on hold for now

docs/source/configuration.rst
FarheenB: @jorgejesus yes I think it will be better to add openAPI section at the later phase of GSoC-2 as by then major part of the implementation will be done and documenting it will be easy.

tests/pygeoapi-test-openapi-config.yml
```diff
@@ -0,0 +1,136 @@
...
+              stn_id:
+                "@id": schema:identifier
+                "@type": schema:Text
+              datetime:
+                "@type": schema:DateTime
+                "@id": schema:observationDate
+              value:
+                "@type": schema:Number
+                "@id": schema:QuantitativeValue
+        filters:
```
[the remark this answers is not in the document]
FarheenB: @francbartoli @jorgejesus  do I need to carry out this change? So for CQL the extensions flag will have filters section? like this (extensions > filters)?

FarheenB:
> @FarheenB @francbartoli 
> 
> IMHO, shouldnt we:
> 
> * Test for a data provider that has no filter support. Currently
>   https://github.com/FarheenB/pygeoapi/blob/27c7c7794117e558fe5f01a99e23ec9d3f35e628/tests/test_openapi.py#L112  has no situation where  it returns `false`
> * And then extent the test aux function to have situations where we have `is_cql=False`  and have `assert` for non presence of filter

Need more clarity on this, as how will a provider specify in its configuration that it supports CQL filter or not? I mean is there any way by which provider is specifying this in the existing system? Under provider folder I can see all the providers have query method implemented. @francbartoli @jorgejesus

tests/pygeoapi-test-openapi-config.yml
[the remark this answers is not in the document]
francbartoli: @FarheenB I would wait a bit before changing the configuration implementation. It is under discussion

pygeoapi/openapi.py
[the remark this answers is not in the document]
jorgejesus: @FarheenB 
- Based on the discussion in gitter the `/queriables` `collections/{collectionId}/queryables` have to be present, you need to update your code based on the disucussion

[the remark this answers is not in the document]
FarheenB: Made the required changes for `/queryables` and `collections/{collectionId}/queryables`. please check https://github.com/FarheenB/pygeoapi/commit/f4aae51a804eb9913442f3fe6bf089ce7f397654 for document generation.
Since these two queryable endpoints are not specific to cql extension, so I removed its assertion from test_openapi.py. Please give a thought on this. @francbartoli @jorgejesus

tests/pygeoapi-test-openapi-config.yml
FarheenB: @jorgejesus added a resource with no cql filter. af58680d1e9da92621ba100638c42b50199334e6

tests/test_openapi.py
```diff
@@ -45,3 +56,465 @@ def test_str2bool():
...
+
+    # assertion for schemas
+    assert get_cql_schemas is not None
+
+    # assertion for queryables
+    assert get_cql_schemas.get('queryables', None) is not None
+    assert get_cql_components['responses'].get('Queryables', None) is not None
+
+    # assertion if filter is not supported
+    if not is_cql:
```
FarheenB, line 490: In reference to https://github.com/geopython/pygeoapi/pull/478#pullrequestreview-443577941
@jorgejesus @francbartoli added test aux function to have assertion for non presence of filter

tests/test_openapi.py
```diff
@@ -45,3 +56,465 @@ def test_str2bool():
...
+            references = []
+            for items_parameters in cql_paths['/collections/'+k+'/items']['get']['parameters']: # noqa
+                if '$ref' in items_parameters:
+                    references.append(items_parameters['$ref'])
+            # if the resource support filter
+            if 'filters' in collections[k]:
+                assert '#/components/parameters/filter' in references # noqa
+                assert '#/components/parameters/filter-lang' in references  # noqa
+            # if the resource does not support filter
+            else:
```
FarheenB, line 171: In reference to https://github.com/geopython/pygeoapi/pull/478#pullrequestreview-443577941
@jorgejesus @francbartoli assertion when resources does not support filter

pygeoapi/openapi.py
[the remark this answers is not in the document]
francbartoli: @FarheenB I would say the second `collections/{collectionId}/queryables` is actually relevant for CQL

docs/source/configuration.rst
francbartoli: @FarheenB @jorgejesus I was thinking about a small piece of docs to avoid forgetting what you have done and the steps for a smooth setup of CQL

tests/test_openapi.py
```diff
@@ -459,13 +443,110 @@ def test_cql_filters_capabilities(get_cql_schemas, is_cql):
         filter_capabilities = ['conformance-classes',
                                'capabilities', 'functions']
         for props in filter_capabilities:
-            assert props in get_cql_schemas['filter-capabilities']['properties'] # noqa
+            assert props in get_cql_schemas['filter-capabilities']['properties']  # noqa
+
+
+def test_cql_queryables_path(get_oas_30_, get_collections, is_cql):
```
FarheenB, line 449: @jorgejesus @francbartoli queryables I have made separate test functions to deal with assertion of root and local queryables

FarheenB:
According to the discussion we had @francbartoli find c864826 for all the latest changes on queryables.
cc. @jorgejesus

pygeoapi/evaluate.py
```diff
@@ -0,0 +1,132 @@
...
+            return filters.arithmetic(
+                to_filter(node.lhs), to_filter(node.rhs), node.op
+            )
+
+        # return the Node
+        return node
+
+
+def to_filter(ast, field_mapping=None, mapping_choices=None):
+    """ Helper function to translate ECQL AST to Django Query expressions.
```
francbartoli, line 118: Refuse from Django integration?

pygeoapi/filters.py
```diff
@@ -0,0 +1,290 @@
...
+def combine(sub_filters, combinator="AND"):
+    """ Combine filters using a logical combinator
+
+        :param sub_filters: the filters to combine
+        :param combinator: a string: "AND" / "OR"
+        :type sub_filters: list of filtered dict 
+        :return: the combined filter
+        :rtype: filtered dict
+    """
+    # check for sub_filters being the evaluation result of multiple sub_filters
```
francbartoli, line 15: I think the assertions should be converted into a try/catch and raise exceptions eventually

pygeoapi/filters.py
```diff
@@ -0,0 +1,290 @@
...
+    "<": "<",
+    "<=": "<=",
+    ">": ">",
+    ">=": ">=",
+    "<>": "!=",
+    "=": "=="
+}
+
+
+def compare(lhs, rhs, op, mapping_choices=None, field_mapping=None):
```
francbartoli, line 65: can you use explicit variable names?

pygeoapi/filters.py
```diff
@@ -0,0 +1,290 @@
...
+            assert lhs in row
+            
+            # perform comparison operation
+            if eval(row[lhs]+comp+str(rhs)):
+                mapping_list.append(row)
+
+    return mapping_list
+
+
+def between(lhs, low, high, not_=False): # TODO!!
```
francbartoli: these functions below should have a `pass` at least or raise a `NotImplementedError`

pygeoapi/provider/csv_.py
```diff
@@ -94,12 +99,24 @@ def _load(self, startindex=0, limit=10, resulttype='results',
         with open(self.data) as ff:
             LOGGER.debug('Serializing DictReader')
             data_ = csv.DictReader(ff)
+            data_list = list(data_)
+            fields = self.fields
+
+            # ===========================================
+            # Added for CQL Filter Expression evaluation
+
+            if filter_expression is not None:
```
francbartoli, line 108: it is recommended and more pythonic to just have:

```python
if filter_expression:
```

pygeoapi/provider/csv_.py
```diff
@@ -94,12 +99,24 @@ def _load(self, startindex=0, limit=10, resulttype='results',
...
+            # ===========================================
+
             if resulttype == 'hits':
                 LOGGER.debug('Returning hits only')
-                feature_collection['numberMatched'] = len(list(data_))
+                feature_collection['numberMatched'] = len(data_list)
                 return feature_collection
             LOGGER.debug('Slicing CSV rows')
-            for row in itertools.islice(data_, startindex, startindex+limit):
+            for row in itertools.islice(data_list, startindex, startindex+limit): # noqa
```
francbartoli, line 119: why `# noqa` ?

pygeoapi/provider/csv_.py
```diff
@@ -149,11 +167,11 @@ def query(self, startindex=0, limit=10, resulttype='results',
         :param datetime: temporal (datestamp or extent)
         :param properties: list of tuples (name, value)
         :param sortby: list of dicts (property, order)
+        :param filter_expression: string of a filter expression
 
         :returns: dict of GeoJSON FeatureCollection
         """
-
-        return self._load(startindex, limit, resulttype)
+        return self._load(startindex, limit, resulttype, filter_expression=filter_expression) # noqa
```
francbartoli, line 174: same above, you should format the code rather than add noqa

pygeoapi/provider/csv_.py
```diff
@@ -94,12 +99,24 @@ def _load(self, startindex=0, limit=10, resulttype='results',
...
             data_ = csv.DictReader(ff)
+            data_list = list(data_)
+            fields = self.fields
+
+            # ===========================================
+            # Added for CQL Filter Expression evaluation
+
+            if filter_expression is not None:
+                ast = pycql.parse(filter_expression)
+                data_list = FilterEvaluator(field_mapping=list(fields.keys()), mapping_choices=data_list).to_filter(ast) # noqa
```
francbartoli, line 110: Remove noqa

pygeoapi/evaluate.py
FarheenB: No, just missed to change this docstring. Fixed now.

pygeoapi/filters.py
FarheenB: like what?

pygeoapi/filters.py
FarheenB: Done

pygeoapi/filters.py
FarheenB: done

pygeoapi/provider/csv_.py
FarheenB: done

pygeoapi/provider/csv_.py
FarheenB: removed

pygeoapi/provider/csv_.py
FarheenB: done

pygeoapi/provider/csv_.py
FarheenB: done

tests/test_csv__provider.py
```diff
@@ -33,7 +33,7 @@
 from pygeoapi.provider.csv_ import CSVProvider
 
 
-path = 'pygeoapi-test.csv'
+path = '/tmp/pygeoapi-test.csv'
```
francbartoli: I would expect to use a loading strategy as already implemented in other tests. An absolute path is only working in certain cases locally not in CI

pygeoapi/filters.py
```diff
@@ -13,14 +13,17 @@ def combine(sub_filters, combinator="AND"):
         :rtype: filtered dict
     """
     # check for sub_filters being the evaluation result of multiple sub_filters
-    assert isinstance(sub_filters, tuple)
+    if not isinstance(sub_filters, tuple):
+        raise AssertionError("Invalid sub_filters value")
```
francbartoli, line 17: I would raise a new specific exception for CQL here (`CQLException` or something?), look at some other specific exceptions in the code base

pygeoapi/filters.py
francbartoli: what does `lhs` stand for?

pygeoapi/filters.py
FarheenB: Okay I will work on it

pygeoapi/filters.py
FarheenB: `lhs` is for Left hand side of the operator and `rhs` is for right hand side of the operator

tests/test_csv__provider.py
FarheenB: `Path='/tmp/pygeoapi-test.csv'` was the path already specified in test_csv_provider.py. For testing locally I changed it to `path='pygeoapi-test.csv'`. But now I reverted back the path location   to as it was before I.e.`Path='/tmp/pygeoapi-test.csv'`. It is same as master now. Do you want me to change the loading strategy? This would affect the master too..

pygeoapi/filters.py
jorgejesus: @FarheenB 
- You have a generic `AssertionError`. The best approach is to have parent class `CQLException` and then chiild clises for the specific error, eg: `CQLExceptionSubFilters` 
@francbartoli : Should we have a module `exceptions.py` and all exception classes would be there described and imported ?

francbartoli: @jorgejesus yep that's the idea

FarheenB: Done

pygeoapi/filters.py
```diff
@@ -0,0 +1,454 @@
+"""
+For evaluating CQL filter queries from Abstract Syntax Tree
+"""
+
+import logging
+from typing import Tuple
```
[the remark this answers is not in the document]
francbartoli: @jorgejesus `typing` is now standard python, my opinion is that maybe we can start using it gradually. Or our coding style doesn't provide it explicitly and then I have lost this reference. Having it along with all the code is huge for now, it would need a large refactor

pygeoapi/provider/csv_.py
```diff
@@ -138,7 +164,8 @@ def _load(self, startindex=0, limit=10, resulttype='results',
         return feature_collection
 
     def query(self, startindex=0, limit=10, resulttype='results',
-              bbox=[], datetime=None, properties=[], sortby=[]):
+              bbox=[], datetime=None, properties=[], sortby=[],
+              filter_expression=None):
```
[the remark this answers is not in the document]
FarheenB: I can change `filter_expression` to `cql_expression`. Also give a look at the latest file where the query() and load() methods are refactored as per the requiremnts https://github.com/geopython/pygeoapi/pull/478/commits/fb23f40b9d845b34ed70cee0e2b8fa4f9eed8db2#diff-499166d2075b4441185bee4b4e1f8d04

pygeoapi/filters.py
```diff
@@ -0,0 +1,322 @@
...
+        if not isinstance(sub_filters, tuple):
+            raise CQLExceptionSubFilters("Invalid sub-filter expression")
+
+        for sub_filter in sub_filters:
+            if not isinstance(sub_filter, list):
+                raise CQLExceptionSubFilters("Invalid sub-filter expression")
+
+        # check for appropriate type of combinator
+        if combinator not in ("AND", "OR"):
+            raise CQLExceptionLogicalCombinator("Invalid logical \
```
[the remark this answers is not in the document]
FarheenB: I saw %s in some files therefore followed the same here. I will fix it.

pygeoapi/filters.py
```diff
@@ -0,0 +1,454 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+class Combinator(enum.Enum):
+    """ Enum for logical operators """
+    AND = "AND"
+    OR = "OR"
+
+
+def combine(sub_filters: Tuple, combination: Combinator = Combinator.AND):
```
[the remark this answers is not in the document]
FarheenB: So should I remove it?

pygeoapi/filters.py
```diff
@@ -0,0 +1,454 @@
...
+                     exclusive
+        :type not_: bool
+        :param mapping_choices: a list of feature dict set to lookup
+                                potential choices for a certain field
+        :type mapping_choices: list
+        :return: filtered feature set
+        :rtype: list
+    """
+
+    # try:
```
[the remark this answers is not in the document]
FarheenB: The code is of use. I am working on this method that's why its commented for now. Will complete it in next commit.

pygeoapi/provider/csv_.py
```diff
@@ -94,12 +99,16 @@ def _load(self, startindex=0, limit=10, resulttype='results',
...
+
+            if resulttype == 'hits' and not filter_expression:
                 LOGGER.debug('Returning hits only')
-                feature_collection['numberMatched'] = len(list(data_))
+                feature_collection['numberMatched'] = count
                 return feature_collection
+
             LOGGER.debug('Slicing CSV rows')
-            for row in itertools.islice(data_, startindex, startindex+limit):
+            for row in itertools.islice(dataset, 0, count):
```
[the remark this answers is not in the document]
FarheenB: @jorgejesus : I shifted pagination after filtering. It is required for keeping `numberMatched` and `numberReturned` values consistent.

pygeoapi/provider/csv_.py
```diff
@@ -72,14 +75,16 @@ def get_fields(self):
             return fields
 
     def _load(self, startindex=0, limit=10, resulttype='results',
-              identifier=None, bbox=[], datetime=None, properties=[]):
+              identifier=None, bbox=[], datetime=None, properties=[],
```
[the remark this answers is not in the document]
FarheenB: refactored code can be found here fb23f40#diff-499166d2075b4441185bee4b4e1f8d04

tests/test_openapi.py
```diff
@@ -45,3 +56,585 @@ def test_str2bool():
...
+
+    cql_paths = get_oas_30_.get('paths', None)
+    # assertion for get paths
+    assert cql_paths is not None
+
+    if is_cql:
+
+        for k, _ in get_collections.items():
+            references = []
+            for items_parameters in cql_paths['/collections/'+k+'/items']['get']['parameters']: # noqa
```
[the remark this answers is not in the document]
FarheenB: @jorgejesus : I will do the changes

pygeoapi/filters.py
```diff
@@ -102,11 +102,23 @@ def compare(lhs, rhs, op, mapping_choices=None):
         if comp:
             # perform comparison operation
             if lhs in mapping_choices[0].keys():
-                mapping_list = list(filter(lambda record, lhs=lhs, comp=comp, rhs=rhs: eval(str(record[lhs]) + comp + str(rhs)), mapping_choices))
```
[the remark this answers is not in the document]
FarheenB: @jorgejesus : I am thinking of an alternative to eval() will update it in next commit.

pygeoapi/provider/csv_.py
```diff
@@ -127,6 +136,25 @@ def _load(self, startindex=0, limit=10, resulttype='results',
                 feature_collection['numberMatched'] = \
                     len(feature_collection['features'])
 
+        if filter_expression:
```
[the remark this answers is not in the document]
FarheenB: @jorgejesus : Json reponse is needed to let CQL filter logic work. So full json response logic has to be evaluated before `filter_expression`  check. And for `hits` to work in case where `filter_expression` is false, then json logic is not needed; `hits` can return the result without proceeding with json logic. But if `filter_expression` is true and `hits` is also true then json response is needed for getting proper count of results returned. Therefore such code checking and structure.

pygeoapi/api.py
```diff
@@ -739,6 +742,22 @@ def get_collection_items(self, headers, args, dataset, pathinfo=None):
             LOGGER.error(exception)
             return headers_, 400, json.dumps(exception)
 
+        # processing cql filter query parameter
+        LOGGER.debug('Processing filter parameter')
+
+        try:
```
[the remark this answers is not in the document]
FarheenB: @jorgejesus : So should I keep it?

pygeoapi/filters.py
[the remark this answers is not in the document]
jorgejesus: Yes the `typing` implementation, not the code by it self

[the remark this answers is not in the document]
FarheenB: Ok will change it

tests/test_openapi.py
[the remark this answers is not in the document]
FarheenB: Done

pygeoapi/filters.py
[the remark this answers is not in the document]
FarheenB: Done

pygeoapi/filters.py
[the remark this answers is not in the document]
FarheenB: Changed

pygeoapi/provider/csv_.py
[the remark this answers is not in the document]
FarheenB: Changed `filter_expression` to `cql_expression`

pygeoapi/provider/csv_.py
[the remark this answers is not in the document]
FarheenB: please check this example which explains the work flow of above code:
image

pygeoapi/filters.py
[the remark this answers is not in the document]
FarheenB: > `http://localhost:5000/collections/lakes/items?f=json&limit=20&startindex=5&filter=id%3E2&filter-lang=cql-text 

But this expression is giving me proper result. 
image

when where you encountering False result?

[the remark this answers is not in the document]
FarheenB: But @jorgejesus even without eval ---> `0>2.0` is returning False.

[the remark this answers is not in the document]
FarheenB: > http://localhost:5000/collections/lakes/items?f=json&limit=20&startindex=5&filter=id%3E2.0&filter-lang=cql-text 

I executed above URL, this expression is giving me proper result. 
image

tests/pygeoapi-test-openapi-config.yml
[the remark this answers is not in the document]
FarheenB: okay, I will need to restructure most of the modules according to this.

pygeoapi/provider/csv_.py
```diff
@@ -163,7 +193,8 @@ def get(self, identifier):
 
         :returns: dict of single GeoJSON feature
         """
-        item = self._load(identifier=identifier)
+
+        item = self.query(identifier=identifier)
```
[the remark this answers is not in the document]
FarheenB: @francbartoli In order to do this 
image
I had to restructure the codebase a bit.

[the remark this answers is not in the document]
jorgejesus: @francbartoli I have asked for the code to be a bit restructured so that the actual functionality of `query`  is there and not on `_load`, I find it easier and more logic this way

tests/pygeoapi-test-openapi-config.yml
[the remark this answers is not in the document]
FarheenB: Done the changes. please check https://github.com/FarheenB/pygeoapi/commit/0463f29b08f7ac385daca40a90b4d43b7fec8ac2

FarheenB:
@francbartoli @jorgejesus I have added a commit for spatial filters like "INTERSECTS", "DISJOINT", "CONTAINS", "WITHIN", "TOUCHES", "CROSSES", "OVERLAPS", "EQUALS", "RELATE", "DWITHIN" and "BEYOND". I have also done certain code restructuring in filters.py to reduce the number of lines and make code more efficient. Added more unit test cases for CSV here. https://github.com/FarheenB/pygeoapi/commit/cbb178869115897de9a9e43d5eea83bee34032d1

FarheenB:
@francbartoli @jorgejesus I have added a commit for temporal filters like "BEFORE", "BEFORE OR DURING", "DURING", "DURING OR AFTER", "AFTER" against #509 . I have also done some code refinement and restructuring in filters.py and other related files. Added more unit test cases for CSV and GeoJSON here https://github.com/FarheenB/pygeoapi/commit/87344f25fec5771e805f9133c90ab2c7b4860505.  Please give it a look.

FarheenB:
@jorgejesus @francbartoli Please check have added functional unit test cases and module pytest for cql here
[test_cql.py](https://github.com/geopython/pygeoapi/pull/478/files#diff-141e1566d396bd2097fcb8eae448e48b) [test_api.py](https://github.com/geopython/pygeoapi/pull/478/files#diff-625efbbf35a2290861f596fee1b19838) 
All the unit test cases related to CQL query filters on CSV Provider can be found here [test_csv__provider.py](https://github.com/geopython/pygeoapi/pull/478/files#diff-40aebfd484a64027b5f0fb453e1c4c21). The same unit test cases need to be written for GeoJSON provider but currently for testing it is using only one hard coded feature. Complex filter expressions to be performed on single feature is becoming difficult.

francbartoli:
@FarheenB can you resolve the conflicts?

tests/test_openapi.py
```diff
@@ -45,3 +58,602 @@ def test_str2bool():
...
+    Get the OpenAPI Document configuration
+
+    :returns: OpenAPI configuration YAML dict
+    """
+    with open(get_test_file_path('pygeoapi-test-openapi-config.yml')) as config_file:  # noqa
+        return yaml_load(config_file)
+
+
+@pytest.fixture()
+def get_oas_30_(config):
```
francbartoli, line 89: Isn’t better to qualify the function name without the verb as it is a fixture?

tests/test_openapi.py
```diff
@@ -45,3 +58,602 @@ def test_str2bool():
...
+
+    :param config: configuration object
+
+    :returns: OpenAPI definition YAML dict
+    """
+    return get_oas_30(config)
+
+
+@pytest.fixture()
+def get_collections(config):
```
francbartoli, line 101: Same here, change the name to `collections`

tests/test_openapi.py
```diff
@@ -45,3 +58,602 @@ def test_str2bool():
...
+    :param config: configuration object
+
+    :returns: list of collection objects
+    """
+    collections = filter_dict_by_key_value(config['resources'], 'type', 'collection')  # noqa
+    return collections
+
+
+@pytest.fixture()
+def get_cql_collections(get_collections):
```
francbartoli, line 114: Same comment here for the name

tests/test_openapi.py
```diff
@@ -45,3 +58,602 @@ def test_str2bool():
...
+    :returns: boolean value
+    """
+    for _, v in get_cql_collections.items():
+        if v:
+            return True
+    return False
+
+
+@pytest.fixture()
+def get_cql_components(get_oas_30_):
```
francbartoli, line 147: Same comment here for the fixture name

pygeoapi/api.py
```diff
@@ -739,6 +740,26 @@ def get_collection_items(self, headers, args, dataset, pathinfo=None):
...
+            cql_expression = args.get('filter')
+            if (cql_expression):
+                cql_handler = load_plugin(
+                    'extensions',
+                    {'name': 'CQL', 'cql_expression': cql_expression}
+                )
+                cql_handler.cql_validation()
+
```
francbartoli, line 754: Isn’t better here to have a dedicated error to catch?

tests/test_csv__provider.py
```diff
@@ -95,15 +99,348 @@ def test_query(config):
...
 def test_get_not_existing_item_raise_exception(config):
     """Testing query for a not existing object"""
+
     p = CSVProvider(config)
     with pytest.raises(ProviderItemNotFoundError):
         p.get('404')
+
+
+# test on common comparisons operations
+def test_eq(config):
```
francbartoli, line 119: It is better to change the name to something that gives more context to cql testing, like `test_cql_eq` in this case

The comment is still valid for all the functions’ name below

tests/data/ne_110m_lakes.geojson
```diff
@@ -2405,178 +2405,6 @@
         "scalerank": 0
       },
       "type": "Feature"
-    },
```
francbartoli, line 2408: I cannot figure out why you are deleting this file

francbartoli (changes requested):

jorgejesus:
@FarheenB  Better that you use the lakes  data  (geojson) to run the  pytest,  tests should also test other geometries aside point `POINT(-75 45)` for example:

- `http://localhost:5000/collections/lakes/items?filter=INTERSECTS(geometry,POINT(60.8729172 31.2695179))&filter-lang=cql-text`
(Ladogameer)
- and filter with a polygons

tests/test_openapi.py
FarheenB: removed from all

pygeoapi/cql.py
```diff
@@ -0,0 +1,293 @@
...
+            Initialize object
+            """
+
+            self.CQLParser = self.CQLParser
+            self.cql_expression = self.cql_expression
+            self.CQLEvaluator = self.CQLEvaluator
+            self.feature_list = self.feature_list
+            self.field_list = self.field_list
+            self.CQLFilter = self.CQLFilter
+
```
justb4, line 234: First of all: great and very useful contribution! Trying to understand the use and lifecycle of Inner Classes here: 

* looks like `CQLFilter` is never instantiated (?) so the `CQLFilter.__init__(self)` is obsolete(?)
* lines 228-233 seem to do nothing, or is there some clever data passing from Outer (FilterHandler) to Inner Class?
* methods of `CQLFilter` are basically called as static/factory methods from `CQLHandler`. See e.g. line 44: `feature_list = self.CQLFilter.get_cql_filtered_list(self)`. Is in e.g. `get_cql_filtered_list(self)` `self` then an instance to the Outer FilterHandler class?
* maybe make the use of static functions in `CQLFilter` more explicit with @static and arguments denoting the FilterHandler object i.s.o. `self`.

justb4:
Would be interesting how to work this out for the OGR Provider, since:

* OGR supports its own [OGR SQL dialect](https://gdal.org/user/ogr_sql_dialect.html), may need to make an (AST)  mapping or may be close to SQLite.
* some OGR Drivers like SQLite/GPKG and PostgreSQL allow passing [their own native non-OGR SQL](https://gdal.org/user/ogr_sql_dialect.html#non-ogr-sql)

Think we should handle OGR Provider with CQL in a separate issue/PR later, when this PR and GSOC project is finished (unless time is left).

pygeoapi/api.py
FarheenB: I have implemented it

tomkralidis:
Closing in lieu of #670

PR #483, opened 2020-07-03
Data Transaction GSoC Phase 1

alex-mathew:
pygeoapi as of now implements [OGC API - Features - Part 1: Core](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html) in which the only HTTP verb supported is GET. So users can perform only read operations from the data providers through the RESTful interface of pygeoapi. This won't cover the needs of most users as they would expect a complete RESTful interface which provides all of the CRUD operations on data providers.

[OGC API - Features - Part 4: Simple Transactions](http://docs.opengeospatial.org/DRAFTS/20-002.html) has defined schemas for creation, updating, replacement, and deletion of resource items via POST, PATCH, PUT, and DELETE HTTP verbs. By implementing this in pygeoapi, we can get a complete RESTful interface which provides all of the CRUD operations on data providers.

None

**Additional context**
I have proposed a sample openapi document with endpoints supporting POST, PATCH, PUT, and DELETE verbs. I have also added a new transaction flag to the config file which tells whether a particular collection resource supports data transactions or not. The generate-openapi-document command now generates transaction schemas in the openapi document if the transactions flag is set to true for the corresponding resource in the config file. I have also added test cases for transaction support for collections and for checking whether generated schemas of all four data transaction operations are in compliance with the proposed schemas. I have also added a subsection for transactions under configuration in RTD. The code passed flake8 linting and unit tests. The build for the data-transaction-GSoC-1 branch was successful in Travis.

- pygeoapi configuration with transactions flag
```
resources:
    obs:
        type: collection
        extensions:
             transactions: True
```

- Generated OpenAPI JSON document
Screen Shot 2020-07-03 at 10 42 30 PM

- Passing test cases
Screen Shot 2020-07-03 at 10 42 06 PM

- Travis Build
https://travis-ci.org/github/alex-mathew/pygeoapi/builds/704651775
Screen Shot 2020-07-03 at 11 07 00 PM

- RTD
https://pygeoapi-data-transaction.readthedocs.io/en/data-transaction-gsoc-1/configuration.html#data-transactions
Screen Shot 2020-07-08 at 11 37 39 PM


pygeoapi-config.yml
```diff
@@ -105,6 +105,7 @@ resources:
               href: https://raw.githubusercontent.com/mapserver/mapserver/branch-7-0/msautotest/wxs/data/obs.csv
               hreflang: en-US
         extents:
+            transactions: true
```
francbartoli, line 108: @alex-mathew @jorgejesus I cannot figure out why the configuration is under the key `extents` that is scoped for the spatial and temporal dimensions. Can you explain what is the design?

tests/test_openapi.py
```diff
@@ -45,3 +52,339 @@ def test_str2bool():
...
+    post_resp_attrib = ['201',
+                        '400',
+                        '404',
+                        '500']
+    post_resp = post['responses']
+    for attrib in post_resp_attrib:
+        assert attrib in post_resp
+
+
+def patch_schema_assertions(verbs):
```
francbartoli, line 218: Is this a test or a utility function? The same comment is valid for several functions below

data_transaction_openapidoc_proposal.yml
```diff
@@ -0,0 +1,727 @@
+components:
```
francbartoli, line 1: I think this file is useless in the code base, maybe it can be annexed in the issue of the proposal

.gitignore
```diff
@@ -112,3 +112,8 @@ ENV/
 
 
 .DS_Store
+
+#Config file and generated openapi doc
```
francbartoli, line 116: Given the name can be user-generated I think we don't need these specific exclusions

tests/test_openapi.py
alex-mathew: I have created functions to group assertions related to each schema. The test function will call them if transactions are enabled for a resource.

data_transaction_openapidoc_proposal.yml
alex-mathew: Yeah. It was just for the initial proposal.

.gitignore
alex-mathew: Yeah, will remove them.

data_transaction_openapidoc_proposal.yml
alex-mathew: removed

.gitignore
alex-mathew: removed

tests/test_openapi.py
francbartoli: I used to have assertions within a test function but I don't know if it is common this coding pattern. cc @jorgejesus

pygeoapi-config.yml
alex-mathew: @francbartoli I talked with Jorge and we agreed on adding a new extensions section under resource and keep transactions it. This section can be used for keeping flags of feature extensions.
```
obs:
       extents:
                  spatial:  . . .
                  temporal: . . .
       extensions:
                  transactions: true
```
What do you think?

francbartoli: sounds good to me but let's involve more developers in the decision to get more consensus cc @tomkralidis @justb4 @alpha-beta-soup

jorgejesus: @francbartoli @tomkralidis  @justb4 @alex-mathew @alpha-beta-soup 

My intial idea was to just to have everything on one place, and anything that would be passed as function arguments, to be under `extents` e.g:

https://github.com/geopython/pygeoapi/blob/d2cbf48cd3f7663caf7520b69451723cfadf47f7/pygeoapi/provider/csv_.py#L140-L141

But  after thinking (nadenken): my current idea is: 
- extents: <dimension data slice, spatial, temporal, pagination, columns>
- extensions: OGC extensions  and functionalities applied to data

tests/test_openapi.py
jorgejesus: @alex-mathew  @francbartoli 

IMHO: 
- Better more asserts than less
- `assert` inside auxiliary functions maybe is less positive but  we have easier to read code with smaller reusable functions (find this more important).
- `assert` in main code is an anti-padron and should not be used

pygeoapi-config.yml
alex-mathew: @jorgejesus @francbartoli Can I move forward with this? (extensions > transactions)

tests/test_openapi.py
francbartoli: @jorgejesus should have anyway the label `test_*` if they are asserting something?

jorgejesus: @francbartoli  IMHO `test_*` from the actual test being performed, then generic function like `check_transaction_schemas_abscent` can have `assert` but they are related to the test they are being called `test_transaction_a_b` what we have now should be ok (my 2 cent)

francbartoli: @jorgejesus the only concern is that those functions get out of pytest which is a little bit weird

PR #487, opened 2020-07-09
fix raw output for OAProc

tomkralidis:
This PR fixes OAProc raw output for non-JSON representations.  Other changes:

- updates OAProc OpenAPI defs
- changes `raw=true` to `document=raw` for raw output
- changes execute HTTP status to 200 (201 is now only for asynchronous)

cc @RousseauLambertLP


tomkralidis:
@francbartoli done

PR #498, opened 2020-07-25
minor alignments

tomkralidis:


francbartoli:
@tomkralidis test needs some more alignments

PR #501, opened 2020-07-27
Issue 499 - Moving id field of geojson providers to root

alex-mathew:
- Moved id field from properties to root for `pygeoapi/tests/data/ne_110m_lakes.geojson`
- Created root level id field for `pygeoapi/tests/data/ne_110m_populated_places_simple.geojson`
- Modified geojson provider code (`geojson.py`) to read id field from root instead of properties


jorgejesus:
Checked and everything and on a running pygeoapi I can see the `id` at root level

image

PR #504, opened 2020-07-30
Corrections to docstrings for delete method in providers

isedwards:
Correct typo in mongo.py provider's `delete` method and use corrected version to replace incorrect description in geojson.py and base.py


tomkralidis:
Thanks @isedwards !

PR #515, opened 2020-08-16
Issue 499 - Move id field back to properties and adjust logic to check id field in both root and properties

alex-mathew:
- Moved id field back to properties in `pygeoapi/tests/data/ne_110m_lakes.geojson`
- Modified geojson provider code (`geojson.py`) to read id field from the root and if not present in the root, then from properties
- Removed a dummy feature in `pygeoapi/tests/data/ne_110m_lakes.geojson` which was inserted in my last PR by mistake


alpha-beta-soup:
There's a lot of this branching pattern in the PR:

```python
if self.id_field in feature:
    if feature[self.id_field] == identifier:
        all_data['features'].pop(i)
    elif self.id_field in feature['properties']:
        if feature['properties'][self.id_field] == identifier:
            all_data['features'].pop(i)
```

Which could possibly be made less repetitive and nested with something like:

```python
_identifier = feature.get(self.id_field, None) or feature['properties'].get(self.id_field, None)
if _identifier == identifier:
    # do thing
```

PR #522, opened 2020-08-27
Issue 499 : Fixing geojson provider

alex-mathew:
i. ``_load`` method now loads features from geojson file, checks each feature and if an "id" field is not present in the feature, creates an "id" field with value loaded from id field(from config) in properties
ii. ``get`` method checks "id" field in each of the loaded features and returns the one matching with  given identifier

For example, if id field from config is ``fid``:

```
type: Feature
geometry:
properties:
    fid: 123
```

```
type: Feature
id: 123
geometry:
properties:
```
Both of these features will pass get.


justb4:
Looks good: GeoHealthCheck succeeds again for [pygeoapi master check](https://demo.geohealthcheck.org/resource/165?lang=en)

PR #525, opened 2020-09-01
xarray Coverage Provider

gpetrochenkov-usgs:
This PR includes the following:

- XarrayProvider with domain set, range set, and queries with subset and rangeSubset.  (Including a doc and unit tests.)
- Generalize the coverage rangeSubset mechanism to accommodate both rasterio and xarray providers.

E.g.
http://localhost:5000/collections/xarray_test/coverage?rangeSubset=SST&f=json
http://localhost:5000/collections/gdps-temperature?rangeSubset=1&f=json

The main problem with the XarrayProvider is that a majority of data sets store NaN as no data.  The native json library will parse NaNs but will make invalid json.  Encoding rules are only possible for data types that are not serializeable so such rules will not work for NaNs.  That leaves two options:

- Introduce simplejson dependency with ignore_nan=True in dump method
- NaN fill data array with a value that will remain valid after being serialized


pygeoapi/provider/xarray.py
```diff
@@ -0,0 +1,435 @@
...
+    * How to raise errors in the provider implementation such that
+    * appropriate HTTP responses will be raised
+    """
+
+    def __init__(self, provider_def):
+        """initializer"""
+        BaseProvider.__init__(self, provider_def)
+
+        try:
+            zarr = self.data.split('.')[-1] == 'zarr'
```
tomkralidis, line 66: faster like: `'zarr' in self.data[5:]`

pygeoapi/provider/xarray.py
```diff
@@ -0,0 +1,435 @@
...
+            data.coords[self.lat_var].values[-1]],
+            "time": self._coverage_properties['time_range'],
+            "driver": "Xarray",
+            "height": data.dims[self.lat_var],
+            "width": data.dims[self.lon_var],
+            "time_steps": data.dims[self.time_var],
+            "variables": {var_name: var.attrs
+                          for var_name, var in data.variables.items()}
+        }
+
```
tomkralidis, line 230: We should ask functionality to return the coverage in the native format.  So if the config specifies:
```
              format:
                  name: GRIB2
                  mimetype: application/x-netcdf
```

See example in rasterio:
https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/rasterio_.py#L254-L261

docs/source/data-publishing/ogcapi-coverages.rst
```diff
@@ -44,6 +44,19 @@ capable of handling.
              name: GRIB2
              mimetype: application/x-grib2
 
+xarray
+^^^^^^^^
+
+The `xarray`_ provider plugin reads and extracts any data that xarray is
```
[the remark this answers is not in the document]
gpetrochenkov-usgs: I added this link to the end of the document.

pygeoapi/provider/xarray.py
gpetrochenkov-usgs: I added the ability to download a native netCDF file.  Unfortunately it added a scipy dependency to write a memory file.

pygeoapi/provider/rasterio_.py
```diff
@@ -161,16 +158,15 @@ def get_coverage_rangetype(self):
 
         return rangetype
 
-    def query(self, bands=[], subsets={}, format_='json'):
+    def query(self, range_type=[], subsets={}, format_='json'):
         """
         Extract data from collection collection
-
-        :param bands: list of bands (int)
+        :param range_type: list of bands (int)
```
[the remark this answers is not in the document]
gpetrochenkov-usgs: This is all strings now, later you map the range subsets with an int type so no pre type casting is necessary.

PR #528, opened 2020-09-11
fix coverage query with no parameters

tomkralidis:
Fix coverage query with no parameters to emit the file, as is, is the file is local on disk.


tomkralidis:
cc @constantinius

PR #539, opened 2020-09-23
Using Blueprint/route list for flask/starlette

### LINKED ISSUE #530: Improving reusability with Flask Blueprints

We are trying to combine pygeoapi with an existing Flask App. There is the way to do so using [werkzeug middlewares](https://flask.palletsprojects.com/en/1.1.x/patterns/appdispatch/#combining-applications), but this is really low level and hard to do correctly.

I think the best solution would be to use the Flask Blueprint to set up the routes to the internal functions, which then can be mounted in the Flask App under the root URL.

With this way, it is easy to just mount the pygeoapi URLs in any other Flask based App.

This would be the change in the `flask_app.py`

```python

# imports and such ...
from flask import Flask, Blueprint, make_response, request, send_from_directory
# ...

URLS = Blueprint('pygeoapi', __name__)
APP = Flask(__name__)
APP.url_map.strict_slashes = False

# ...

@URLS.route('/')
def landing_page():
    # ...

# other routes
# ...

# finally mount the Blueprint
APP.register_blueprint(URLS, url_prefix='/')

# ... everything else
```

Mentioned in the description above using werkzeug middlewares.

**Additional context**

none

constantinius:
Attempt to fix #530 

Needs testing, especially with starlette.

### DIFF
--- a/pygeoapi/flask_app.py
+++ b/pygeoapi/flask_app.py
@@ -34,13 +34,15 @@ import os
 
 import click
 
-from flask import Flask, make_response, request, send_from_directory
+from flask import Flask, Blueprint, make_response, request, send_from_directory
 
 from pygeoapi.api import API
 from pygeoapi.util import get_mimetype, yaml_load
 
 APP = Flask(__name__)
 APP.url_map.strict_slashes = False
+
+routes = Blueprint('pygeoapi', __name__)
 
 CONFIG = None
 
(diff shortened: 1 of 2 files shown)


constantinius:
@tomkralidis This PR is now ready. There is a test failing, but I doubt it is related to this PR.

Could you have a look if this new functionality makes sense?

tomkralidis:
Thanks @constantinius.  Can you address the flake8 issue in `pygeoapi/starlette_app.py` ?

PR #541, opened 2020-09-24
fix failing test

tomkralidis:


tomkralidis:
@justb4 tests pass. Wondering if we need to consider strategy to handle data changes from the WFS service provider?

justb4:
Ok if test passes.Not sure why "Mosselsepad" does not work.

justb4:
Ok, I now see that a specific DB-generated id is used. That is brittle as the dataset daily changes...

PR #542, opened 2020-09-24
Improve docker image cache usage

totycro:
Before, any file change in the repo would cause a reinstall of all
dependencies.

This commit splits the dockerfile in 2 phases:
First the files relevant for installation are added and all dependencies
are installed. Then the actual source code is added.

This way, rebuilding after sources have been changes takes a few seconds
rather than minutes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,7 +66,10 @@ ENV TZ=${TIMEZONE} \
 	DEB_PACKAGES="locales libgdal27 python3-gdal libsqlite3-mod-spatialite ${ADD_DEB_PACKAGES}" \
 	PIP_PACKAGES="gunicorn==20.0.4 gevent==1.5a4 wheel==0.33.4 ${ADD_PIP_PACKAGES}"
 
-ADD . /pygeoapi
+RUN mkdir -p /pygeoapi/pygeoapi
+# Add files required for pip/setuptools
+ADD requirements*.txt setup.py README.md /pygeoapi/
+ADD pygeoapi/__init__.py /pygeoapi/pygeoapi/
 
 # Run all installs
 RUN \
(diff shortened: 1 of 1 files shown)


totycro:
btw the test which fails here is fixed in https://github.com/geopython/pygeoapi/pull/541

francbartoli:
> btw the test which fails here is fixed in #541

@totycro can you update then?

kalxas (approved):

totycro:
Thanks for the review!
I've rebased the branch to master now.

justb4 (approved):
I agree that with these changes the Docker Image will be quicker to build. Main reason, as I recall, for having `ADD . /pygeoapi` at the top of the Dockerfile is that the build not only `pip3`-installs requirements but also installs the `pygeoapi` package into Python `site-packages` and installs CLI commands. See the `pip3 install -e .`  in the list of `RUN` commands. Guess the copy of the `__init__.py` file and `pip3 install -e .` still installs the CLI needed for `generate_openapi_document` during Docker Container startup. Have you tested that that works?

Then only if users want to build on this Dockerfile importing pygeoapi packages from `site-packages` may this be a problem, or they can set PYTHONPATH...So ok with me but maybe add some comments to explain.

I hope I made this clear/understandable.

totycro:
I had to add `__init__.py` (as well as `README.md`) because `setup.py` reads it, that's the only reason for this.

After building the image with this, i verified that the tests pass and I'm currently using it mostly for the process API. The openapi documentation is generated just fine.
As far as I can tell there's nothing broken when building like this. However there are large parts which I haven't used.

justb4:
@totycro yes the `setup.py` (via `RUN pip3 install -e .`) installs the `pygeoapi-CLI` commands like `generate_openapi_document` defined in `__init__.py` . So lets move on with merging if the other reviewers agree.

francbartoli (approved):

PR #543, opened 2020-09-24
Improvements in .gitignore and installation docs

plant99:
1. Example configs generated as mentioned in installation docs are tracked by git
2. Current directory structure if quick installation docs is followed is as follows
```
pygeoapi/
└── <env related directories>
└──pygeoapi
         └──<project related directories>
```

With the change in installation docs the directories would be structured as follows
```
pygeoapi-env/
└── <env related directories>
pygeoapi
└──<project related directories>
```
 3. Also `source` removes the additional overhead to make `bin/activate` executable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,7 @@ ENV/
 
 
 .DS_Store
+
+# example configs
+
+example-*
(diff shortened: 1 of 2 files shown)


plant99:
@francbartoli couldn't add you as reviewer. Please review this.

francbartoli:
@plant99 thanks for this proposal. However, I would leave this to the specific workflow of various developers. I don’t think, honestly, that would add particular value to the docs journey

PR #544, opened 2020-09-24
Improve signal handling in entry point

totycro:
Currently gunicorn is started in the entrypoint as subprocess. This
means that the entrypoint stays as process with PID 1 and eats the
signals that e.g. kubernetes sends to it. Thus when trying to update
the container, kubernetes must terminate the process ungracefully.

See e.g. this resources for a more detailed explanation:
https://hynek.me/articles/docker-signals/

This change replaces the entrypoint-shellscript with gunicorn, such that
signals are directly forwarded to gunicorn, where they are then handled.

One drawback of this approach is that lines after `exec gunicorn` are
not executed anymore. This currently affects `echo "END /entrypoint.sh"`.

A different approach would be to use `tini` to handle the system (see
link above). To do this, we would need to install tini.


francbartoli:
@totycro can you please update this so tests can pass?

totycro:
I've rebased the branch to master now.

francbartoli (approved):

PR #548, opened 2020-09-29
app mount updates

tomkralidis:
This PR adds builds on 8fe060173ac82c0c45149eb13d9c025cf489b809 and adds docs as well as some renaming and refactoring for Starlette.

cc @constantinius


constantinius:
Looks good to me! 👍

Issue #495, opened 2020-07-23
One OGR WFS test fails and one stalls test process


justb4:
**Description**
`test_gs_force_getting_gml_id fails` and `test_get_gs_with_geojson_output_too_complex_raise_exception` stalls process.
Both are OGR WFS tests in `tests/test_ogr_wfs_provider.py `

**Steps to Reproduce**

* run `pytest -v tests/test_ogr_wfs_provider.py `
* seeing test 6 fail and 7 stalling

**Expected behavior**
* expect all tests succeed

**Screenshots/Tracebacks**
```
pytest -v tests/test_ogr_wfs_provider.py 
============================================================== test session starts ===============================================================
platform linux -- Python 3.8.4, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /pygeoapi, inifile: pytest.ini
plugins: env-0.6.2, cov-2.10.0
collected 21 items                                                                                                                               

tests/test_ogr_wfs_provider.py::test_get_fields_gs PASSED                                                                                  [  4%]
tests/test_ogr_wfs_provider.py::test_get_ms PASSED                                                                                         [  9%]
tests/test_ogr_wfs_provider.py::test_get_geosol_gs PASSED                                                                                  [ 14%]
tests/test_ogr_wfs_provider.py::test_get_gs PASSED                                                                                         [ 19%]
tests/test_ogr_wfs_provider.py::test_gs_not_getting_gml_id PASSED                                                                          [ 23%]
tests/test_ogr_wfs_provider.py::test_gs_force_getting_gml_id FAILED                                                                        [ 28%]
tests/test_ogr_wfs_provider.py::test_get_gs_with_geojson_output_too_complex_raise_exception ^C
```

**Environment**
 - OS: any
 - Python version: 3.8 (used the Docker image)
 - pygeoapi version: master on demo  (used the Docker image to reproduce)

**Additional context**
N.A.

francbartoli:
It is not a bug per se. This call https://geonode.wfp.org/geoserver/wfs?service=WFS&request=DescribeFeatureType&typeName=geonode:wld_bnd_adm0 gives an error server side. So the gml_id is not available and the other stalls

justb4:
True, it is not a `pygeoapi`-bug. Any chance to fix this on the GeoServer-side?

tomkralidis:
[...]

PR #552, opened 2020-10-10
fix OGR WFS tests using MapServer demo

tomkralidis:
This PR updates the OGR WFS testing to use the [MapServer demo WFS service](https://demo.mapserver.org/cgi-bin/wfs) in support of #495.

I'm not sure if any specific functionality native to the geodata.nationaalgeoregister.nl WFS services is now lost, but we would benefit from more consistent testing of remote endpoints.

cc @francbartoli @pvgenuchten

### DIFF
--- a/tests/test_ogr_wfs_provider.py
+++ b/tests/test_ogr_wfs_provider.py
@@ -2,9 +2,11 @@
 #
 # Authors: Just van den Broecke <justb4@gmail.com>
 # Authors: Francesco Bartoli <xbartolone@gmail.com>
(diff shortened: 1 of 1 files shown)


francbartoli (approved):

justb4:
Ok, sorry folks, missed the comms completely. Good that this is solved. 

What is lost now are the various SRS combinations, that forces OGR/Proj to reproject, using the source/target config params of OGRProvider. That was one of the reasons to choose a Dutch (EPSG:28992) WFS . But that Dutch WFS service appeared to be too fragile. I see the same in GHC monitoring for that WFS.

PR #557, opened 2020-10-13
Implement job deletion in UI, api and tinydb

totycro:
This is a very basic implementation for deleting jobs, but I'm already using it successfully with my kubernetes backend. I'm definitely open to further suggestions.

It basically adds this button:

Screenshot_2020-10-13_10-26-22

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -1660,6 +1660,16 @@ class API:
             return headers_, 200, response
         return headers_, 200, json.dumps(job_output, sort_keys=True, indent=4, default=json_serial)
 
+    def delete_job(self, process_id, job_id):
+        """
+        :param process_id: name of process
+        :param job_id: ID of job
+
+        :returns: tuple of headers, status code, content
+        """
+        success = self.manager.delete_job(process_id, job_id)
+        return {}, 204 if success else 404, ""
+
 def check_async(args, headers):
     """
     Check execution mode requested from arguments or headers. Returns
--- a/pygeoapi/flask_app.py
+++ b/pygeoapi/flask_app.py
@@ -280,7 +280,7 @@ def describe_processes(process_id=None):
 
 
 @APP.route('/processes/<process_id>/jobs', methods=['GET', 'POST'])
-@APP.route('/processes/<process_id>/jobs/<job_id>', methods=['GET'])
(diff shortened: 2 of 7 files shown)


alpha-beta-soup:
This would be useful in the list of jobs as well, especially for deleting failed or pending jobs.

totycro:
I was considering that too, so now I also added it.

As i've mentioned in the commit the UI is quite basic, but i didn't want to build any thing fancy in plain javascript right now. Maybe it's also a good idea to work towards merging this branch with a basic but functioning feature set and later improving on it.

alpha-beta-soup:
Simple is good. 

My only other comment is about the tinydb deletion. Should the job output (if any) be deleted from disk too?

totycro:
I've added the output file deletion as well.

alpha-beta-soup:
I can't get the deletion tests to pass locally, with either the default (tinyDB) or my own processing manager (redis), though I'm not sure why. I'd be interested in seeing your processing manager implementation; and if these tests pass for you with the tinyDB manager.

After merging, in the `manager` branch, as part of a merge commit with the latest `master`, I also modified the behaviour of the delete button slightly: if clicked while on `/jobs/{id}` it will wait then redirect to `/jobs`. If already on the `/jobs` page then it will wait then refresh.

totycro:
My implementation is here in the eoxhub branch:
https://github.com/eurodatacube/pygeoapi/blob/eoxhub/pygeoapi/process/manager/kubernetes.py
https://github.com/eurodatacube/pygeoapi/blob/eoxhub/pygeoapi/process/notebook.py

Good point about adding the redirect behavior! It was kind of odd that you would still see the deleted job after deletion.

I ran the tests in docker and as far as i remember they passed there. What error message you getting?

Btw is there a time frame on when `manager` might be merged to `master`?

alpha-beta-soup:
I was testing in Docker too, I got some error about there being a malformed JSON string that was being `json.load`ed. It seemed like a race condition. It could have been something on my end though, the relevant tests are passing on Travis (`tests/test_api.py`).

> Btw is there a time frame on when manager might be merged to master?

I believe @tomkralidis is reviewing the branch at the moment.

For my part, I'm starting to use this (with redis manager) for internal clients, hopefuly to discover as many issues as possible while it's still in an isolated feature branch. If you have any more ideas to propose then now is an excellent time to expedite them. The branch is  now up-to-date with master.

I'd like to improve test coverage a bit.

PR #559, opened 2020-10-27
reproject dataset bounds to 4326 while analyzing stac content

pvgenuchten:
map viewer (and i assume geojson in general) expects geometry to be in 4326, with this reprojection the bounds look fine on map (before were covering full earth)
also added original projection as additional property
adds shp and tif to stac supported filetypes

image

### DIFF
--- a/pygeoapi/provider/filesystem.py
+++ b/pygeoapi/provider/filesystem.py
@@ -259,10 +259,16 @@ def _describe_file(filepath):
     except rasterio.errors.RasterioIOError:
         LOGGER.debug('Testing vector data detection')
         d = fiona.open(filepath)
-        tcrs = CRS.from_epsg(4326)
-        bnds = transform_bounds(CRS(d.crs), tcrs,
-                                d.bounds[0], d.bounds[1],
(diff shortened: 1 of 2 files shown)


pygeoapi/provider/filesystem.py
```diff
@@ -257,31 +259,35 @@ def _describe_file(filepath):
     except rasterio.errors.RasterioIOError:
         LOGGER.debug('Testing vector data detection')
         d = fiona.open(filepath)
-
+        tcrs = CRS.from_epsg(4326)
```
tomkralidis, line 262: Should/can we test whether the data's CRS is 4326, and if not, then reproject?

tomkralidis (changes requested):

pygeoapi/provider/filesystem.py
pvgenuchten: Or projection nill, also is quite common

tomkralidis:
Failing tests unrelated.  Merging.

pvgenuchten:
Thanx Tom

PR #560, opened 2020-10-28
add bbox and datetime support to OACov providers

tomkralidis:
Adds OACov support for `bbox` and `datetime` API parameters  to rasterio and xarray providers.  Note that bbox values are reprojected into native coordinates.  xarray based updates are being made on behalf of @RousseauLambertLP (thanks LP!)


tomkralidis:
@gpetrochenkov-usgs can you review the xarray based updates and approve if they are good for you?  Thanks in advance.

gpetrochenkov-usgs (approved):
Everything looks good.  A minor suggestion would be to make sure that splitting the datetime_ arg has dates in chronological order.

tomkralidis:
Thanks @gpetrochenkov-usgs I've added a test to order accodingly and to provide logging debug if reversed.

Issue #562, opened 2020-10-28
improve stac-item according to stac spec


pvgenuchten:
Current implementation of a STAC item varies quite a bit between the types of resources being crawled. For example for shapefiles the file columns are added as properties

According to [STAC Item Spec](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md) the following aspects would need to be minimally implemented

- a datetime property is required, maybe we can use the file timestamp for this
- the name of the collection of which this resource is part of

In order to facilitate [common metadata fiels](https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#stac-common-metadata), it would be desirable to:
- parse some metadata included in the resource (gpkg_metadata table, xxx.shp.xml file, note that this requires awareness of various metadata formats (DC, iso, esri, ...))
- and/or be able to provide some metadata by placing an additional file in the folder which pygeoapi includes in the stac-item response

An example of such a file {resource-name}.md.yml
```YAML
title: Forest lakes in province of Utrecht
description: This dataset contains all lakes in province
license: https://opendefinition.org/licenses/cc-by/
providers:
  name: Province of Utrecht
  roles: ContactPoint
  url: https://utrecht.nl

```

pvgenuchten:
Fixed in #563

PR #563, opened 2020-10-28
adds metadata properties for stac items

pvgenuchten:
Metadata items added:
- datetime
- filesize
- some attributes from an optional metadata file: {basename}.md.yml
- for vector files, moves columnnames to a property 'model'

This is an initial response to #562

Potential improvements:
- if md file already lists a property (size/date/title, use that in stead of the one extracted from file)
- properties do incidentally contain array/object, html visualization does not properly manage this yet
- the name of the collection should be added as property (is that property stored somewhere, or should be extracted from path?)

example md.yml file included for obs.csv

image

### DIFF
--- a/pygeoapi/provider/filesystem.py
+++ b/pygeoapi/provider/filesystem.py
@@ -220,11 +220,11 @@ def _describe_file(filepath):
     }
 
     try:
-        from datetime import datetime
+        from datetime import datetime as dt
         t = os.path.getmtime(filepath)
         s = os.path.getsize(filepath)
-        format = "%Y-%m-%d %H:%M"
-        content['properties']['datetime'] = datetime.fromtimestamp(t).strftime(format)
+        f = "%Y-%m-%d %H:%M"
+        content['properties']['datetime'] = dt.fromtimestamp(t).strftime(f)
         if s > 0:
             content['properties']['size'] = str(round(s/100)/10)
     except ValueError as err:
@@ -286,7 +286,7 @@ def _describe_file(filepath):
                     [d.bounds[0], d.bounds[1]]
                 ]]
             }
-        
+
         model = {}
         for k, v in d.schema['properties'].items():
             model[k] = v
--- a/pygeoapi/provider/filesystem.py
+++ b/pygeoapi/provider/filesystem.py
@@ -220,6 +220,16 @@ def _describe_file(filepath):
     }
 
     try:
+        import datetime
+        t = os.path.getmtime(filepath)
+        s = os.path.getsize(filepath)
+        content['properties']['datetime'] = datetime.datetime.fromtimestamp(t)
+        if s > 0:
+            content['properties']['size'] = str(round(s/100)/10)
+    except ValueError as err:
+        LOGGER.debug(err)
+
+    try:
         import rasterio
         LOGGER.warning('rasterio not found. Cannot derive geospatial properties')  # noqa
     except ImportError as err:
@@ -275,8 +285,22 @@ def _describe_file(filepath):
                     [d.bounds[0], d.bounds[1]]
                 ]]
             }
+        
+        model = {}
         for k, v in d.schema['properties'].items():
-            content['properties'][k] = v
+            model[k] = v
+        if model.items():
+            content['properties']['model'] = model
+
+        from pygeoapi.util import yaml_load
+        pth = os.path.splitext(filepath)[0]
+        try:
+            with open('./{}.md.yml'.format(pth)) as fh:
+                md = yaml_load(fh)
+            for k, v in md.items():
+                content['properties'][k] = v
+        except (FileNotFoundError, IOError):
+            LOGGER.debug('file {}.md.yml not found.'.format(pth))
 
         if d.driver == 'ESRI Shapefile':
             id_ = os.path.splitext(os.path.basename(filepath))[0]
--- a/pygeoapi/provider/filesystem.py
+++ b/pygeoapi/provider/filesystem.py
@@ -30,7 +30,11 @@
 import io
 import logging
 import os
+from json import loads
+from datetime import datetime as dt
 from urllib.parse import urljoin
+from pygeometa.core import read_mcf, MCFReadError
+from pygeometa.schemas.stac import STACItemOutputSchema
 
 from pygeoapi.provider.base import (BaseProvider, ProviderConnectionError,
                                     ProviderNotFoundError)
@@ -207,6 +211,8 @@ class FileSystemProvider(BaseProvider):
 def _describe_file(filepath):
     """
     Helper function to describe a geospatial data
+    First checks if a sidecar mcf file is available, if so uses that
+    if not, script will parse the file to retrieve some info from the file
 
     :param filepath: path to file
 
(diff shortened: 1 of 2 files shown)


pygeoapi/provider/filesystem.py
```diff
@@ -275,8 +285,22 @@ def _describe_file(filepath):
...
             }
+        
+        model = {}
         for k, v in d.schema['properties'].items():
-            content['properties'][k] = v
+            model[k] = v
+        if model.items():
+            content['properties']['model'] = model
+
+        from pygeoapi.util import yaml_load
```
tomkralidis, line 295: In terms of a metadata sidecar file, do we have anything more formal? An OARec record model would be perfect here (once it is ready that is).

tomkralidis (changes requested):

pygeoapi/provider/filesystem.py
pvgenuchten: +1 for more formal, I used the stac model for now.

pvgenuchten: @tomkralidis would you suggest to use pygeometa or ogc-api-records format? Or even dcat (as json-ld). Any of those would need a transformation for use in stac, but is more future proof if at some point we want to use the filesystem provider as a source for ogc-api-records

tomkralidis: @pvgenuchten ideally we would allow a user to specify what the side car files are in config (i.e. xml, yaml), as well as their representation so then pygeoapi can handle accordingly.  Having said this, perhaps pygeometa [MCF files](https://github.com/geopython/pygeometa/blob/master/sample.yml) will do the trick here (of course MCF file properties may need some tweaking over in pygeometa).

pvgenuchten: I prefer a limited set of formats, no on the fly metadata conversion from various formats, there's other tools for that. A scheduled process can run over the files and prepare mcf files for each.

Actually you can already add mcf with this PR, it will however not be transformed to the model required by STAC. 

I propose to merge this and continue the work in next PR's, what do you think?

pvgenuchten:
I need some help with the build failure

```
pygeoapi/provider/rasterio_.py:32: in <module>
    from pyproj import CRS, Transformer
/usr/local/lib/python3.8/dist-packages/pyproj/__init__.py:50: in <module>
    from pyproj import _datadir
pyproj/_datadir.pyx:6: in init pyproj._datadir
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
    """
    Module for managing the PROJ data directory.
    """
    import os
    import sys
>   from distutils.spawn import find_executable
E   ModuleNotFoundError: No module named 'distutils.spawn'
/usr/local/lib/python3.8/dist-packages/pyproj/datadir.py:6: ModuleNotFoundError
```


```
pygeoapi/provider/rasterio_.py:32: in <module>
    from pyproj import CRS, Transformer
/usr/local/lib/python3.8/dist-packages/pyproj/__init__.py:50: in <module>
    from pyproj import _datadir
pyproj/_datadir.pyx:6: in init pyproj._datadir
```


```
ImportError while importing test module '/pygeoapi/tests/test_rasterio_provider.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_rasterio_provider.py:33: in <module>
    from pygeoapi.provider.rasterio_ import RasterioProvider
pygeoapi/provider/rasterio_.py:32: in <module>
    from pyproj import CRS, Transformer
/usr/local/lib/python3.8/dist-packages/pyproj/__init__.py:50: in <module>
    from pyproj import _datadir
pyproj/_datadir.pyx:6: in init pyproj._datadir
    ???
/usr/local/lib/python3.8/dist-packages/pyproj/datadir.py:6: in <module>
    from distutils.spawn import find_executable
E   ModuleNotFoundError: No module named 'distutils.spawn'
```

error seems related to a reprojection library, probably triggered from my code changes, but I don't quite get how to assess/solve.

pygeoapi/provider/filesystem.py
pvgenuchten: > ideally we would allow a user to specify what the side car files are in config (i.e. xml, yaml), as well as their representation so then pygeoapi can handle accordingly. 

We could (similar to how spatial file types are detected) add a configuration specifying what type of file extensions are made available as metadata files, potentially defining a transformer/formatter function per file type...

tomkralidis: What about the following:
- enable pygeoapi to look for a filename with `.yml` extension, which is an MCF file
- add capability to pygeometa to export an MCF file as a STAC item, which would then provide the STAC item representation back to pygeoapi, like:
```python
        from pygeometa.core import read_mcf, MCFReadError

        pth = os.path.splitext(filepath)[0]
        try:
            md = read_mcf(pth)
            content = md.to_stac_item()  # not implemented yet
        except MCFReadError as err:
            LOGGER.warning(err)
```

pvgenuchten: Nice! 
But is it a responsibility of the provider? It means the provider can only be used by STAC. I hope we can also re-use this provider for OAPI-Records or even OAPI-Features at some point...

tomkralidis: Good point. For OARec and STAC API, wouldn't this imply a search index, from which we would have pygeoapi scrape the directory tree to build up an index (which would happen outside of this provider)?

pvgenuchten: Correct, my impression is that this provider caches the tree structure in memory, which seem ok for 100-ish files, but will be challenging if it is 1000's. Besides that, it now retrieves some metadata, such as dataset bounds at opening the record, which we would need prior to that for the whole tree if we want to enable spatial queries on the index, i already noticed this ogrinfo into files is very time & resource consuming, so for sure something you want to cache through restarts.

We then come to a discussion: is filesystem really a provider? Maybe it is even more a process (systemSpider) which prepares data for pygeoapi in an elastic or alternative index and both STAC and OAPI-Records engage with the index only.

pvgenuchten: Maybe we can use OAPI-Process to trigger the rebuild of the index 😄

tomkralidis:
@pvgenuchten FYI related pygeometa MCF STAC output PR at https://github.com/geopython/pygeometa/pull/130

pvgenuchten:
updated pr to use mcf syntax for sidecar file
note that https://github.com/geopython/pygeometa/pull/130 needs to be available prior to merging this
https://github.com/geopython/pygeometa/pull/130 returns json, would be better if it returns a dict

tomkralidis (changes requested):
@pvgenuchten : I see another sidecar file `tests/data/obs.md.yml`.  Can we support only the MCF format.  If there are gaps in the MCF format we can cover them there?

tomkralidis:
Note this PR also needs to resolve merged conflicts.

pvgenuchten:
> @pvgenuchten : I see another sidecar file `tests/data/obs.md.yml`. Can we support only the MCF format. If there are gaps in the MCF format we can cover them there?

correct, this is from the first PR, before mcf ws available

pvgenuchten:
refactored in #605

PR #564, opened 2020-11-01
prints stack trace of generic errors

pvgenuchten:
many times str(exception) only returns 'init'. 
With the full stacktrace the cause of the error (for example: fail to open dataset, because it has no projection) can be better evaluated.

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -40,6 +40,7 @@ import re
 import urllib.parse
 
 from dateutil.parser import parse as dateparse
+from traceback import print_exc as printtrace 
 import pytz
 
 from pygeoapi import __version__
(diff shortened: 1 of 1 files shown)


ricardogsilva (changes requested):
What about just using `logger.exception()` instead? Something akin to:

```python
In [1]: import logging
In [2]: logging.basicConfig()
In [3]: logger = logging.getLogger('tester')
In [4]: try:
   ...:     3 / 0
   ...: except ZeroDivisionError:
   ...:     logger.exception('There has been an error')
   ...: 
ERROR:tester:There has been an error
  File "<ipython-input-4-560924172f2a>", line 2, in <module>
    3 / 0
ZeroDivisionError: division by zero
```

pvgenuchten:
open to any improvement suggestions here... I'll gladly delegate to pythonistas... my main goal of the PR is to indicate that this fix greatly helped me while debugging my configuration

tomkralidis:
+1/echo @ricardogsilva to use standard logging here.

Issue #51, opened 2018-06-15
add support for items propertyname parameter


tomkralidis:
`propertyname=foo,bar` should return the response with just `foo` and `bar` properties.

francbartoli:
I'd like to have also the possibility to return only some of the properties without any geometry at all. Maybe to discuss with the spec team since in such cases the mime type would change into simple `application/json`

justb4:
GeoHealthCheck observes errors from the pygeoapi demo instances that use latest Docker Image from PR #565 related to this issue:

e.g. on URL: https://demo.pygeoapi.io/master/collections/utah_city_locations/items

```
[2020-11-02T13:17:12Z] {/usr/local/lib/python3.8/dist-packages/flask/app.py:1891} ERROR - Exception on /collections/utah_city_locations/items [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.8/dist-packages/flask_cors/extension.py", line 165, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.8/dist-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/pygeoapi/pygeoapi/flask_app.py", line 213, in collection_items
    headers, status_code, content = api_.get_collection_items(
  File "/pygeoapi/pygeoapi/api.py", line 921, in get_collection_items
    content = p.query(startindex=startindex, limit=limit,
TypeError: query() got an unexpected keyword argument 'select_properties'
```

Is there a config adaptation needed on demo server endpoints?

justb4:
No, I think some Providers, like OGR, are missing `select_properties=` param on `query()` function.

[...]

PR #566, opened 2020-11-02
fix ref

tomkralidis:
Fixes https://github.com/geopython/pygeoapi/issues/51#issuecomment-720468913


justb4 (approved):
Yes, passing a query object in later phase is more maintainable.

PR #572, opened 2020-11-17
add support for case insensitive query params

### LINKED ISSUE #514: Support case-insensitive query parameter keys
**Description**
pygeoapi appears to be sensitive to the casing of query parameter keys. For example, a query with `startindex=10` will succeed: https://demo.pygeoapi.io/master/collections/lakes/items?f=json&startindex=10

...but a query with `startIndex=10` will fail: https://demo.pygeoapi.io/master/collections/lakes/items?f=json&startIndex=10

Ideally, all clients should just use the "next" link verbatim. However, current pygeoapi behavior can negatively affects compatibility with older OGC-API clients.  The parameter `startIndex` was even part of the core specification at one point (pre-https://github.com/opengeospatial/ogcapi-features/commit/aa09714ce0ee4ff12a7cd40367c117051a96d100).

**Steps to Reproduce**
1. Load https://demo.pygeoapi.io/master/collections/lakes/items?f=json&startIndex=10
2. Receive InvalidParameterValue error

**Expected behavior**
I expected either `startIndex` or `startindex` to be acceptable.

**Environment**
 - OS: whatever powers https://demo.pygeoapi.io/master
 - Python version: whatever powers https://demo.pygeoapi.io/master
 - pygeoapi version: 0.9.dev0

ricardogsilva:
This PR adds support for case insensitive query parameters.

Implementation does not touch the core `pygeoapi.api.API` class at all. Changes have been made in the flask and starlette wrapper classes only.

Also added some unit tests to verify the behavior of the query param parsing.

### Flask changes

- Implement a custom parser for query parameters that stores parameter names as lower case, regardless of what is originally specified
- Then implement a custom flask request class that specifies our query params parser as its [parameter_storage_class](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Request.parameter_storage_class)
- Leverage a flask application's ability to [specify a custom request class](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.request_class) to make the pygeoapi flask app use our custom request class

### Starlette changes

Starlette changes are a bit more involved, since starlette does not support customizing the request class. As such this PR subclasses a bunch of starlette's internal classes in order to achieve similar functionality as the one done for the flask case:

- Subclass `starlette.applications.Starlette` in order to be able to specify a custom router class
- Subclass `starlette.routing.Router` in order to be able to specify a custom route class
- Implement a `PygeoapiRoute` class that calls the custom `pygeoapi_request_response()` request-maker function.
- Implement custom `PygeoapiStarltteRequest` and `PygeoapiQueryParams` classes that parse an incoming request's query parameters and store each parameter's name as lower case, regardless of what is originally specified in the request.

  

fixes #514


ricardogsilva:
On second thought, I think I'm going to take a different approach here (about to move status back to draft).

Instead of baking support for case-insensitive query params in each framework (flask, starlette) I'm instead thinking about introducing a common `PygeoAPIQueryParameters` class that we can then populate from the query params that are provided by whatever framework sits on top. This way we'll be able to ensure that query params are handled correctly and on a single framework-agnostic place (which seems more inline with pygeoapi's ethos)

At some point it might even make sense to extend this thinking into having a `PygeoapiRequest` class in order to have it get all relevant information from the framework (query params, headers, body) and then feed that to the code API class

@tomkralidis what are your thoughts on this?

tomkralidis:
@ricardogsilva this makes sense and will let us scale accordingly (Django, etc.)

tomkralidis:
@ricardogsilva closing for now. Let us know once this one's updated and feel free to submit/update accordingly.

PR #575, opened 2020-11-18
remove schema-org microdata, because it is now embedded as json-ld

### LINKED ISSUE #574: duplicated schema-org annotation & parsing failure
Some observations when validating https://demo.pygeoapi.io/master/collections/lakes in https://search.google.com/structured-data/testing-tool

**Duplicated annotation**
The pygeoapi templates still have microdata 'itemscope' attributes, combined with the embedded json-ld, it leads to duplicated parsing in validation tools. validator extracts the same dataset information twice

image

**Parse error**

The embedded json-ld throws a parse error on
```
 "@context": "https://schema.org/docs/jsonldcontext.jsonld",
```
error:
```
There was an invalid type in your JSON-LD.
```
This results in the object described being a:
image

In other implementations I notice this string seems to work fine:
``` 
"@context": "http://schema.org/",
```
In that case the type is recognized as:
image

pvgenuchten:
resolves #574

### DIFF
--- a/pygeoapi/templates/collection.html
+++ b/pygeoapi/templates/collection.html
@@ -5,16 +5,10 @@
 / <a href="./{{ data['id'] }}">{{ data['title'] }}</a>
 {% endblock %}
 {% block body %}
-    <section id="collection" itemscope itemtype="https://schema.org/Dataset">
-      <span itemprop="includedInDataCatalog" itemscope itemtype="https://schema.org/DataCatalog">
-        <meta itemprop="url" content="{{ config['server']['url'] }}/collections" />
-        <meta itemprop="name" content="{{ config['metadata']['identification']['title'] | striptags }}" />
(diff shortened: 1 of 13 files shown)


tomkralidis (approved):

alpha-beta-soup:
Just noting that the PR essentially reverts to the original, that had to be changed for _reasons_ https://github.com/geopython/pygeoapi/issues/444

alpha-beta-soup (approved):

alpha-beta-soup:
Incidentally, there's no reason microdata and json-ld can't coexist, they should just not have consistent information. http://www.thesempost.com/marking-content-json-ld-microdata-not-risk-google/

pvgenuchten:
co exist brings an additional maintenance effort

Thanx for merging, note that we should also fix https://github.com/geopython/pygeoapi/pull/576 else there is a risk of google not parsing any schem-org

PR #576, opened 2020-11-18
google expects plain schema-org url

### LINKED ISSUE #574: duplicated schema-org annotation & parsing failure

pvgenuchten:
This PR started out as a 'simple' namespace redefine. However it turns out to be related to a wider issue on the schema-org-as-json-ld domain. https://schema.org itself does not advertise a ld-context, but crawlers still require this url to be used as namespace. pyld, which is used to validate the json-ld, however requires a valid ld-context.

That's why the initial commit was reverted and a hacky-workaround is suggested, within the ajax-response the schema-org url is rewritten to fit the needs of search engine crawlers.

Issue does resolve #574

### DIFF
--- a/pygeoapi/linked_data.py
+++ b/pygeoapi/linked_data.py
@@ -60,7 +60,7 @@ def jsonldify(func):
         provider = meta.get('provider', {})
         ident = meta.get('identification', {})
         fcmld = {
-          "@context": "https://schema.org/",
+          "@context": "https://schema.org/docs/jsonldcontext.jsonld",
           "@type": "DataCatalog",
           "@id": cfg.get('server', {}).get('url', None),
           "url": cfg.get('server', {}).get('url', None),
--- a/pygeoapi/linked_data.py
+++ b/pygeoapi/linked_data.py
@@ -60,7 +60,7 @@ def jsonldify(func):
         provider = meta.get('provider', {})
         ident = meta.get('identification', {})
         fcmld = {
-          "@context": "https://schema.org/docs/jsonldcontext.jsonld",
+          "@context": "https://schema.org/",
           "@type": "DataCatalog",
           "@id": cfg.get('server', {}).get('url', None),
           "url": cfg.get('server', {}).get('url', None),
--- a/pygeoapi/templates/base.html
+++ b/pygeoapi/templates/base.html
@@ -69,7 +69,8 @@
           var head = document.getElementsByTagName('head')[0];
           var jsonld_datablock = document.createElement('script');
           jsonld_datablock.type = "application/ld+json";
-          jsonld_datablock.textContent = xhr.responseText;
+          //remove full context path, because search engines don't expect it here, pyld requires it.
+          jsonld_datablock.textContent = xhr.responseText.replace('docs/jsonldcontext.jsonld','');
           head.appendChild(jsonld_datablock);
         }
       };
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -154,7 +154,7 @@ def test_root_structured_data(config, api_):
     assert root['description'] == 'pygeoapi provides an API to geospatial data'
 
     assert '@context' in root
-    assert root['@context'] == 'https://schema.org/docs/jsonldcontext.jsonld'
+    assert root['@context'] == 'https://schema.org/'
     expanded = jsonld.expand(root)[0]
     assert '@type' in expanded
     assert 'http://schema.org/DataCatalog' in expanded['@type']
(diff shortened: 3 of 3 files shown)


tomkralidis:
+1 from me, thanks @pvgenuchten

alpha-beta-soup:
This implies reverting to the original, that had to be changed for _reasons_ https://github.com/geopython/pygeoapi/issues/444 and these reasons still apply (that's why the tests failed). It's still an upstream problem with pyld.

alpha-beta-soup (changes requested):
This change causes tests to fail, since they depend on pyld to parse the JSON-LD metadata. Either we fix pyld upstream, or use a different Python JSON-LD parsing library (or perhaps don't parse it at all).

pvgenuchten:
i was afraid of this... 
Or fix the schema-org validator...
as a workaround, maybe we can shorten the context url as part of the ajax request that introduces the snippet?

pvgenuchten:
just tested with google webmaster console
that tool seems to trigger only on the microdata and not on the embedded json-ld (in contradiction to structured data testing tool)
it could be related to the schem-org context url, but a bit puzzled/worried here about the impact of this

dblodgett-usgs:
@ksonda -- do you have any recollection of the nuances here?

ksonda:
I'm not familiar with pyld or why that would be affected, but Google will be retiring the structured data test tool, which didn't seem to pull in remote contexts other than schema.org. You could specify predicates manually though. The new tool, Google Rich Results, [will only parse microdata for resources of the following schema.org types](https://support.google.com/webmasters/answer/7445569):

image

pvgenuchten:
hi @dblodgett-usgs, 

nice to see the demo based on master is redeployed with the microdata being removed from last merged PR... helps testing

The json-ld is now only [injected using ajax](https://github.com/geopython/pygeoapi/blob/4e5cfd8e50ad760d8f2a958645ac0cb256d88236/pygeoapi/templates/base.html#L62-L73). 

I noticed structured-data-testing tool picks up the injected script code. 

image

However [rich results test](https://search.google.com/test/rich-results) does not find any embedded structured data.

image

Also yandex is not able to retrieve any 

image

I guess the ajax approach is not optimal, at least for some clients. Do we have an option to embed the json-ld using python?

some reading:
https://webmasters.stackexchange.com/questions/100935/why-wouldnt-google-be-able-to-read-ajax-generated-json-ld-schema-org-markup-par

pvgenuchten:
Note that the some of the problems may also be caused by https://github.com/geopython/pygeoapi/issues/577,

indeed:

image

but still curious why it finds a button and not a dataset

pvgenuchten:
I implemented the following test scenario:
- i made a basic html page available at http://pygeoapi.genuchten.net
- the page ingests schema org using ajax similar as pygeoapi
- i tested the page in google search console
- i updated context url to be plain schema.org
result: schema.org is parsed

image

- i updated the context to the [full context path](http://pygeoapi.genuchten.net/full.html)

then google complains:

image

alpha-beta-soup:
I still think the correct option is to attempt a PR to pyld so that it is able to use the appropriate schema.org URL; this would benefit other projects, and avoids the inconsistency between the direct JSON-LD response and the embedded data.

alpha-beta-soup:
https://github.com/digitalbazaar/pyld/pull/129

pvgenuchten:
@alpha-beta-soup i agree with the direction, however currently schema.org is broken since the removal of schema-org-microdata, i see this as a workaround until pyld would be optimised, are you ok to adopt it as a temporary workaround? Alternative option would be to revert the microdata removal.

alpha-beta-soup:
Not sure it's my call. My preference would be to keep the microdata, but I'm OK with the "temporary workaround" too as long as there's a corresponding issue/s that so it doesn't drop off the radar.

pvgenuchten:
@tomkralidis this is ready to be merged (fixing schema-org in crawlers), while keeping open the discussion with pyld

pvgenuchten:
thanx Tom

Issue #581, opened 2020-11-28
collection itemType can contain multiple types?


pvgenuchten:
when implementing #579 i was surprised to find only "features" as collection itemType.

Then i found that itemType is calculated from the 'default provider' at https://github.com/geopython/pygeoapi/blob/48d00d67d5dc7b9a3dcc2980aaa472c1fff2d630/pygeoapi/api.py#L340

In case a dataset is exposed as features and maps (typical behaviour?) i'd expect both types to be available in collection-itemType

I suggest we allow collection.itemType to contain an array of types.

tomkralidis:
As I understand it:

1. pygeoapi configuration `collection.type`: this tips off pygeoapi whether this is a collection (which can have 1..n providers), or something else (like a STAC collection or a process atm).  See [docs](https://docs.pygeoapi.io/en/latest/configuration.html#resources) for more information
2. `itemType`: this is rooted in expressing various item types that are bound to features or records (e.g. rasters do not have a type).  Note that [`itemType` in OAFeat is also a string](https://github.com/opengeospatial/ogcapi-features/blob/master/core/openapi/schemas/collection.yaml#L30).
3. pygeoapi configuration `resource.provider.type`: from docs:

```
The providers block is a list of 1..n providers with which to operate the data on. Each provider requires a type property. Allowed types are:
- feature
- coverage
- tile

A collection’s default provider can be qualified with default: true in the provider configuration. If default is not included, the first provider is assumed to be the default.
```

4. Note that one can also derive how they can interact with the data from a collection response's link relations.

Based on this, the first `provider.type` (or `provider.type=true`) should be interpreted as is what the data format/type really *is*.  So to render on a UI, the best bet is to either a. display all the `provider.type` values by assembling them all and including them in the dict that gets passed to the j2 template, or b. in the j2 template, looping through all the link relations and getting a flattened list of provider types.  Option a seems easier at first glance.

pvgenuchten:
Thanx Tom, yes if itemType is a string, then we don't have much options

I guess the open api document will be leading to api developers in exploring the API. However other users, such as incidental human beings, search engine crawlers, traditional GIS clients, etc may arrive at a collection endpoint without being aware of the open api document (or how to consume it). I find the follow-the-links a bit too technical and loose for ease-of-use. But this is probably an item to raise at oapi-common

Let me describe some scenario's and how oapi-common / pygeoapi would advertise them:

- a geojson file as oapif, made available also as tiles and maps
[...]

PR #582, opened 2020-11-28
WIP: itemType as an array of types

pvgenuchten:
suggestion to improve #581

image

currently still finds the default type, then adds 'tile' if available. needs improvement to add all types on array


pvgenuchten:
Seems itemType is a string field defined by a subset of the OAPI specifications, I'm going to consider alternative strategies to facilitate exploring the api.

closing this

PR #365, opened 2020-02-20
Process manager

alpha-beta-soup:
This is not yet ready for merging: apart from the WPS JSON binding spec still being in flux, I'm just making this PR prematurely so there's space for some discussion, review, criticism and hopefully some commits from interested people.

PR #588, opened 2020-12-07
Job submit form improvements

totycro:
These are just 2 minor UI niceties:
* Add an asterisk for mandatory input parameters in the form
* Allow entering float values if the data type is float

### DIFF
--- a/pygeoapi/templates/process.html
+++ b/pygeoapi/templates/process.html
@@ -141,7 +141,7 @@
                 <!-- TODO max/min for numbers -->
                 {% set typeMap = {'string': 'text', 'integer': 'number', 'float': 'number', 'boolean': 'checkbox', 'enum': 'radio'} %}
                 {% for input_ in data['inputs'] %}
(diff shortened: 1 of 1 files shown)


alpha-beta-soup:
Does `step="0.1"` disallowing typing more precise numbers, or does it just control the behaviour of the arrow buttons?

totycro:
It controls the arrow buttons and also validation. So you can input more precise numbers with the keyboard, but they are considered invalid and you can't actually submit a job like this.

I've just learned that there's also `step="any"`, which makes arrow buttons change the value by 1, but allow any manually entered value. What do you think about using that instead?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#step

alpha-beta-soup:
Yeah, that rings a bell. I played around with these but didn't settle on anything, as both of those leave something to be desired. On balance I think it would need to be "any" by default and perhaps open a discussion upstream about how an implementation shuld specifiy input-level precision options for number inputs.

tomkralidis:
Closing given simplification updates i #365

Issue #585, opened 2020-12-02
Revisit pygeoapi Dockerfile


justb4:
**Description**
Follow-up of issue #584. The pygeoapi Dockerfile needs to have a reproducible build. Based on Bullseye-slim Base Image is a moving target because of core package version changes. 

**Steps to Reproduce**

* build the Dockerfile and observe log output
* see that at the end (cleanup) `python3-gdal` package is auto-removed
* temp fix with #584 reinstalls but is a temp solution

**Expected behavior**
Expect clean and reproducible build with pinned package versions, for Debian, Python and Debian-Python packages.
 
**Screenshots/Tracebacks**
![image](https://user-images.githubusercontent.com/582630/100872628-5e843100-34a2-11eb-907b-f12801f93928.png)

**Environment**
 - OS: Bullseye Slim (Docker Base image)
 - Python version: 3.8 (hmm and `python3-gdal` installs Python 3.9!)
 - pygeoapi version: latest master

**Additional context**

Suggestions:

* earlier versions of the Dockerfile used Buster Slim as base image, which is more stable
* reason to move to Bullseye Slim was to have GDAL v3
* we may need to see if we can use a more stable base image and still have GDAL 3
* maybe use an OSGeo GDAL base image: https://hub.docker.com/r/osgeo/gdal/tags
* maybe a Ubuntu Small image with Ubuntu-GIS PPA and GDAL v3
* use Python virtualenv as to not depend on OS Python version fluctuations
* more version pinning

justb4:
Closing, as resolved via PR #598.

PR #596, opened 2020-12-23
Add missing distutils for Travis

kalxas:


justb4:
@kalxas are you aware of issue #585. This PR seems to address that as well. 
I am trying out the Dockerfile now.

kalxas:
I will have a look and provide updates

kalxas:
This mix of pip and debian packages is a recipe for trouble.
I would recommend to only use only debian packages for things that are available.

kalxas:
New PR #598

PR #598, opened 2020-12-25
Review dockerfile (#585)

kalxas:


kalxas:
Travis has passed for python 3.6, we are facing a pull limit from dockerhub

kalxas:
ok, rebuilding worked fine with Travis

tomkralidis (approved):

Dockerfile
```diff
@@ -57,14 +57,12 @@ LABEL maintainer="Just van den Broecke <justb4@gmail.com>"
...
-ARG ADD_PIP_PACKAGES=""
 
 # ENV settings
 ENV TZ=${TIMEZONE} \
 	DEBIAN_FRONTEND="noninteractive" \
-	DEB_BUILD_DEPS="tzdata build-essential libpython3-dev libgdal-dev python3-pip apt-utils curl git unzip" \
-	DEB_PACKAGES="locales libgdal28 python3-distutils python3-gdal libsqlite3-mod-spatialite ${ADD_DEB_PACKAGES}" \
-	PIP_PACKAGES="setuptools==49.3.1 greenlet==0.4.17 gunicorn==20.0.4 gevent==20.9.0 wheel==0.33.4 ${ADD_PIP_PACKAGES}"
+	DEB_BUILD_DEPS="tzdata build-essential apt-utils curl git unzip" \
+	DEB_PACKAGES="locales python3-pip python3-setuptools python3-distutils python3-gdal python3-psycopg2 python3-xarray python3-scipy python3-netcdf4 python3-rasterio python3-fiona python3-yaml python3-dateutil python3-pandas python3-tz python3-flask python3-unicodecsv python3-pyproj python3-click python3-greenlet python3-gevent python3-wheel gunicorn libsqlite3-mod-spatialite python3-elasticsearch ${ADD_DEB_PACKAGES}"
```
francbartoli, line 65: Does it make sense to move optional packages for providers into the `ADD_DEB_PACKAGES` arg variable? This could allow having a thinner base image

justb4 (changes requested):
Built and tested the Dockerfile, including most providers like OGR, SQLite but not ES and PostGIS. Went Ok. Python packages are now found:
- via `pip`  `/usr/local/lib/python3.8/dist-packages/` 
- via apt-get `python-*`: `/usr/lib/python3/dist-packages/`

The latter (`apt-get $DEB_PACKAGES`) are installed twice (line 78 and 101) as when uninstalling all deps, many like `libgdal`, GEOS etc are also removed by `apt`. Not sure why. Tried several variations but still the packages required for pygeoapi are removed.

It all works but feels a bit brittle. Also as Ubuntu-GIS packages, not version-pinned, may get unexpected updates.  

I also removed all lines dealing with locale and Timezone. That was from Debian. IMHO for Ubuntu one only needs to set `TZ` env var to have timezone right. Also `locale` command (when bashed into container) still shows empty and had no effect.

It is tough getting this Dockerfile right, builds take a long time. I agree with @francbartoli that we may try to supply a smaller base image and a full (all Providers) image.

Dockerfile
kalxas: yes it does

kalxas:
Debian packages might not be version pinned but changes in a distribution are very slow comparing to pip.

kalxas:
Purging one of the dev packages causes GDAL to be removed, I will try to figure out which one causes this.

justb4:
I've slimmed/built/tested the focal-based `Dockerfile`. It is now 766MB unpacked. Appears `build-essential` is not required. Think that package also triggered the excessive package removal. Need more testing but the main Providers (OGR, SQLite etc) are working. Can still be slimmed further as e.g. `psycopg2(-binary)` is installed twice via `apt` and `pip`.

I have attached my `Dockerfile` here, though could PR as well...

[Dockerfile.txt](https://github.com/geopython/pygeoapi/files/5751266/Dockerfile.txt)

kalxas:
Thanks @justb4 I have just pushed your changes

totycro:
Thanks for improving the Dockerfile! I'm running and developing pygeoapi exclusively in docker, so that's very important to me.

With the new version, it's now possible to run `pip install` in the image, which makes it easier to use for dev as well.

I was just wondering if the `apt-get upgrade` is really in the spirit of docker, since the image contains the old packages and after the upgrade also new ones, which increases the image size (and will do so more the more upgrades there are).
Also it means that dependencies are not pinned.
What do you think about instead of this approach we use fixed base tags such as `ubuntu:focal-20201106`?
Then we also get upgrades, but only when a dev actually changes the base image. There is a new ubuntu docker image tag more or less every month.

kalxas:
I agree that apt-get upgrade should be removed.

Dockerfile
kalxas: Done in https://github.com/geopython/pygeoapi/pull/598/commits/a638ef78c5941e661d273f0f00cd867babfb96cd

francbartoli (approved):

justb4:
Thanks folks for the useful comments! Played around with the latest Dockerfile above, making the suggested changes and added some extra, in summary:

* pin base image to `ubuntu:focal-20201106`
* no `apt-get upgrade`
* remove `pip install -r requirements-provider.txt` (as Providers are installed as `python-*` packages)
* dev build: make `pip install -r requirements-dev.txt` optional via build arg `BUILD_DEV_IMAGE` (`true|false` default `false`)
* had to move `flask_cors` from `requirements-dev.txt` to `requirements.txt`, think it belongs there (?).

The `prod` version image size is now down to 619MB (unpacked). The `dev` image is 736MB.

TODO: 

* the .travis config should build/test with the `dev` image. 
* more testing
* forgot why http://schemas.opengis.net/SCHEMAS_OPENGIS_NET.zip is downloaded/expanded, but may not add significantly
* may move `python3-pip` to DEB_BUILD_DEPS, maybe `python3-setuptools python3-distutils python3-wheel` as well? 
* may also via build arg build for starlette config

Not the most convenient, but I've attached my Dockerfile here:

[Dockerfile.txt](https://github.com/geopython/pygeoapi/files/5754243/Dockerfile.txt)

justb4 (changes requested):
See my suggested changes in [my comment](https://github.com/geopython/pygeoapi/pull/598#issuecomment-752556132).

kalxas:
Thanks @justb4 
I have just committed changes to address your comments.
I do not think we need to pin focal base image, the distribution does not change dramatically, there is very low risk for base image to break our image.
I have also removed the manager requirements, since they are also included in `ADD_DEB_PACKAGES`
`flask_cors` I think should remain to `requirements-dev`

justb4 (changes requested):
Almost there:

* no strong opinion as to (not) pin Focal base Image version
* delete line `pip3 install /requirements.txt`: all installed via `python3-*` deb packages and anyway via `setup.py` 
* Flask-Cors: still believe should be in `requirements.txt` (or in Dockerfile `python3-flask-cors` package, see below). CORS is enabled via server config, default is false/disabled. 
* my working Dockerfile (619 MB) attached

[Dockerfile.txt](https://github.com/geopython/pygeoapi/files/5761785/Dockerfile.txt)

kalxas:
We are good to merge

justb4:
All looks good, also all endpoints on https://demo.pygeoapi.io/. Compressed image down to 228.25 MB, also reasonable.
Thanks for your patience @kalxas ! Will close #585.

PR #599, opened 2020-12-29
add links, refactor job handling, HTML updates

tomkralidis:
cc @totycro

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -1903,13 +1903,21 @@ tiles/{{{}}}/{{{}}}/{{{}}}/{{{}}}?f=mvt'
                 link = {
                     'type': 'text/html',
                     'rel': 'collection',
-                    'href': jobs_url,
-                    'title': 'Collection of jobs for the {} process'.format(
-                        key),
+                    'href': '{}?f=html'.format(jobs_url),
+                    'title': 'jobs for this process as HTML',
                     'hreflang': self.config['server'].get('language', None)
                 }
-
                 p2['links'].append(link)
+
+                link = {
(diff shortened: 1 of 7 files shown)


tomkralidis:
FYI Travis failures are unrelated.

pygeoapi/api.py
```diff
@@ -2330,8 +2288,12 @@ def get_process_job_result(self, headers, args, process_id, job_id):
                                           data)
             return headers_, 200, response
 
-        content = json.dumps(job_output, sort_keys=True, indent=4,
-                             default=json_serial)
+        if mimetype in [None, 'application/json']:
```
totycro, line 2291: Here we have the requested `format_` and the `mimetype` which can be contradictory

E.g. if you request json with `?f=json`, but the mimetype is something else, you get something else.
However for html, there's the `if format_ == 'html'` above, which acts based on `format_` directly.

Btw in this case we can't differentiate between the cases where users request html via `?f=html` or if the browser just added the `Accept: text/html` header as it always does.

So when thinking about all those cases, we could argue that if a mimetype has been set, then it should always be returned, because it might be something that's not renderable as html or json.

The logic could be something like this:
```
if mimetype not in (None, 'application/json'):
    return data as mimetype
else:
    if format == 'json'
        return data as json
    else:
         return data as html
```
Then for json-data, you can choose the representation via the format (also in the browser by setting `?f=json`). For other kind of data which is not json-serialzable, we get just the data via the api and also in the browser, which makes sense because the data would usually not be renderable in html.
Or are there other cases where you could render the data in the html template?

totycro (approved):
Looks great!

PR #604, opened 2021-01-08
fallback to default HTML template if custom template does not exist

tomkralidis:
cc also @kngai for review/comment.

This PR provides a fallback for users implementing custom templating/themes, so users wishing to only change specific templates are able to do so without carrying/maintaining the entire template tree/set.

Once @pvgenuchten and @kngai test/verify, I will update the docs accordingly.


pvgenuchten:
This is a good idea, it prevents to require to duplicate all templates remotely, but only the ones with changes, which is easier to maintain

pvgenuchten (approved):
this works as expected, thanx!

tomkralidis:
Thanks for the idea @pvgenuchten!

kngai:
@tomkralidis +1. Working for me.

Issue #602, opened 2021-01-05
Deploying to AWS Lambda with Serverless - problem/solution


christophersbarrett:
I was having trouble with packaging and deploying to AWS Lambda, so while I labeled this as a bug out of haste I will document what it took to get it to work as well as put up an accompanying PR. 

Python: 3.8
OS: Ubuntu

Initial problem was running serverless from the aws-lambda subdirectory, the wsgi plugin would throw all the dependencies into the aws-lambda/ directory and create venv files that were sym links to the directories. 

To solve this, moving the serverless.yml file, and the pygeoapi-serverless.config.yml file to the root directory as well as updating the wsgi plugin config to: 
custom:
  wsgi:
    pythonBin: python3
    app: pygeoapi/flask_app.APP
    packRequirements: false
    textMimeTypes:
      - application/ld+json

PR #606, opened 2021-01-11
Serverless fixes

christophersbarrett:
Updated readme with new instructions on utilizing serverless IaC to build and deploy, updated pygeo config, updated serverless config so wsgi builds correctly

### DIFF
--- a/aws-lambda/pygeoapi-serverless-config.yml
+++ b/aws-lambda/pygeoapi-serverless-config.yml
@@ -32,11 +32,11 @@ server:
     bind:
         host: 0.0.0.0
         port: 80
-    url: /dev/
+    url: /
     mimetype: application/json; charset=UTF-8
     encoding: utf-8
     language: en-US
-    cors: true
+    cors: false
     pretty_print: true
     limit: 10
     # templates: /path/to/templates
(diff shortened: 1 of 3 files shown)


aws-lambda/pygeoapi-serverless-config.yml
```diff
@@ -32,11 +32,11 @@ server:
     bind:
         host: 0.0.0.0
         port: 80
-    url: /dev/
```
christophersbarrett, line 35: It could be I was doing this wrong, but the links were failing when relative to the /dev path.

aws-lambda/serverless.yml
```diff
@@ -31,26 +31,33 @@ service: pygeoapi
...
+    app: pygeoapi/flask_app.APP
+    packRequirements: false
     textMimeTypes:
       - application/ld+json
 functions:
   app:
     handler: wsgi_handler.handler
     events:
       - http: ANY /
-      - http: 'ANY {proxy+}'
```
christophersbarrett, line 54: This failed validation with: 
  at 'functions.app.events[1].http': value 'ANY {proxy+}' does not satisfy pattern /^(?:\*|(GET|POST|PUT|PATCH|OPTIONS|HEAD|DELETE|ANY) (\/\S*))$/i

christophersbarrett:
related to issue: https://github.com/geopython/pygeoapi/issues/602

tomkralidis:
Thanks @christophersbarrett !

Issue #133, opened 2019-05-22
configure which field to use as label (by lang)


pvgenuchten:
It is interesting to know which of the feature properties can be used as label (for a given language). This label can be shown on the feature-popup when hovering (in stead of id). Also it can act as the feature page title (and schema.org/name).

![image](https://user-images.githubusercontent.com/299829/58213573-01c38a00-7cf3-11e9-9a1f-9c531af3c450.png)

Inside dataset configuration I suggest to add a title-property configuration, which can have multiple values per language (in case the dataset contains properties per language)
```
title_property:
   en: "title"
   fr: "titre"
```

pvgenuchten:
Same on crumbs path and page title <head><title/> would be good to list the item title (in below case: "De Slokop"), in stead of repeating the dataset name

![image](https://user-images.githubusercontent.com/299829/70116153-27e2c300-1662-11ea-9c80-7212d20527af.png)

jorgejesus:
@pvgenuchten  I would first consider decouple this from language request, and have a ticket to properly check how languages could be implemented on all the code base structure. Something like:

```
languages: ["en","fr","nl","pt"]
```
And then for the dataset maybe implement standard language, and optional ?? 

```
lakes:
        type: collection
        lang: nl
```
Then we could have other system/code/implementation to deal with language translation. Mainly different templates based on language?

Then you could do:

```
title_property: NAAM
```
And on the template:

https://github.com/geopython/pygeoapi/blob/917f1c0a959ca9962ceedfcca04a08b9e1298418/pygeoapi/templates/item.html#L13

The on the api level fiddle with `id` 

https://github.com/geopython/pygeoapi/blob/917f1c0a959ca9962ceedfcca04a08b9e1298418/pygeoapi/api.py#L932

and/or with the template that returns the item page items

pvgenuchten:
Let's first set up the title_property element, and later improve to allow the element in multiple languages, for example by:

```YAML
title_property: Name
```

vs 

```YAML
title_property:
   en_us: name-en
   nl_nl: name-nl
   fr_ca: name-fr
```

or if the record is a json document, allow hierarchical

```YAML
title_property:
   en_us: name["en"]
   nl_nl: name["nl"]
   fr_ca: name["fr"]
```

Issue #100, opened 2019-04-26
add multilingual capability


tomkralidis:
Following on from https://github.com/opengeospatial/WFS_FES/issues/134, add capability to support multiple languages:

- support requests with `Accept-Language`
- support responses with `Content-Language`
- multilingual metadata in configuration
- handling `hreflang` in links in config
- other?

pvgenuchten:
A thing to consider, adding a browser language override mechanism as url-param &lang=fr (and a language switcher in html), requires any override value to be included in all links

I guess we'll use web locales (en/fr/nl/en-GB/fr-CA), not iso (dut/fre/eng)

Consider to let admin set a default language, if requested language is not in available languages

There should probably also be localization files for hardcoded strings (and documentation)

tomkralidis:
cc @francbartoli 

Good points @pvgenuchten.  One consideration is that of pygeoapi being integrated into arbitrary Python frameworks.  For example, all of pygeoapi (assuming werkzeug) is framework agnostic except for `pygeoapi.flask_app`.  So a Django wrapper need only create something like `pygeoapi.django_app` and implement the same thing obeying Django-isms.  Still some work, but IMHO tractable and pretty self-contained.

Having said this, implementing i18n in pygeoapi would need some thought w.r.t. being as agnostic as possible.  Options:

- text in Python code: implement standalone i18n in all Python code.  Works anywhere
- text in Jinja2 templates: this differs between Flask and Django, and perhaps other frameworks.  Any ideas?  Would Flask's `{{ _('text to translate') }}` approach work in non-Flask frameworks?

pvgenuchten:
Check out the [inspire with wfs3 discussion doc](https://webgate.ec.europa.eu/fpfis/wikis/download/attachments/299735841/DOC-5%20IRs-WFS3.0%20mapping%20discussion%20paper.pdf?version=1&modificationDate=1540280014657&api=v2) with proposals to add a Lang parameter

alpha-beta-soup:
Pingback to related https://github.com/geopython/pygeoapi/issues/317

The corollary is that it might be important to consider the degree of equivalence between `Accept-Language`, `hreflang` and the potential URL parameter `lang`; and to consider standardising values around [RFC 7231 (s5.3.5)](https://tools.ietf.org/html/rfc7231#section-5.3.5).

francbartoli:
> cc @francbartoli
> 
[...]

PR #608, opened 2021-01-15
indicate which attribute to use as label for an item

pvgenuchten:
This is a proposal for optimizing #133, it also brings in an aspect of #100. I wanted to combine the two to see the full benefit of it. In config you can assign one attribute as being the label/name/title of the item. 

image

This value is then used as label on the map and title of the item page

image

image

I also introduced the capability to set this value in a multilingual way

image

In case you ask the french representation of the page, the french attribute will be used as label

image

This required a language to be taken from accept-language or url-override, which had quite impact on the code, i can image people may want to refactor that aspect, please make suggestions

I considered also an approach in which i would add this aspect as part of the context, maybe @alpha-beta-soup can provide some suggestion there, maybe both approaches can co-exist?

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -175,7 +175,8 @@ default.
                 name: CSV
                 data: tests/data/obs.csv  # required: the data filesystem path or URL, depending on plugin setup
                 id_field: id  # required for vector data, the field corresponding to the ID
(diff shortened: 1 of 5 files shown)


pvgenuchten:
the language mechanism introduced in #470 seems defined in a more generic way, does it include the option to get language from header. maybe better to merge 470 first and adopt the approach?

PR #614, opened 2021-01-26
fix docker port, if it binds to 80 (previous line) you don't need a port here

pvgenuchten:


justb4 (commented):
Don't think this change is needed. These are two different port uses:

* the 'bind' port is for internal use or between containers
* the 'url' port is the external host port for outside access

Though both should match when run via docker or docker-compose.This config is also an example. See e.g. demo config: https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_master/local.config.yml, though external port is 443, that port is exposed by the Traefik proxy. On a localhost port 80 is usually already taken or not permitted, so we use 5000.

pvgenuchten:
if you run the image locally the bind port should match the configured port

alternatively we could document this startup option
```
docker run -it -p 5000:80 geopython/pygeoapi
```
and then tell them to visit localhost:5000

people with experience in traefic will know how to manage, however my audience are developers wanting to try out things locally (with limited python skills, then preconfigured docker is heaven (no gdal/sqlite hassle)

justb4:
Let's at agree, that there is no need to fix. 
Your commandline is already in the example, though it also mounts a pygeoapi config:

https://github.com/geopython/pygeoapi/blob/master/docker/examples/simple/run_pygeoapi.sh

It is only not good practice if Docker Image hardcodes opening a port on the host (security). 

But I agree, we should somehow make it easy and document for folks to run pygeoapi within a minute with Docker, first with defaults and then build-up with custom config and datastores, docker-compose and proxy-ing (Traefik, nginx, apache2 etc). Especially for Windows folks (not sure how Docker works there) That could be a new issue.

Issue #173, opened 2019-06-05
sparql query as a source for a wfs3 collection


pvgenuchten:
This issue wants to capture some of the gitter discussion we had at the wfs3 hackathon in nl.

There is generally 2 use cases for wfs
1. organisations with an ambition to publish some data on the web as a webservice
2. organisations having services online, but wanting to expand their audience by exposing to the geo community

In 1) the published concept is unique, In 2) the published feature is a representation of an object published elsewhere (it should use the external uri as id?).

My expectation is that we'll see more of 2) in future. We can then also think about having a wfs3 endpoint on the rdf sources, so it exposes the same content. This may be facilitated by having pygeoapi exposing the result of a sparql query on a triple store as wfs3.

Maybe we can use someting like https://rdflib.github.io/sparqlwrapper/ to retrieve the data?

dblodgett-usgs:
@ksonda and I have been discussing something very similar in the context of work on the geoconnex.us and geoconnex.ca feature repositories. 

The use case we have is where we want to add rich JSON-LD to a feature landing page and provide JSON-LD and other semantic formats for the feature its self. In this case, pyGeoAPI is hosting features that are identified by persistent URIs -- we want to query a triple store / graph database for linked data related to the persistent URI and template that content into an HTML header as well as some stand alone media types. 

My with PlantUML fascination is maybe becoming problematic, but here's something visual to describe the concept.

![image](https://user-images.githubusercontent.com/1492803/94195147-424b8e80-fe78-11ea-8bee-2323d2e622d6.png)

```
@startuml

package "App-Server" {
  [pyGeoAPI]
  [Linked\nData]
}

[Triple Store] - [Linked\nData] : SPARQL

[Spatial Features] - [pyGeoAPI] : SQL

database "Triple Store" {
}

database "Spatial Features" {
}

[pyGeoAPI] --> HTTP

[Linked\nData] ... [pyGeoAPI] : URI

@enduml
```

ksonda:
Not sure if this should be combined with #615 but the use case seems different enough.

Our idea is to use the capabilities introduced by #676 to use pygeoapi to publish lightweight geospatial collections that include URI attributes. 

The URI could then be used by a SPARQL provider to pipe in predicate-object pairs as feature properties into the item-level pygeoapi responses, with configurable list of predicates and the variable names they would take in the pygeoapi feature properties.

[...]

PR #615, opened 2021-01-27
GeoSPARQL

ldesousa:
Early support for GeoSPARQL as a data provider. For now only the `get` and `query` operations are supported, but there is more to come. The provider creates GeoSPARQL queries to retrieve RDF triplets from the GeoSPARQL endpoint in the JSON-LD format. These triplets are then transformed into a GeoJSON document to be returned by the API. The URIs of related ontological classes and other RDF resources are kept in the `properties` section of the GeoJSON document. 

Issues to note:

1. Modified the name of the GeoJSON provider to `geojson_prov` to avoid import conflicts. There might be a more elegant way around this.

2. The Docker set up is not fully tested. That can only happen when a new image with this provider is available from DockerHub. But the Virtuoso bit is tested.

3. So far only tested with Virtuoso. Will later test with Fuseki.


tomkralidis:
Thanks @ldesousa this is a great contribution.  RE: "more to come", should we mark this PR as WIP, or is this ready to review/merge?

ldesousa:
It is work in progress as I still want to introduce more features. But what is in this pull request is functional. I would suggest a merge into development so others can test it too.

ldesousa:
Travis is trying to run the tests, but there is no Virtuoso instance set up in the environment. Should something be done about the tests? Or can the PR be merged like this?

tomkralidis:
In general, if there are tests, then they should have the appropriate setup.

ksonda:
This is highly related with #469 and #173 . Ideally, you could specify a URI field populated with external/global persistent identifiers that resolves to the pygeoapi /items/{feature id} page, which is populated by the results of querying the GeoSPARQL endpoint as demonstrated here.

pvgenuchten:
Hi @ldesousa, I'm curious for your use case here. From what I understand SPARQL is very powerfull via its capability to compose queries on complex data, but it is quite poor on performance. For that reason app developers that build on triple stores, usually query a structured subset of a graph and store it in a fast responding cache, such as a database or search index, before exposing it via an API.

ldesousa:
@pvgenuchten The GeoSPARQL provider is independent of the back-end. Whether it is a true triple store or something else, as long as it can reply to GeoSPARQL queries it will work. Currently I am working with Virtuoso, that internally stores triples in a relational database. So far I had no performance related issues.

GeoJSON, JSON-LD and GeoJSON-LD are all independent specifications that do not square easily with one another. Moreover, GeoJSON-LD has no institutional backing and can't be regarded exactly as a standard. JSON-LD is by itself enough to convey geo-spatial information using the GeoSPARQL ontology. The bit of innovation needed in this field is the extension of GeoSPARQL to raster data.

pvgenuchten:
I just had a irl meeting to discuss geosparql backend with @ldesousa 
My doubts if sparql will provide enough performance for this use case have not gone away
(alternative would be to insert daily the sparql-response into a postgis table (or elastic) as id, time, spatial, json-ld-blob, and use that as a source for pygeoapi)
@ldesousa raised that for them creating proper json-ld from a postgres backend was quite a challenge (sparql provides that efficiently). And also that the triple store may contain objects in multiple evolutions of the (soil) standard.
We also considered the case that at some point users may use pygeoapi to post new or updated items to the collection using ogc-api-features, these updates would then be included in the triplestore.

Enough potential to make this an interesting direction of development.

ldesousa:
Hi folks, any news on this PR?

tomkralidis:
@ldesousa can you rebase, after which I'm guessing this is ready for a review?

ldesousa:
> This suggests that item-level JSON-LD response for feature collections might have an option for GeoJSON-LD (default) and some kind of GeoSPARQL representation. Since #676 the optional JSON-LD using schema:geo vocabulary for point-type geometry is implemented, but now I'm wondering if GeoSPARQL represenation should be another option or just preferable overall, and support non-point geometry as well.

I would strongly advise against pursuing the GeoJSON-LD route. It is not a standard and is unlikely to become one soon. 

GeoSPARQL is an ontology, not an encoding scheme. A resource making use of GeoSPARQL can be encoded as XML/RDF, Turtle, JSON-LD, etc. GeoSPARQL is not meant as a "representation".

ksonda:
Thanks @ldesousa 

GeoJSONLD is the default behavior so #676 left it in for now, but we agree that standard JSON-LD should be used to be used to communicate geometry using existing ontologies. What I mean by alternative "representation" is that geometry in JSON-LD generated by pygeoapi could be the geosparql ontology or the schema.org geo types.

ldesousa:
GeoJSON-LD was not the default behaviour when I submitted this pull request. Neither was it JSON-LD. At the time all providers were returning GeoJSON objects. Unless we are speaking of different things.

ksonda:
appending ?f=jsonld to /collections/{collection}/items/{item id}, or requesting application/ld+json, returns a json-ld document, in a GeoJSONLD format since at least May 2020. We added a configuration option in #676 for this to instead return a JSON-LD document with schema.org type geometry.

But I'm asking whether we should add an option to return that JSON-LD document with geosparql geometry.

pvgenuchten:
I'm interested to keep this PR in focus. Especially in academia there is a lot of triple data which is hard to consume by spatial clients. If pygeoapi can provide a proxy into the triple stores, this is of high interest. I guess we should have a look at the current conflicts.

ksonda:
@webb-ben we should contribute as bendwidth allows. I'm wondering if the configuration should allow more complex sparql queries to allow more than one "hop" to enrich the geometry with relevant properties as desired

ldesousa:
After 18 months it is better to close this pull request. The functionality I proposed here is now implemented in Prez, and in a more sofisticated way. If you ever revisit the possibility of adding a GeoSPARQL triple store as features provider, I would advise you to follow a similar approach, making use of the ogcldapi OWL profile.

PR #618, opened 2021-02-13
update STAC MCF support/dependencies/docs

tomkralidis:
cc @pvgenuchten


pvgenuchten (approved):
Thanx for adding docs, I missed that aspect indeed

PR #635, opened 2021-02-19
resolves small oarec regression on html output

pvgenuchten:
go to items doesn't display on html


justb4 (approved):
Looks good enhancement and solves bug: `itemType` is indeed all lowercase like `'feature'` or `'record'` so fragment `{% if data['itemType'] == 'Feature' %}` is currently missing, also on demo site. @tomkralidis can you look and hopefully merge?

So overall: LGTM.

Issue #639, opened 2021-02-20
update sortby parameter handling


tomkralidis:
**Is your feature request related to a problem? Please describe.**
OARec and STAC have convened on updating the `sortby` parameter as per https://docs.opengeospatial.org/DRAFTS/20-004.html#clause-sorting

**Describe the solution you'd like**

- `sortby=myprop:D`, `sortby=-myprop` # descending
- `sortby=myprop:A`, `sortby=+myprop` # ascending
- `sortby=myprop`, `sortby=myprop`. # default ascending

**Describe alternatives you've considered**
NA

**Additional context**
NA

PR #640, opened 2021-02-20
update sortby parameter handling

tomkralidis:
This PR implements #639

- `sortby=myprop:D`, `sortby=-myprop` # descending
- `sortby=myprop:A`, `sortby=+myprop` # ascending
- `sortby=myprop`, `sortby=myprop`. # default ascending


pvgenuchten:
I'm not in favour of this approach, this PR does not make pygeoapi understand ?sortby=+title, but i don't blame pygeoapi for it. ?sortby=%2Btitle works as expected, but is very uncomfortable from a usage perspective

pvgenuchten (approved):
while we monitor the discussion at the ogc api records side, i'm ok to merge this, make sure we don't use/generate any sortby=+title links ourselves (it doesn't work), sortby=title gives the same result

PR #649, opened 2021-02-25
update queryables response and allow enums in provider fields (#647)

### LINKED ISSUE #647: configure an enum for query values
We have a case that a query parameter (theme) can only be filled with a limited set of values

OpenApi facilitates to define an [enum property on a query parameter](https://swagger.io/docs/specification/describing-parameters/), can we support this feature so a provider can return an enum as part of the get_field method for queryable fields, to be included in the openapi document response?

tomkralidis:
Fixes #647

### DIFF
--- a/docs/source/tour.rst
+++ b/docs/source/tour.rst
@@ -62,7 +62,7 @@ Collection queryables
 
 http://localhost:5000/collections/obs/queryables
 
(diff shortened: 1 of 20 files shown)


GeoSander:
Looking good Tom, and also seems to work for us. Thanks for the rapid implementation!

Issue #554, opened 2020-10-11
Error messages are returned in json only


pvgenuchten:
Error messages are returned in json, even if the requested response encoding is html (probably also for xml, gpkg, ...)
For example https://demo.pygeoapi.io/master/collections/lakes/items?f=html&NAME=L.%20Erie returns json

PR #650, opened 2021-02-26
refactor exceptions

tomkralidis:
This PR refactors exception handling, addressing #554


pvgenuchten:
image

really nice!

Any change to stay in the context of items, so the user can browse back to the collection through the crumbspath? maybe refactore the crumbspath based on current url? I can have a look at that in a new iteration.

pvgenuchten (approved):

PR #651, opened 2021-02-26
adds twitter and opengraph tags to html encoding

pvgenuchten:
This brings in an aspect of improved vizualisation of pygeoapi items in social media.
Whenever you place a url in platforms such as linkedin, twitter, slack. The client heads out to find any microdata in that url so it can vizualize the url as a nice card. by default the settings are added from main config, unless you arrive at a collection or record endpoint, in that case the properties of the collection/record will be used to populate the microdata.

image
_example of rich snippet of INSPIRE portal in Mattermost client_

### DIFF
--- a/pygeoapi/templates/_base.html
+++ b/pygeoapi/templates/_base.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="{{ config['server']['encoding'] }}">
-    <title>{% block title %}{{ config['metadata']['identification']['title'] }} -{% endblock %}</title>
+    <title>{% block title %}{% endblock %}{% if not self.title() %}{{ config['metadata']['identification']['title'] }}{% endif %}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="language" content="{{ config['server']['language'] }}">
     <meta name="description" content="{{ config['metadata']['identification']['title'] }}">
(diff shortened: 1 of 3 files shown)


tomkralidis:
@pvgenuchten can you update this PR given that we now use bootstrap?

pvgenuchten:
rebased on latest

pygeoapi/templates/_base.html
```diff
@@ -20,8 +21,14 @@
       <link rel="{{ link['rel'] }}" type="{{ link['type'] }}" title="{{ link['title'] }}" href="{{ link['href'] }}"/>
       {% endif %}
     {% endfor %}
+    <link rel="shortcut icon" href="{{ config['server']['url'] }}/static/img/favicon.ico" type="image/x-icon">
+    <link rel="stylesheet" href="https://unpkg.com/mini.css@3.0.1/dist/mini-default.min.css">
```
tomkralidis, line 25: Aren't we using bootstrap now (instead of mini.css)? https://github.com/geopython/pygeoapi/blob/master/pygeoapi/templates/_base.html

pygeoapi/templates/_base.html
```diff
@@ -1,8 +1,9 @@
 <!doctype html>
+{% set og={} %}
```
tomkralidis, line 2: `{% set og = {} %}`

pygeoapi/templates/collections/collection.html
```diff
@@ -1,5 +1,11 @@
 {% extends "_base.html" %}
 {% block title %}{{ super() }} {{ data['title'] }} {% endblock %}
+{% if 'description' in data %}
```
tomkralidis, line 3:
```
{% block desc %}{{ data.get('description', '') | truncate(250) }}{% endblock %}
{% block tags %}{{ data.get('keywords', []) | join(',') }}{% endblock %}
```

pygeoapi/templates/collections/items/item.html
```diff
@@ -1,11 +1,16 @@
 {% extends "_base.html" %}
 {% set ptitle = data['properties'][data['title_field']] or ("Item {}".format(data['id'])) %}
+{% block desc %}{{ data.get('properties',{}).get('description',{}) | string | truncate(250) }}{% endblock %}
```
tomkralidis, line 3:
```
{% block desc %}{{ data.get('properties', {}).get('description',{}) | string | truncate(250) }}{% endblock %}
{% block tags %}{{ data['properties'].get('themes', [{}])[0].get('concepts', []) | join(',') }}{% endblock %}
{% if 'thumbnailUrl' in data['properties'] %}
  {% block thumb %}{{ data['properties']['thumbnailUrl'] }}{% endblock %}
{% endif %}
```

Then again, isn't `thumbnailUrl` a bit specific?

tomkralidis (changes requested):

pygeoapi/templates/_base.html
pvgenuchten: Indeed, must be a problem in the rebase, ll check

pygeoapi/templates/collections/items/item.html
pvgenuchten: Thumbnail url indeed is only relevant for metadata records, probably better to remove it here

pvgenuchten:
build on python3.6 fails with a sensorthings frost server error, not related to this branch

PR #655, opened 2021-03-03
add schema.org/Dataset representation for items of ogc api records

pvgenuchten:
This PR adds a representation of the record schema of OGC API Records as json-ld.
Most of the field mappings are managed from the ld-context (and could probably be managed from the existing context configuration. However the 'properties' object of geojson itself, and the associations/distributions require specific changes. The frequent use of 'type' as a property name is not optimal for a ld conversion.

currently this is a work in progress
- [ ] finalization of the record schema (how will associations be constructed)
- [ ] populating more properties in record.yml, such as thumbnail, license, encodingformat, creator (contactPoint)
- [ ] tests and flake8
- [x] tested in GGL structured data testing tool and is functional (for jsonld as well as html encoding)

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -46,8 +46,8 @@ from dateutil.parser import parse as dateparse
 import pytz
 
 from pygeoapi import __version__
-from pygeoapi.linked_data import (geojson2geojsonld, jsonldify,
-                                  jsonldify_collection)
+from pygeoapi.linked_data import (geojson2geojsonld, geojson2recordjsonld,
+                                  jsonldify, jsonldify_collection)
 from pygeoapi.log import setup_logger
 from pygeoapi.process.base import (
     ProcessorExecuteError
@@ -1028,7 +1028,11 @@ class API:
             return headers_, 200, content
         elif format_ == 'jsonld':
             headers_['Content-Type'] = 'application/ld+json'
-            content = geojson2geojsonld(self.config, content, dataset)
+            itemType = self.config['resources'][dataset].get('providers', [{}])[0].get('type','')
+            if itemType == 'record':
+                content = geojson2recordjsonld(self.config, content, dataset)
+            else:
+                content = geojson2geojsonld(self.config, content, dataset)
             return headers_, 200, content
 
         return headers_, 200, to_json(content, self.pretty_print)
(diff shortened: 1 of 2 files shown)


pygeoapi/linked_data.py
```diff
@@ -203,3 +203,100 @@ def geojson2geojsonld(config, data, dataset, identifier=None):
...
+                        *[config['server']['url'], dataset]),
+        **fts
+    }
+    
+    for i, feature in enumerate(fts['features']):
+        feature.pop('type', '')
+        feature.pop('links', '')
+        feature['id'] = '{}/collections/{}/items/{}'.format(
+                        *[config['server']['url'], dataset, feature['id']])
+        crds = feature.get('geometry',{}).get('coordinates',[[]])[0]
```
alpha-beta-soup, line 269: You could replace `get` with `pop` and remove the `pop` on l280. Pop returns the value (or a default) value in the same way as get, but also removes it.

pygeoapi/api.py
```diff
@@ -1028,7 +1028,11 @@ def get_collection_items(self, headers, args, dataset, pathinfo=None):
             return headers_, 200, content
         elif format_ == 'jsonld':
             headers_['Content-Type'] = 'application/ld+json'
-            content = geojson2geojsonld(self.config, content, dataset)
+            itemType = self.config['resources'][dataset].get('providers', [{}])[0].get('type','')
```
tomkralidis, line 1031: Can be written as:

```python
try:
    _ = get_provider_by_type(se.f.config['resources'][dataset]['providers'], 'record')
    content = geojson2recordjsonld(self.config, content, dataset)
except ProviderTypeError:
    content = geojson2geojsonld(self.config, content, dataset)

File: pygeoapi/api.py [hunk 3]
```diff
@@ -1152,8 +1156,13 @@ def get_collection_item(self, headers_, format_, dataset, identifier):
             return headers_, 200, content
         elif format_ == 'jsonld':
             headers_['Content-Type'] = 'application/ld+json'
-            content = geojson2geojsonld(
-                self.config, content, dataset, identifier=identifier
+            itemType = self.config['resources'][dataset].get('providers', [{}])[0].get('type','')
```
tomkralidis, line 1159: Same here as previous comment.

pygeoapi/linked_data.py
```diff
@@ -203,3 +203,100 @@ def geojson2geojsonld(config, data, dataset, identifier=None):
...
+
+    :param config: dict of configuration
+    :param data: dict of data:
+    :param dataset: dataset identifier
+    :param identifier: item identifier (optional)
+
+    :returns: string of rendered JSON (GeoJSON-LD)
+    """
+    isCollection = identifier is None
+    if isCollection:
```
tomkralidis, line 222: Prefer `snake_case` for variables.

tomkralidis (changes requested):

tomkralidis:
@pvgenuchten any update on this PR or should we close for now?

pvgenuchten:
Ah, I was not aware this was still pending, let me have a look asap

PR #656, opened 2021-03-05
Multilingual support

GeoSander:
Enables multilingual support for pygeoapi (core and plugins). Refer to issue #100 for discussion.

### Work done
- [x] Server-independent language check (from request `l` parameter or `Accept-Language` header)
- [x] Make providers, processes and formatters language-aware
- [x] Add translate functionality for JSON language structs based on the current language setting (best match)
- [x] Support responses with `Content-Language` header
- [x] Multilingual metadata in configuration
- [x] Add documentation 
- [ ] Handle `hreflang` in links in configuration

I've decided to push the last item to a next PR. This PR is already big enough...
The generation/processing of all API links should be done in a separate function, imho (see "Upcoming PRs" below).

### What does it do?
Users can now append a `l=fr` parameter to a pygeoapi URL and pygeoapi will return the requested page in French (for example). If the user connects to pygeoapi using a browser where the language has been set to French, the requested page will automatically be rendered in French (without the `l=fr` parameter), because the browser sent an `Accept-Language` header along with the request. However, this can always be overridden again using the `l` query parameter.

### Remarks
- This PR is quite large and mainly impacts the API core (and not so much the plugins). Most of the language support stuff revolves around the new `l10n` (localization) module, which turns the requested language (e.g. "fr", "fr-CH" or "fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5" into a best matching Babel `Locale` object. These `Locale` objects integrate well with Jinja2 templates using the `i18n` extension;
- The only text that is translatable at this moment, are the text strings (names, titles, descriptions, etc.) in the server configuration. The `pygeoapi-config.yaml` now contains some example language structs for US English and Canadian French (more translations coming up);
- In order to efficiently pass on language settings, I decided to make all API method calls more consistent, so `api.py` has undergone quite a lot of refactoring. All methods that require `Request` object data are now decorated by the `@pre_process` function, which converts the incoming Flask/Starlette `Request` into a (newly introduced) `APIRequest` instance. This class processes headers and query parameters to extract the required properties for the API methods. This results in less boilerplate and cleaner API method signatures, while still being explicit;
- Developers of third-party providers and other plugins should add a `requested_locale=None` argument to the `__init__` method of their plugin, and pass the `requested_locale` to the `super().__init__()` call as a second argument. This is all that needs to be done to make the plugin work again and to make it language-aware. However, it is up to the developer to implement the logic to query/process data and return it in the requested language. Please refer to the provided [Sphinx documentation](https://github.com/GeoSander/pygeoapi/blob/multilingual/docs/source/language.rst) for more info.

### Follow-ups
This PR provides the foundation to make pygeoapi language-aware. However, for full language support, there are still some things that need to be done.

Upcoming PRs should address:
- Links generated by pygeoapi: the links underneath each collection item (for example) should become language-aware, i.e. internal link URLs should be appended with a `l=<language>` query parameter if the current page also was requested using that parameter (note: the `add_locale` function in the `l10n` module can be used for that.
Internal links should also set the `hreflang` property to the user-specified language. For external links, this only applies if the server speaks that language (hard to figure out).
- HTML templates: all hard-coded translatable strings in the templates need to be stored in separate files and processed by `pybabel`. so they can be managed by a translation tool (e.g. Transifex);
- The Jinja2 `i18n` extension needs to be configured and the HTML templates need to be modified so that Jinja2 and Babel/gettext can inject the translations;
- All core plugins (e.g. providers) are now language-aware, but the ones that *can* support multiple languages (i.e. the plugin backend supports it) should still implement functionality for it. The `l10n` module provides the necessary tools to do so;
- Other suggestions...?

Some useful links concerning Jinja2/Babel/Transifex:
- https://www.mattlayman.com/blog/2015/i18n/
- https://phrase.com/blog/posts/i18n-advantages-babel-python/#Integration_with_Jinja2_templates
- https://jinja.palletsprojects.com/en/2.11.x/templates/#i18n-in-templates

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -29,6 +29,7 @@ pygeoapi |release| documentation
    data-publishing/index
    plugins
    html-templating
+   language
    development
    ogc-compliance
    contributing
--- a/docs/source/language.rst
+++ b/docs/source/language.rst
@@ -1,0 +1,216 @@
+.. _language:
+
+Multilingual support
+====================
+
+pygeoapi is language-aware and can handle multiple languages if these have been defined in pygeoapi's configuration (see `maintainer guide`_).
+Plugins (e.g. providers) can also handle multiple languages if configured. These may even be different from the languages that pygeoapi
+supports. Out-of-the-box, pygeoapi "speaks" English.
+
+The following sections provide more information how to use and set up languages in pygeoapi.
+
+End user guide
+--------------
+
+There are 2 ways to affect the language of the results returned by pygeoapi, both for the HTML and JSON(-LD) formats:
+
+1. After the requested pygeoapi URL, append a ``l=<code>`` query parameter, where ``<code>`` should be replaced by a well-known language code.
+   This can be an ISO 639-1 code (e.g. `de` for German), optionally accompanied by an ISO 3166-1 alpha-2 country code (e.g. `de-CH` for Swiss-German).
+   Please refer to this `W3C article <https://www.w3.org/International/articles/language-tags/>`_ for more information or
+   this `list of language codes <http://www.lingoes.net/en/translator/langcode.htm>`_ for more examples.
+   Another option is to send a complex definition with quality weights (e.g. `de-CH, de;q=0.9, en;q=0.8, fr;q=0.7, \*;q=0.5`).
+   pygeoapi will then figure out the best match for the requested language.
+
+   For example, to view the pygeoapi landing page in Canadian-French, you could use this URL:
+
+   https://demo.pygeoapi.io/master?l=fr-CA
+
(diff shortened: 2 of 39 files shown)


GeoSander:
Build should actually work, but 1/2 failed because of:
image

jodygarnett:
> Note: build should actually work, but 1/2 failed because of:
> 
> You have reached your pull rate limit

I note that the GEOS and PostGIS projects are using osgeo docker repository https://repo.osgeo.org/#browse/search/docker for test images (no pull rate limit there).

tomkralidis:
Closing in lieu of #664

PR #657, opened 2021-03-11
add support for OpenAPI document as JSON

tomkralidis:
cc @jvanulden

This PR provides functionality for generating the OpenAPI document as JSON (in addition to the default YAML output).  

Here's why:

In pygeoapi, we use a modified YAML parser which includes support for environment variables in the YAML itself.  See https://github.com/geopython/pygeoapi/blob/master/pygeoapi/util.py#L105

This works great in most current cases, but with huge YAMLs this suffers performance at runtime.  This is because the modified YAML parser we use in `/openapi?f=json` mode returns JSON, not YAML, hence a transformation on the fly.

As a result, this PR updates `pygeoapi generate-openapi-description` to accept an `-f json` parameter (default is still YAML), where a JSON would be written to disk.

Then, pygeoapi will check at runtime if `PYGEOAPI_OPENAPI` is a YAML or JSON file via checking for the file extension.  If the former, then current functionality.  If the latter, then just dump the JSON file as is with no transformation (faster).

### DIFF
--- a/docs/source/administration.rst
+++ b/docs/source/administration.rst
@@ -27,6 +27,15 @@ This will dump the OpenAPI document as YAML to your system's ``stdout``.  To sav
 
    pygeoapi generate-openapi-document -c /path/to/my-pygeoapi-config.yml > /path/to/my-pygeoapi-openapi.yml
 
+To generate the OpenAPI document as JSON, run:
+
+.. code-block:: bash
+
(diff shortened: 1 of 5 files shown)


francbartoli (approved):

jvanulde:
Just wondering if the JSON should be generated by default in addition and in parallel with the YAML.

tomkralidis:
Thanks @jvanulde.  The flag should allow for composable workflow either way (albeit offline and not at runtime).

PR #662, opened 2021-03-17
improve MIME type handling for raw payload responses

tomkralidis:
This PR fixes MIME type handling for custom processes that return non-JSON responses.  pygeoapi requires this information to correctly set the response MIME type when OAProc clients invoke a job with `"response": "raw"` in the request.

As a result custom processes' `execute` functions must return a tuple of the mime type and the response payload.

cc @alpha-beta-soup @totycro

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -2014,7 +2014,7 @@ tiles/{{{}}}/{{{}}}/{{{}}}/{{{}}}?f=mvt'
 
         try:
             LOGGER.debug('Executing process')
-            outputs, status = self.manager.execute_process(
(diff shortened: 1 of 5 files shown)


tomkralidis:
Travis failures unrelated.

totycro:
Nice, this basically fixes a TODO in pygeoapi-kubernetes-papermill: https://github.com/eurodatacube/pygeoapi-kubernetes-papermill/blob/main/pygeoapi_kubernetes_papermill/kubernetes.py#L250

I didn't get around to implement this myself because it was not a priority for my use case, but this PR really fills a gap.

PR #630, opened 2021-02-17
Add support for the django framework

francbartoli:
Implement #586

Pending tasks:

- [x] Rebase with master
- [x] Add support for Django 3 and 4
- [x] Align the current endpoints in the Django application
- [x] Fix the statics for the HTML pages
- [x] Make sure the Django application is pluggable when the pygeoapi package is installed
- [x] Add documentation

PR #664, opened 2021-03-21
Multilingual support (alternative)

GeoSander:
This PR is similar to #656, but there are some important changes:

- The `l` HTTP query parameter to request a certain language has been renamed to `lang`.
- The code to determine the supported provider language has been moved from the provider base classes to the `API` class.
- The provider methods that might benefit from language-awareness need to add a `**kwargs` (or `language=None`) parameter.

Updated PR text from #656:

### What does it do?
Users can now append a `lang=fr` parameter to a pygeoapi URL and pygeoapi will return the requested page in French (for example). If the user connects to pygeoapi using a browser where the language has been set to French, the requested page will automatically be rendered in French (without the `lang=fr` parameter), because the browser sent an `Accept-Language` header along with the request. However, this can always be overridden again using the `lang` query parameter.

### Remarks
- This PR is quite large and mainly impacts the API core (and not so much the plugins). Most of the language support stuff revolves around the new `l10n` (localization) module, which turns the requested language (e.g. "fr", "fr-CH" or "fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5" into a best matching Babel `Locale` object. These `Locale` objects integrate well with Jinja2 templates using the `i18n` extension (see _Follow-ups_ section below);
- The only text that is translatable at this moment, are the text strings (names, titles, descriptions, etc.) in the server configuration. The `pygeoapi-config.yaml` now contains some example language structs for US English and Canadian French (more translations coming up);
- In order to efficiently pass on requests (and the requested language) from Flask/Starlette, `api.py` has undergone quite a lot of refactoring. All API methods that require `Request` object data are now decorated by the `@pre_process` function, which converts the incoming Flask/Starlette `Request` into a newly introduced `APIRequest` instance. This class processes headers and query parameters to extract the required properties for the API methods. This results in less boilerplate and cleaner API method signatures, while still being explicit. Please refer to the `APIRequest` docstring for more information and examples;
- Developers of third-party providers and other plugins should add `language=None` or `**kwargs` to any `query` , `get` and 
 `get_metadata` method implementations of their plugin. This is all that needs to be done to resolve any `TypeError` (i.e. `got an unexpected keyword argument 'language'`) and to make the plugin language-aware. However, it is up to the developer to implement the logic to query/process data and return it in the requested language. Please refer to the provided [Sphinx documentation](https://github.com/GeoSander/pygeoapi/blob/multilingual/docs/source/language.rst) for more info.

### Follow-ups
This PR provides the foundation to make pygeoapi language-aware. However, for full language support, there are still some things that need to be done.

Upcoming PRs should address:
- Links generated by pygeoapi: the links underneath each collection item (for example) should become language-aware, i.e. internal link URLs should be appended with a `lang=<language>` query parameter if the current page also was requested using that parameter (note: the `add_locale` function in the `l10n` module can be used for that). Internal pygeoapi links should also set the `hreflang` property to the user-specified language.
- HTML templates: all hard-coded translatable strings in the templates need to be stored in separate files and processed by `pybabel`. so they can be managed by a translation tool (e.g. Transifex);
- The Jinja2 `i18n` extension needs to be configured and the HTML templates need to be modified so that Jinja2 and Babel/gettext can inject the translations;
- Provide a schema for the pygeoapi configuration YAML that clearly specifies which strings support language structs.

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -458,11 +458,8 @@ class API:
                                         'type', 'stac-collection'):
                 fcm['stac'] = True
 
-            fcm['keywords'] = l10n.translate(
-                self.config['metadata']['identification']['keywords'],
-                request.locale)
-
-            content = render_j2_template(self.config, 'landing_page.html', fcm)
+            content = render_j2_template(self.config, 'landing_page.html', fcm,
+                                         request.locale)
             return headers, 200, content
 
         if request.format == 'jsonld':
@@ -491,7 +488,8 @@ class API:
             data = {
                 'openapi-document-path': path
             }
-            content = render_j2_template(self.config, 'openapi.html', data)
+            content = render_j2_template(self.config, 'openapi.html', data,
+                                         request.locale)
             return headers, 200, content
 
         headers['Content-Type'] = 'application/vnd.oai.openapi+json;version=3.0'  # noqa
@@ -521,7 +519,7 @@ class API:
         headers = request.get_response_headers()
         if request.format == 'html':  # render
             content = render_j2_template(self.config, 'conformance.html',
-                                         conformance)
+                                         conformance, request.locale)
             return headers, 200, content
 
         return headers, 200, to_json(conformance, self.pretty_print)
@@ -844,10 +842,11 @@ class API:
             if dataset is not None:
                 content = render_j2_template(self.config,
                                              'collections/collection.html',
-                                             fcm)
+                                             fcm, request.locale)
             else:
                 content = render_j2_template(self.config,
-                                             'collections/index.html', fcm)
+                                             'collections/index.html', fcm,
+                                             request.locale)
 
             return headers, 200, content
 
@@ -938,7 +937,7 @@ class API:
                 self.config['resources'][dataset]['title'], request.locale)
             content = render_j2_template(self.config,
                                          'collections/queryables.html',
-                                         queryables)
+                                         queryables, request.locale)
 
             return headers, 200, content
 
@@ -1245,7 +1244,7 @@ class API:
 
             content = render_j2_template(self.config,
                                          'collections/items/index.html',
-                                         content)
+                                         content, request.locale)
             return headers, 200, content
         elif request.format == 'csv':  # render
             formatter = load_plugin('formatter',
@@ -1395,7 +1394,7 @@ class API:
 
             content = render_j2_template(self.config,
                                          'collections/items/item.html',
-                                         content)
+                                         content, request.locale)
             return headers, 200, content
 
         elif request.format == 'jsonld':
@@ -1587,7 +1586,7 @@ class API:
                 self.config['resources'][dataset]['title'], request.locale)
             content = render_j2_template(self.config,
                                          'collections/coverage/domainset.html',
-                                         data)
+                                         data, request.locale)
             return headers, 200, content
         else:
             return self.get_format_exception(request)
@@ -1636,7 +1635,7 @@ class API:
                 self.config['resources'][dataset]['title'], request.locale)
             content = render_j2_template(self.config,
                                          'collections/coverage/rangetype.html',
-                                         data)
+                                         data, request.locale)
             return headers, 200, content
         else:
             return self.get_format_exception(request)
@@ -1737,7 +1736,8 @@ class API:
             tiles['maxzoom'] = p.options['zoom']['max']
 
             content = render_j2_template(self.config,
-                                         'collections/tiles/index.html', tiles)
+                                         'collections/tiles/index.html', tiles,
+                                         request.locale)
 
             return headers, 200, content
 
@@ -1892,7 +1892,7 @@ class API:
 
             content = render_j2_template(self.config,
                                          'collections/tiles/metadata.html',
-                                         metadata)
+                                         metadata, request.locale)
 
             return headers, 200, content
 
@@ -1982,10 +1982,11 @@ class API:
             if process is not None:
                 response = render_j2_template(self.config,
                                               'processes/process.html',
-                                              response)
+                                              response, request.locale)
             else:
                 response = render_j2_template(self.config,
-                                              'processes/index.html', response)
+                                              'processes/index.html', response,
+                                              request.locale)
 
             return headers, 200, response
 
@@ -2087,7 +2088,8 @@ class API:
                 'jobs': serialized_jobs,
                 'now': datetime.now(timezone.utc).strftime(DATETIME_FORMAT)
             }
-            response = render_j2_template(self.config, j2_template, data)
+            response = render_j2_template(self.config, j2_template, data,
+                                          request.locale)
             return headers, 200, response
 
         return headers, 200, to_json(serialized_jobs, self.pretty_print)
@@ -2285,8 +2287,9 @@ class API:
                     'job': {'id': job_id},
                     'result': job_output
                 }
-                content = render_j2_template(
-                    self.config, 'processes/jobs/results/index.html', data)
+                content = render_j2_template(self.config,
+                                             'processes/jobs/results/index.html',  # noqa
+                                             data, request.locale)
 
         return headers, 200, content
 
@@ -2445,8 +2448,9 @@ class API:
             headers['Content-Language'] = p.locale
 
         if request.format == 'html':  # render
-            content = render_j2_template(
-                self.config, 'collections/edr/query.html', data)
+            content = render_j2_template(self.config,
+                                         'collections/edr/query.html', data,
+                                         request.locale)
         else:
             content = to_json(data, self.pretty_print)
 
@@ -2494,7 +2498,7 @@ class API:
 
         if request.format == 'html':  # render
             content = render_j2_template(self.config, 'stac/collection.html',
-                                         content)
+                                         content, request.locale)
             return headers, 200, content
 
         return headers, 200, to_json(content, self.pretty_print)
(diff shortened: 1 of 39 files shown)


GeoSander:
Not sure why, but Travis seems to fail on `test_api.py` that it's pulling from the `master` branch instead of the `multilingual2` branch?
I believe the tests should pass...

pygeoapi/api.py
```diff
@@ -104,28 +113,320 @@
...
+        achieved in the :meth:`__init__` method of the `APIRequest`.
+        However, `APIRequest` can still be initialized using :meth:`__init__`,
+        but then the :attr:`data` property value will always be empty.
+
+        :param request:             The web platform specific Request instance.
+        :param supported_locales:   List or set of supported Locale instances.
+        :returns:                   An `APIRequest` instance with data.
+        """
+        api_req = cls(request, supported_locales)
+        if hasattr(request, 'data'):
```
francbartoli, line 240: @GeoSander how much is this strategy flexible and would that fit with other frameworks like Django (see #630)?

francbartoli (commented):
A couple of comments/questions after a first quick review:

1.  Does this PR support multilingual as an optional feature? In case it doesn't then I would expect that an option is raised if the configuration is not appropriate and a test is added to avoid unexpected behaviour
2. The multilingual is supposed to have a default language based on the order defined in the configuration. Can we add tests to make sure that the default language is preserved in the responses?

pygeoapi/api.py
GeoSander: Good point @francbartoli. The honest answer is: I don't know. However, looking at Django's [Request](https://docs.djangoproject.com/en/3.1/ref/request-response/) object, I see that all methods and/or properties are there to have the `APIRequest` object extract all the stuff we need.

pygeoapi/api.py
```diff
@@ -616,81 +911,74 @@ def describe_collections(self, headers_, format_, dataset=None):
...
+                    for c in fcm.get('collections', [])
+                ]
+            return headers, 200, to_json(jsonld, self.pretty_print)
+
+        return headers, 200, to_json(fcm, self.pretty_print)
 
     @pre_process
     @jsonldify
-    def get_collection_queryables(self, headers_, format_, dataset=None):
+    def get_collection_queryables(self, request: Union[APIRequest, Any], dataset=None):  # noqa
```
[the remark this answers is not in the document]
GeoSander: The line was 4 characters too long. Frankly, I don't like multi-line function or method signatures, so if can avoid that in any way, I tend to do so.
Personally, I actually think the 79 char limit from PEP8 is ridiculous these days, but that's another discussion. Most of us have big hi-res screens. It's like the Bible... Maybe it once made sense, but nowadays, it feels rather arcane to me ;)

Anyway, I'd be happy to indent that line for you :)

GeoSander:
> A couple of comments/questions after a first quick review:

Thank you for taking the time to look at this. Much appreciated 🙏 

> 1. Does this PR support multilingual as an optional feature? In case it doesn't then I would expect that an option is raised if the configuration is not appropriate and a test is added to avoid unexpected behaviour

As it is designed now, the multilingual support is always on. Believe it or not, but I feel that this was the least intrusive way to implement it. The code simply looks for language structs anywhere (e.g. in the config) and "flattens" those structures into translated strings (or lists of strings if the input was a list of strings too) for the best-matching language or default language if no match was found. 

Turning it "off" would now be achieved by specifying only supported 1 language in the config. However, this does not change the behavior of the language-related code: it will still look for any language structs (note: it only considers the lowest level `dict`s where all its keys are valid language codes, i.e. if it encounters a `dict` that contains another `dict` in any of its values or where one of the keys is not a language code, then it is not considered to be a language struct and left as-is) and pick the best-matching language string. If the user did not put any language structs in the config and pygeoapi has been set up to only support 1 language, it is simply assumed that all strings are in that language.
Personally, I like the flexibility here. Like this, users/maintainers can gradually translate strings at their own pace instead of taking an "all-or-nothing" approach. Perhaps @pvgenuchten can elaborate?

But am I assuming correctly that you would rather have a built-in check somewhere that enforces the user to **not** use any language structs in the configuration if only 1 supported language has been specified? Note that it's easy to scan for non-allowed language structs, but doing it the other way around can be tricky: if a user configures multiple languages, we can only enforce the user to specify language structs at certain places if there is a strict schema that defines how it should look like and where it may be applied.

> 2. The multilingual is supposed to have a default language based on the order defined in the configuration. Can we add tests to make sure that the default language is preserved in the responses?

I thought I already did that, but if I forgot, then sure! I can do that 👍

pygeoapi/linked_data.py
```diff
@@ -126,10 +137,10 @@ def jsonldify_collection(cls, collection):
             cls.config['server']['url'],
             collection['id']
         ),
-        "name": collection['title'],
-        "description": collection['description'],
+        "name": l10n.translate(collection['title'], locale_),
+        "description": l10n.translate(collection['description'], locale_),
```
alpha-beta-soup, line 141: I think using [`@language`](https://json-ld.org/spec/ED/json-ld-syntax/20120522/#string-internationalization) construct is better practice; I don't think the JSON-LD document should be sensitive to the requested language which seems more suited for _human reading_.

GeoSander: Thanks for reviewing and bringing this up. I'm not too familiar with JSON-LD (yet) and I wasn't aware of its language support, but this definitely seems a good approach. I will look into it, thanks! 👍

GeoSander:
Just to keep things organized: some previous discussion with @tomkralidis [here on Gitter](https://gitter.im/geopython/pygeoapi?at=6062827beb75085d47f7a569).

pvgenuchten:
The conversation here seems stalled by two items:

- pygeoapi can either have a strict schema for its config or a loose, in case of strict a value which potentially has multiple translations always should be a cluster lang:value pairs, with at least 1 pair. In case of loose, a user can place a string directly (single language) or a cluster of lang:value pairs. For us option 2 fits better with the project, but we're fine to implement option 1 if requested.
- pygeoapi can implement some json-ld conventions, including language support. I personally like the json-ld approach, but it requires some R&D/development first. Also both approaches can probably exist besides each other.

> The multilingual is supposed to have a default language based on the order defined in the configuration. Can we add tests to make sure that the default language is preserved in the responses?

@francbartoli can you explain what is the risk you want to prevent with such a test? I guess returning the default language is a common case (italian browser asks a service which has english and french configured, of which english is default; it should return english)

francbartoli:
> The conversation here seems stalled by two items:
> 
> * pygeoapi can either have a strict schema for its config or a loose, in case of strict a value which potentially has multiple translations always should be a cluster lang:value pairs, with at least 1 pair. In case of loose, a user can place a string directly (single language) or a cluster of lang:value pairs. For us option 2 fits better with the project, but we're fine to implement option 1 if requested.
> * pygeoapi can implement some json-ld conventions, including language support. I personally like the json-ld approach, but it requires some R&D/development first. Also both approaches can probably exist besides each other.
> 
> > The multilingual is supposed to have a default language based on the order defined in the configuration. Can we add tests to make sure that the default language is preserved in the responses?
> 
> @francbartoli can you explain what is the risk you want to prevent with such a test? I guess returning the default language is a common case (italian browser asks a service which has english and french configured, of which english is default; it should return english)

Just preventing the configuration has robust checks for the languages. As an example, would this work:

languages:
        # First language is the default language
        - en-US
        - fr-CA


> The conversation here seems stalled by two items:
> 
> * pygeoapi can either have a strict schema for its config or a loose, in case of strict a value which potentially has multiple translations always should be a cluster lang:value pairs, with at least 1 pair. In case of loose, a user can place a string directly (single language) or a cluster of lang:value pairs. For us option 2 fits better with the project, but we're fine to implement option 1 if requested.
> * pygeoapi can implement some json-ld conventions, including language support. I personally like the json-ld approach, but it requires some R&D/development first. Also both approaches can probably exist besides each other.
> 
> > The multilingual is supposed to have a default language based on the order defined in the configuration. Can we add tests to make sure that the default language is preserved in the responses?
> 
> @francbartoli can you explain what is the risk you want to prevent with such a test? I guess returning the default language is a common case (italian browser asks a service which has english and french configured, of which english is default; it should return english)

@pvgenuchten my primary concern is not actually the browser as at the end we are always talking about an API so not necessarily it has to be consumed by a real user. Now, if the default language is set implicitly by the configuration order to `fr` for example, I do believe we have to test that effectively, in the various conditions a request can happen, then the expected default language is returned when the request doesn't have the language parameter or the header.
Also, is there a way for a machine to deduct what is the default language from the responses? Does this PR allow to do that implicitly or explicitly? Or it doesn't at all?

GeoSander:
Hi @tomkralidis, @francbartoli et al., I have updated the PR, making sure that:
- Responses always have a `Content-Language` header (consistency);
- The `Content-Language` header is always set to the default language if no specific language is requested;
- The default language always is the first language defined in the pygeoapi server configuration YAML `languages` property (if multiple languages have been configured).

Note that there is still quite some work to do. The bulk of this work focuses on making the hard-coded English content strings in `api.py` easily translatable and adding HTML templates for other languages as well. But this PR lays the groundwork for this.

pygeoapi/api.py
```diff
@@ -2251,15 +2313,15 @@ def execute_process(self, request: Union[APIRequest, Any], process_id):
...
-                    data_dict[id] = value
-                elif id in data_dict and isinstance(data_dict[id], list):
-                    data_dict[id].append(value)
+            for input_ in data.get('inputs', []):
+                id_ = input_['id']
+                value = input_['value']
+                if id_ not in data_dict:
+                    data_dict[id_] = value
+                elif id_ in data_dict and isinstance(data_dict[id_], list):
+                    data_dict[id_].append(value)
```
GeoSander, line 2322: Note that `input` and `id` are built-in Python types!

tomkralidis (approved):

tomkralidis:
Thank you @GeoSander @pvgenuchten for this impressive and valuable contribution!

PR #676, opened 2021-04-19
Support for PIDs and (non-Geo)JSON-LD

ksonda:
This PR attempts to update #465/#469 to `master`, in order to support external/global identifiers in the JSON-LD and HTML representations of individual features and to support the JSON-LD use cases of @ksonda and @dblodgett-usgs, which involve standard JSON-LD and not GeoJSON-LD. This PR changes or adds a number of intertwined behaviors.

1. To partially address #384 and #457, an optional boolean `geojsonld` option is now obtained from the configuration which changes the JSON-LD representation. This option does not change HTML or JSON representations.  This option allows the user to choose between providing JSON-LD data about a real-world feature that is located at a certain geometry (`geojsonld: False`), versus providing JSON-LD data about a particular GeoJSON feature which has attributes that may or may not have semantic specifications (`geojsonld: True`; i.e. the default/current behavior). When `geosjonld:True` or unspecified, the current behavior is preserved as default. When `geojsonld:False`...
 - If the geometry is non-point, the geometry is omitted from the JSON-LD, since non-point geometry is not proper JSON-LD
 - If the geometry is point, the geojson vocabulary in the default context is replaced with schema.org specifications sufficient to represent the feature as a https://schema.org/Place associated with geometry of type https://schema.org/GeoCoordinates.
 -  The default context is replaced for the /items endpoint to represent the feature collection as a https://schema.org/ItemList  instead of a geojson FeatureCollection
 - The `properties` block is removed in favor of representing all feature attributes on the same level as `@id` and `@type`. The default behavior results in all feature attributes being represented in a blank node. This is not the desired behavior for linked data provision, if feature attributes are meant to be RDF predicates with `@id` as their subject.
 

2. To address #433, optional `uri_field` is obtained from configuration to support custom URIs for features, which determine JSON-LD `@id` at the level of individual features, and canonical URL in item html templates. 
3. To address #453, when `uri_field` is specified a `uri` property is created in item tables that link to the pygeoapi item itself. 

Documentation on how to use these two options is added to `configuration.rst`

The way that `geojsonld` and `uri_field` interact can be shown with the demo deployment available on Dockerhub: webbben/pygeoapi-uri:

`docker run -p 5000:80 --rm webbben/pygeoapi-uri`

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -176,7 +176,7 @@ default.
                 name: CSV
                 data: tests/data/obs.csv  # required: the data filesystem path or URL, depending on plugin setup
                 id_field: id  # required for vector data, the field corresponding to the ID
-                uri_field: uri # opetional field corresponding to the Uniform Resource Identifier (see Linked Data section)
+                uri_field: uri # optional field corresponding to the Uniform Resource Identifier (see Linked Data section)
                 time_field: datetimestamp  # optional field corresponding to the temporal property of the dataset
                 title_field: foo # optional field of which property to display as title/label on HTML pages
                 format:  # optional default format
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -242,7 +242,7 @@ For collections, at the level of an item or items, the default the JSON-LD repre
 
 The optional configuration options for collections, at the level of an item of items, are:
 
-- If ``geojsonld`` is not specified or is True, the JSON-LD will conform to the default configuration. If geojsonld is False, the properties block will be expanded   into the main body, non-point geometry will be removed, and a custom ``@context`` will be used in the JSON-LD.
+- If ``geojsonld`` is not specified or is True, the JSON-LD will conform to the default configuration. If ``geojsonld`` is False, the properties block will be expanded   into the main body, non-point geometry will be removed, and a custom ``@context`` will be used in the JSON-LD.
 - If ``uri_field`` is specied, JSON-LD will be updated such that ``@id:uri``.
 
 .. note::
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -147,7 +147,7 @@ default.
           keywords:  # list of related keywords
               - observations
               - monitoring
-          geojsonld: True # use default geojsonld behavior (see Linked Data section)
+          geojsonld: true # use default geojsonld behavior (see Linked Data section)
           context:  # linked data configuration (see Linked Data section)
               - datetime: https://schema.org/DateTime
               - vocab: https://example.com/vocab#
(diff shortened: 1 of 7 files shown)


pygeoapi/linked_data.py
```diff
@@ -171,35 +171,89 @@ def geojson2geojsonld(config, data, dataset, identifier=None):
...
+                            config['server']['url'], dataset, feature['id'])
+
+            if not geojsonld:
+                feature, geocontext = make_jsonld(feature)
+                # Note: @id or https://schema.org/url or both or something else?
+                feature[identifier_field] = identifier
+            else:
+                feature['id'] = identifier
+           
+            data['features'][i] = feature
```
alpha-beta-soup, line 205: Not sure if this is necessary, or if `feature` is already being mutated in the enumeration?

pygeoapi/linked_data.py
```diff
@@ -171,35 +171,89 @@ def geojson2geojsonld(config, data, dataset, identifier=None):
...
+    else:
+        ldjsonData = {
+            "@context": [
+                defaultVocabulary,
+                *(context or [])
+            ],
+            **data
+        }
+
+    return ldjsonData
```
alpha-beta-soup, line 233: `json.dumps(ldjsonData)`, i.e. this function no longer returns a string - is that correct?

alpha-beta-soup:
This is nice, I thought the division in treatment would make this too complicated but it seems OK to me.

ksonda:
Thanks @alpha-beta-soup .  @webb-ben , if you have any clarifications regarding the review comments feel free to respond here.

pygeoapi/linked_data.py
webb-ben: I think it happens on line 199 when the `feature` is re-declared from the custom json-ld context. This felt like the simplest way to ensure the feature retained was the updated one.

pygeoapi/linked_data.py
webb-ben: I moved that into api.py so the json-ld could refer to `self.pretty_print` (line 1209 of api.py) early on for readability. I can see how this might not fit the specifications of `geojson2geojsonld(` though.

docs/source/configuration.rst
```diff
@@ -147,6 +147,7 @@ default.
           keywords:  # list of related keywords
               - observations
               - monitoring
+          geojsonld: True # use default geojsonld behavior (see Linked Data section)
```
tomkralidis, line 150: Should be (JSON) `true`

tomkralidis (changes requested):

docs/source/configuration.rst
ksonda: Thanks to @webb-ben for changing this. All checks pass now.

tomkralidis:
@alpha-beta-soup looks good to you?  I can squash / merge with your +1.

tomkralidis (approved):

PR #680, opened 2021-04-24
fixes JSON and JSON-LD links in HTML base template

tomkralidis:
This PR fixes the HTML base template's JSON and JSON-LD links.  In `pygeoapi/templates/_base.html` :

```
<span style="float:right">
    <a href="{{ data['links'] | map(rel='self') | attr('href') }}?f=json">JSON</a>
    <a href="{{ data['links'] | map(rel='self') | attr('href') }}?f=jsonld">JSON-LD</a>
</span>
```

rendered the following at all times:

```html
<a href="?f=json">JSON</a>
<a href="?f=jsonld">JSON-LD</a>
```

Which indicates that the rendered links were always empty.  This PR fixes the issue, so that links with extra query parameters (i.e. items queries, coverage queries, etc.), have the equivalent JSON link for consistency.

### DIFF
--- a/pygeoapi/templates/_base.html
+++ b/pygeoapi/templates/_base.html
@@ -40,8 +40,25 @@
         <a href="{{ config['server']['url'] }}">Home</a>
         {% endblock %}
         <span style="float:right">
-          <a href="{{ data['links'] | map(rel='self') | attr('href') }}?f=json">JSON</a>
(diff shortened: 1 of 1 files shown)


dblodgett-usgs:
@tomkralidis, @pvgenuchten and @alpha-beta-soup -- can we make sure that #676 gets some attention before too long? I want to make sure that PR doesn't fall behind and it looks like this might be touching some of the same files.

pvgenuchten (approved):
Looks good tom, thanx.

PR #689, opened 2021-05-18
add geometry representations for json-ld output format

ksonda:
This is a draft/ work-in-progress w/ @webb-ben  to give linked-data views of `/items/{item}?f=jsonld` - level feature geometry that more fully meet the use cases of users that would request a json-ld document. 

It adds onto #676, which allows users to specify a `geojsonld:` option in the configuration. 

In this branch, specifying `geojsonld: false` results in the following json-ld document format:

1. Rather than GeoJSON-LD, the JSON-LD document is standard JSON-LD, with the GeoJSON `properties` block removed in favor of top-level attributes. (This remains from #676)
2. Three versions of geometry are presented
  i. A geojson-style `geometry:` block
 ii. A wkt literal enclosed in a <http://www.opengis.net/ont/geosparql#hasGeometry> block
iii. A schema:GeoCoordinates (for Point-type geometry) or a schema:GeoShape (for all other types of geometry). In the case of multipoint, multipolygon and multilinestring features, the feature geometry is simplified into a single polygon via a convex hull. 
3. The `/items?f=jsonld` endpoint removes all geometry, in favor or presenting a schema:itemList of URIs pointing to the appropriate /items/{item id} features.

We believe this better represents the needs of users using pygeoapi to publish json-ld features than GeoJSON-LD, which is not a standard. A geosparql/wkt representation allows ingestion of geometry information into spatially-aware graph databases, and the schema:geo representation allows indexing by Google etc. The geojson-style geometry block can be parsed in a workflow as a simple feature without a separate API call to the /item?f=json endpoint.

https://info.geoconnex.us is running this branch. See https://geoconnex.us/ref/states/08?f=jsonld for an example /item-level output. See https://info.geoconnex.us/collections/states/items?f=jsonld for an example /items level output.

An alternative approach would be to explore content negotiation by profile, and enabling a kind of additional ?view= query parameter that can be paired alongside ?f= to allow machines to specify both the media type and the data profile (i.e. GeoJSON-LD, geosparql predicates, schema:geo predicates) desired.

Welcome your thoughts, @dblodgett-usgs, @alpha-beta-soup , @pvgenuchten


alpha-beta-soup:
I support this idea in principle. I introduced GeoJSON-LD only as a stand-in until something more robust could be proposed.

ksonda:
If no one is particularly wedded to the geojson-ld format, in the short term we could get rid of the `geojsonld: ` config option and have this be the new default (including the itemList w/o geometry in the /items?f=jsonld endpoint.)

Alternatively, we can explore adding more configuration options to specify exactly which geometry versions people want in the json-ld.

I think adding a profile ?view would require more systematic changes throughout pygeoapi but perhaps this could be put on a longer-term roadmap.

tomkralidis (commented):
LGTM.  I can merge once @alpha-beta-soup reviews/approves.

docs/source/configuration.rst
```diff
@@ -234,15 +233,20 @@ The metadata for an instance is determined by the content of the `metadata`_ sec
 This metadata is included automatically, and is sufficient for inclusion in major indices of datasets, including the
 `Google Dataset Search`_.
 
-For collections, at the level of an item or items, the default the JSON-LD representation adds:
+For collections, at the level of item, the default the JSON-LD representation adds:
```
alpha-beta-soup, line 236: Typo `the default the`

docs/source/configuration.rst
```diff
@@ -234,15 +233,20 @@ The metadata for an instance is determined by the content of the `metadata`_ sec
...
-- The GeoJSON JSON-LD `vocabulary and context <https://geojson.org/geojson-ld/>`_ to the ``@context``.
-- An ``@id`` for each item in a collection, that is the URL for that item (resolving to its HTML representation
-  in pygeoapi)
+- An ``@id`` for the item, which is the URL for that item. If uri_field is specified,
+  it is used, otherwise the URL is to its HTML representation in pygeoapi.
+- separate GeoSPARQL/WKT and `schema.org/geo` versions of the geometry.
+- ``@context`` for the GeoSPARQL and schema geometries.
+- The unpacked properties block into the main body of the item.
+
+For collections, at the level of items, the default the JSON-LD representation adds:
```
alpha-beta-soup, line 244: same typo

docs/source/configuration.rst
```diff
@@ -234,15 +233,20 @@ The metadata for an instance is determined by the content of the `metadata`_ sec
...
+
+- A schema.org itemList of the ``@id`` and ``@type`` of each feature in the collection.
 
 The optional configuration options for collections, at the level of an item of items, are:
 
-- If ``geojsonld`` is not specified or is ``true``, the JSON-LD will conform to the default configuration presenting GeoJSON-LD. If ``geojsonld`` is ``false`` the individual item JSON-LD will conform to standard JSON-LD, and the ``@context`` will not include the geojson vocabulary, but only one specifying the ``@id`` property, the ``@type`` property, and schema.org/ vocabulary. In addition, properties block will be expanded into the main body, non-point geometry will be removed, and point-type geometry will be represented as https://schema.org geometry instead of geojson.
 - If ``uri_field`` is specified, JSON-LD will be updated such that ``@id:uri_field`` for each item in a collection.
```
alpha-beta-soup, line 250: This line isn't clear to me. Perhaps "...such that the ``@id`` has the value of ``uri_field`` for each item in a collection."?

pygeoapi/linked_data.py
```diff
@@ -173,104 +175,146 @@ def jsonldify_collection(cls, collection, locale_):
...
-    :param config: dict of configuration
-    :param data: dict of data:
-    :param dataset: dataset identifier
-    :param identifier: item identifier (optional)
-    :param id_field: item identifier_field (optional)
+        :param config: dict of configuration
+        :param data: dict of data:
+        :param dataset: dataset identifier
+        :param identifier: item identifier (optional)
+        :param id_field: item identifier_field (optional)
```
alpha-beta-soup, line 186: Remind me: what's the difference between `identifier` and `id_field`?

pygeoapi/linked_data.py
```diff
@@ -173,104 +175,146 @@ def jsonldify_collection(cls, collection, locale_):
 
 def geojson2geojsonld(config, data, dataset, identifier=None, id_field='id'):
```
alpha-beta-soup, line 176: Time to rename the function?

pygeoapi/linked_data.py
```diff
@@ -173,104 +175,146 @@ def jsonldify_collection(cls, collection, locale_):
...
+
+        # Include multiple geometry encodings
+        data['type'] = 'schema:Place'
+        jsonldify_geometry(data)
+        data[id_field] = identifier
 
     else:
         # Collection of geojsonld
+        defaultVocabulary.update({
+                'features': 'schema:itemListElement',
```
alpha-beta-soup, line 214: Odd indenting - linting?

pygeoapi/linked_data.py
```diff
@@ -173,104 +175,146 @@ def jsonldify_collection(cls, collection, locale_):
...
-                    "features": "schema:itemListElement",
-                    "FeatureCollection": "schema:itemList"
-                })
-                # Note: @id or https://schema.org/url, both or something else?
-                feature[id_field] = identifier
-            else:
-                feature['id'] = identifier
-
-            data['features'][i] = feature
+            # Note: @id or https://schema.org/url, both or something else?
```
alpha-beta-soup, line 230: Has this note been resolved now? (If so, remove.)

pygeoapi/linked_data.py
```diff
@@ -173,104 +175,146 @@ def jsonldify_collection(cls, collection, locale_):
...
+    if geom.geom_type == 'Point':
+        feature['schema:geo'] = {
+            '@type': 'schema:GeoCoordinates',
+            'schema:longitude': geom.x,
+            'schema:latitude': geom.y
+        }
+        return
+
+    elif geom.geom_type == 'LineString':
+        _ = [f'{x},{y}' for (x, y) in geom.coords[:]]
```
alpha-beta-soup, line 285: I wonder if these operations can be split out into smaller dedicated functions, that can then also be tested in isolation.

pygeoapi/linked_data.py
```diff
@@ -173,104 +175,146 @@ def jsonldify_collection(cls, collection, locale_):
...
+
+    elif geom.geom_type == 'MultiPoint':
+        poly_geom = [(p.x, p.y) for p in geom.geoms]
+        poly_geom.append(poly_geom[0])
+
+    elif geom.geom_type == 'Polygon':
+        poly_geom = geom.exterior.coords[:]
+
+    elif geom.geom_type == 'MultiPolygon':
+        poly = unary_union(geom.buffer(0))
```
alpha-beta-soup, line 304: Perhaps a comment to explain why a geometry is being buffered by zero?

pygeoapi/linked_data.py
```diff
@@ -173,104 +175,146 @@ def jsonldify_collection(cls, collection, locale_):
...
+        if poly.geom_type.startswith('Multi') or not poly.is_valid:
```
alpha-beta-soup, line 305: And perhaps something in documentation about the level of support for `Multi*` type geometries?

pygeoapi/linked_data.py
```diff
@@ -173,104 +175,146 @@ def jsonldify_collection(cls, collection, locale_):
...
-                        "lat": "schema:latitude",
-                        "long": "schema:longitude"
-        }, ]
-
-    return feature, geocontext
+        poly_geom = list()
+        for p in geom.geoms:
+            try:
+                poly_geom.extend(p.coords[:])
+            except NotImplementedError:
```
alpha-beta-soup, line 316: Not clear to me what's happening here? Something to do with holes?

tests/test_api.py
```diff
@@ -772,22 +756,30 @@ def test_get_collection_item_json_ld(config, api_):
...
+            '@value'] == '35'
+    assert expanded['http://www.opengis.net/ont/geosparql#hasGeometry'][0][
+            'http://www.opengis.net/ont/geosparql#asWKT'][0][
+            '@value'] == 'POINT (-75 45)'
+    assert expanded['https://schema.org/geo'][0][
+            'https://schema.org/latitude'][0][
+            '@value'] == 45.0
+    assert expanded['https://schema.org/geo'][0][
+            'https://schema.org/longitude'][0][
+            '@value'] == -75.0
```
alpha-beta-soup, line 782: Is it possible to do tests for non-point geometries, too?

pygeoapi/linked_data.py
webb-ben: `def geojson2jsonld(... `?

pygeoapi/linked_data.py
webb-ben: I think there is a lot of space here for improvement of this section. I don't even know the name of this coordinate system. It resembles the geometry section of GML but there doesn't appear to be an easy way to transform to these shapes. I left it all in the same function because I was trying to keep `geom` isolated to within jsonldify geometry. The ideal situation would probably be an addition to shapely geometry outputs written in C/cython. Interested to hear what you think is the best short and long term implementations would look like.

pygeoapi/linked_data.py
webb-ben: This is to catch `GeometryCollection` or anything else. The reduction from GeoJSON & WKT to the schema.org geometry is pretty large, and I couldn't think of a better way to reduce something like a `GeometryCollection`

pygeoapi/linked_data.py
ksonda: lets change to `geojson2jsonld`

pygeoapi/linked_data.py
ksonda: we're going with `@id`

docs/source/configuration.rst
webb-ben: amended

docs/source/configuration.rst
webb-ben: amended

docs/source/configuration.rst
webb-ben: amended

pygeoapi/linked_data.py
```diff
@@ -173,104 +175,146 @@ def jsonldify_collection(cls, collection, locale_):
...
-    geojsonld = config['resources'][dataset].get('geojsonld', True)
+    context = config['resources'][dataset].get('context', []).copy()
+    defaultVocabulary = {
+        'schema': 'https://schema.org/',
+        id_field: '@id',
+        'type': '@type'
+    }
 
     if identifier:
         # Single geojsonld
```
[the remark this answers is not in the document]
webb-ben: amended

pygeoapi/linked_data.py
webb-ben: updated this in api.py and linked_data.py

pygeoapi/linked_data.py
webb-ben: don't know how flake8 missed this. fixed now

pygeoapi/linked_data.py
webb-ben: amended

pygeoapi/linked_data.py
webb-ben: Added comment

pygeoapi/linked_data.py
webb-ben: Added documentation for behavior of schema in the linked data section.

tests/test_api.py
webb-ben: added tests for LineString, Polygon, MultiPoint, MultiLineString, and MultiPolygon

pygeoapi/linked_data.py
ksonda: For now we've isolated the geometry to schema.org conversion function, but not gone so far as to make a function for each of the geom types for the medium term. Schema.org geometry is so nonstandard and hard to parse I doubt anything useful could get into main shapely branch anytime soon though.

ksonda:
@tomkralidis , @alpha-beta-soup I believe we have addressed everything, except for fully exploding the schema/geo into separate geometry type conversion functions.

alpha-beta-soup:
LGTM

alpha-beta-soup (approved):

tomkralidis:
Great work!

Issue #567, opened 2020-11-06
Default map provider (maps.wikimedia.org) tile requests are returning HTTP 403 response codes.


Dukestep:
**Description**
The default map provider (https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png) is returning a 403 forbidden status code when displaying items via the HTML UI.

**Steps to Reproduce**
https://demo.pygeoapi.io/master/collections/lakes/items

Behaviour is also reproduced on a completely different private instance.

**Expected behavior**
Tiles requests should be returning 200 OK code.

**Environment**
 - OS: Not applicable
 - Python version: Not applicable
 - pygeoapi version: master

According to their [Terms of Use](https://foundation.wikimedia.org/wiki/Maps_Terms_of_Use):
```
Using maps in third-party services

If you use the Wikimedia Maps service for your third-party project (that is, outside of the Wikimedia projects), 
please respect our limited services and resources. We reserve the right to discontinue or change our service, 
block or limit certain users or applications, or take other measures in cases at our sole discretion 
at any time without notice. If you use the service too heavily, it could affect the service's stability for 
others or degrade our quality of service.

If you are developing an application that uses the Wikimedia Maps service, you must provide a valid HTTP 
User-Agent that includes your application, version, and sufficient information to easily contact you 
(e.g., your email address).

The following are prohibited on the Wikimedia Maps service:

    - Excessive downloading (such as downloading significant areas of tiles for later offline usage)
    - Accessing the service without a proper HTTP User-Agent or HTTP referer
    - Using the service without compliance with any license or copyright terms
```

It seems like we access the service with a valid HTTP referer but may fall in the "use the service too heavily" category?

jasonwhitedev:
Gives out this error for the missing tiles/pngs: 
`Error: 403, Forbidden: Map tiles are restricted to Wikimedia & affiliated sites only. Please post on https://phabricator.wikimedia.org/T261694 if you believe your usage supports the Movement. at Tue, 10 Nov 2020 11:48:34 GMT`

As mentioned, someone probably needs to go & do this: 
```
If you are developing an application that uses the Wikimedia Maps service, you must provide a valid HTTP 
User-Agent that includes your application, version, and sufficient information to easily contact you 
(e.g., your email address).
```

justb4:
[...]

PR #697, opened 2021-05-27
Changed default background map to OSM

doublebyte1:
Default map provider (maps.wikimedia.org) tile requests are returning HTTP 403 response codes.

Reference to this issue:
https://github.com/geopython/pygeoapi/issues/567

### DIFF
--- a/pygeoapi-config.yml
+++ b/pygeoapi-config.yml
@@ -42,7 +42,7 @@ server:
       # path: /path/to/Jinja2/templates
       # static: /path/to/static/folder # css/js/img
     map:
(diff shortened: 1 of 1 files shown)


justb4:
Some history here:

* originally `pygeoapi` was using tiles.openstreetmap.org
* switched to maps.wikimedia.org mainly as using tiles.openstreetmap.org was "not done"
* at the time maps.wikimedia.org had no usage restrictions
* that changed, like it seems
* back to tiles.openstreetmap.org as in this PR works, but is this still in line with current OSM ToU? 

This is a recurring issue in many projects, maybe time for a neutral party, like OSGeo (?), to host OSM tiles without any restrictions?

Issue #616, opened 2021-02-04
Migrate to GitHub Actions


tomkralidis:
Given recent changes in Travis we should make the move to GitHub Actions.

jodygarnett:
Some notes from migrating core-geonetwork project from travis to github actions:

* it was pretty easy to do (it took hours not days)
* So much faster than Travis
* Have multiple small actions rather than a long script (like we did with Travis previously). This let’s you get feedback faster to folks making a pull-request
* Copy a working example from [geonetwork](https://github.com/geonetwork/core-geonetwork/tree/master/.github/workflows) or [geoserver](https://github.com/geoserver/geoserver/tree/master/.github/workflows)

plant99:
Started working on this, @tomkralidis please assign this to me if someone hasn't already taken this up.

tomkralidis:
@plant99 thanks for taking this on. Any updates?

plant99:
Hi @tomkralidis , [this branch](https://github.com/plant99/pygeoapi/commits/master) has some commits, haven't been able to find some time lately to actively take it to completion, please feel free to unassign me from the task, it'd be better if someone else can finish this faster.

tomkralidis:
Progress in https://github.com/geopython/pygeoapi/pull/703#issuecomment-851766559 for review/comment.

tomkralidis:
Implemented first pass in #703.  As discussed, we need to add the rest of the test functionality per https://github.com/geopython/pygeoapi/pull/703#issuecomment-851766559

tomkralidis:
(copying/updating comment https://github.com/geopython/pygeoapi/pull/703#issuecomment-851766559 here as part of the issue lineage)

@kalxas @justb4 @francbartoli 

FYI GitHub Actions setup passing so far.  Here's what is outstanding/not ported from the Travis setup in this PR:

- loading PostgreSQL/PostGIS data
- properly getting the GDAL Python bindings properly installed
- properly getting the pyspatialite dependencies installed
- running any tests that depend on the above
- anything Docker
- building Debian package

I've listed out each test in https://github.com/geopython/pygeoapi/blob/master/.github/workflows/main.yml so as to add/remove accordingly as part of testing.  Lines that are commented out need updates/fixes to get CI working properly on those tests.  Once all tests pass, then this step will simply be run as `pytest`.

PR #703, opened 2021-06-01
[WIP] migrate to GitHub Actions (#616)

tomkralidis:
Per #616


tomkralidis:
@kalxas @justb4 @jodygarnett @francbartoli 

FYI GitHub Actions setup passing so far.  Here's what is outstanding/not ported from the Travis setup in this PR:

- loading PostgreSQL/PostGIS data
- properly getting the GDAL Python bindings properly installed
- properly getting the pyspatialite dependencies installed
- running any tests that depend on the above
- anything Docker
- building Debian package

I've listed out each test in https://github.com/tomkralidis/pygeoapi/blob/issue-616/.github/workflows/main.yml so as to add/remove accordingly as part of testing.  Once all tests pass, then this step will simply be run at `pytest`.

Any help is appreciated to help get our tests working again.  Having said this, we may want to consider merging even with the abovementioned gaps given Travis testing is blocked by limits atm.

kalxas:
This looks great. Having this working in GH Actions is better that a more complete Travis list of test that will shut down soon.

kalxas (approved):

francbartoli (approved):
I’m inclined to merge as it is since we are not able to have any CI now. Then we can add the rest of the pieces missing from the travis pipeline

README.md
```diff
@@ -1,6 +1,6 @@
 # pygeoapi
 
-[![Build Status](https://travis-ci.org/geopython/pygeoapi.png)](https://travis-ci.org/geopython/pygeoapi)
+[![Build Status](https://github.com/geopython/pygeoapi/workflows/build%20%E2%9A%99%EF%B8%8F/badge.svg)](https://github.com/geopython/pygeoapi/actions)
```
justb4, line 3: think the badge string should be
 
[Build Status](https://github.com/geopython/pygeoapi/actions/workflows/main.yml)

justb4 (approved):
I think good to go as first step. Then you have something working.
Next we can fix GDAL (and uncomment from requirements-provider.txt) and other deps. Beste to look into the pygeoapi Dockerfile, where e.g. the deb packages `python3-gdal python3-psycopg2 python3-xarray` are installed.

Then we can also add the Docker build, test and push. As @jodygarnett indicated can be separate file. And maybe more selective (maybe only on push trigger).
I have some examples from other projects like:
https://github.com/justb4/docker-awstats/blob/master/.github/workflows/main.yml 
Also much faster than building on DockerHub.

PR #709, opened 2021-06-09
Add support for deploying pygeoapi as a lambda container

christophersbarrett:
The base pygeoapi implementation can deployed as a lambda function, but once you start adding geo processing libraries it quickly exceeds the 250MB limit that AWS puts on deployed lambda functions. To still use pygeoapi as a lambda, you need to deploy it as a lambda container.


christophersbarrett:
Unsure how to deal with that failing build ^^ The file it is complaining about is moved to the top level per the readme.

aws-lambda/container/.serverless-wsgi
```diff
@@ -0,0 +1 @@
+{"app":"pygeoapi/flask_app.APP","text_mime_types":["application/ld+json"]}
```
tomkralidis, line 1: Do we need to add all the default supported text MIME types here?

- `application/json`
- `text/html`
- `application/geo+json`
- `application/prs.coverage+json`
- `application/xml`

aws-lambda/container/wsgi.py
```diff
@@ -0,0 +1,9 @@
+import os
```
tomkralidis, line 1: The CI tests run flake8 and this file throws flake8 error. Does the following work?
```python
import os

from pygeoapi.flask_app import APP

os.environ['PYGEOAPI_CONFIG'] = 'pygeoapi-test-config.yml'
os.environ['PYGEOAPI_OPENAPI'] = 'pygeoapi-test-openapi.yml'

if __name__ == "__main__":
    APP.run()
```

tomkralidis (changes requested):
cc'ing @bitner @jvanulden and @bo-lu here for possible feedback as well.

PR #711, opened 2021-06-10
Docker locale tz

tomkralidis:
Update of #470 given #664 and recommendation from @alpha-beta-soup (cc @GeoSander)


alpha-beta-soup (approved):

francbartoli (approved):

justb4 (changes requested):
I agree with having default UTC as default tz, but IMO setting of TZ is incorrect: line 68 should assign `TZ=${TIMEZONE}` not `TZ=${TZ}`.

tomkralidis:
Thanks @justb4, fixed.

justb4 (approved):

PR #716, opened 2021-06-18
Fix SQLite provider for GPKG with multiple layers

### LINKED ISSUE #519: SQLiteGPKG plugin doesn't support gpkg with multiple tables
**Description**
Method __load in pygeoapi/provider/sqlite.py raises an unnecessary InvalidPluginError ([line 221](https://github.com/geopython/pygeoapi/blob/85939109b1946f2bc951f7e2ebce603561318ee6/pygeoapi/provider/sqlite.py#L221)) if query `SELECT AutoGPKGStart();` returns >1. AutoGPKGStart returns an integer of how many VirtualGPKG tables have been created. According to the log message this exception should only be raised if the value is 0. In my case the gpkg has 2 tables, AutoGPKGStart returns 2, and pygeoapi raises an exception and fails to start.

**Steps to Reproduce**
- Create a .gpkg file with two or more spatial tables
- Try to publish a table from the geopackage as a resource with pygeoapi using the SQLiteGPKG provider

**Expected behavior**
The exception should only be raised if `SELECT AutoGPKGStart();` returns 0.

**Environment**
- Docker image: geopython/pygeoapi:latest

mvaaltola:
Running SELECT AutoGPKGStart(); returns the number of layers in a GPKG
file.
Therefore, only 0 is an invalid value which should raise an error.

An example GPKG can be provided for testing purposes, if needed.

Resolves #519.

### DIFF
--- a/pygeoapi/provider/sqlite.py
+++ b/pygeoapi/provider/sqlite.py
@@ -212,11 +212,11 @@ class SQLiteGPKGProvider(BaseProvider):
         if self.application_id:
             cursor.execute("SELECT AutoGPKGStart()")
             result = cursor.fetchall()
-            if result[0][0] == 1:
+            if result[0][0] >= 1:
                 LOGGER.info("Loaded Geopackage support")
             else:
                 LOGGER.info("SELECT AutoGPKGStart() returned 0." +
-                            "Detected GPKG but couldnt load support")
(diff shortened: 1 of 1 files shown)


tomkralidis:
Thanks very much @mvaaltola !

Issue #553, opened 2020-10-11
more/better validation of yaml content 


jorgejesus:
**Is your feature request related to a problem? Please describe.**
Currrent `yaml` config file is big and with data, that requires validity checks in multiple points. This makes the code confusing and time consuming to implement fixes 

**Describe the solution you'd like**
- Check possibility of implementing [Yamale](https://github.com/23andMe/Yamale), this tool could even create the validation schema from a pre-existing yml (bootstrap).
-  Strict validation possibility inside code or as an extra support tool

**Describe alternatives you've considered**
None

**Additional context**
This came out looking at #316  that requires datetime validation (not only the simple  `datetime` object when parsing the yml). The yml parser on `utils` could copy/paste some code from [piny](https://github.com/pilosus/piny)

jorgejesus:
After discussion, better to implement `jsonschema` for validation.

AelMalinka:
We are currently working on a system that generates a pygeoapi config (along with other configs and does various data manipulations) and ran into an issue where it generated an array for `resources -> providers -> geom_fields`; obviously, that's wrong but it would be very beneficial to have some mechanism by which to validate the generated config

tomkralidis:
FYI implemented in master via #719

PR #719, opened 2021-06-29
add support for configuration schema and validation (#553)

tomkralidis:
This implements #553 by adding a JSON schema and validation for pygeoapi configuration.  Notes:

- JSON is actually managed in YAML for brevity, and is transformed during runtime validation
- schema is in `pygeoapi/schemas/config/pygeoapi-config-0.x.yml`
  -  areas that could potentially be improved:
     - JSON-LD `context` object is loosely specified for now (@alpha-beta-soup feel free to suggest improvements)
     - GDAL-based provider object is loosely specified for now (@justb4 feel free to suggest improvements)
     - text-based properties (strings or arrays) have custom types (`multilingual_string` and `multilingual_array`) to be able to handle language-aware (or not) configuration (@GeoSander feel free to suggest improvements)
- jsonschema (now a core dependency per #718) is integrated to validate any config against the schema
- a few updates/fixes were applied to various configurations (nice to have a schema now :) )
- aside: a few boolean test updates which were found were also updated

This is a non-trivial enhancement, so I've asked the PSC and core devs to review.

### DIFF
--- a/pygeoapi/schemas/config/pygeoapi-config-0.x.yml
+++ b/pygeoapi/schemas/config/pygeoapi-config-0.x.yml
@@ -125,13 +125,13 @@ properties:
                 description: server identification
                 properties:
                     title:
-                        $ref: '#/definitions/multilingual_string'
+                        $ref: '#/definitions/i18n_string'
                         description: the title of the service
                     description:
-                        $ref: '#/definitions/multilingual_string'
+                        $ref: '#/definitions/i18n_string'
                         description: some descriptive text about the service
                     keywords:
-                        $ref: '#/definitions/multilingual_array'
+                        $ref: '#/definitions/i18n_array'
                         description: list of keywords about the service
                     keywords_type:
                         type: string
@@ -143,7 +143,7 @@ properties:
                             - theme
                             - stratum
                     terms_of_service:
-                        $ref: '#/definitions/multilingual_string'
+                        $ref: '#/definitions/i18n_string'
                         description: terms of service
                     url:
                         type: string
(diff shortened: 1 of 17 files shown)


pygeoapi/schemas/config/pygeoapi-config-0.x.yml
```diff
@@ -0,0 +1,472 @@
...
+
+definitions:
+    multilingual_string:
+        oneOf:
+            - type: string
+            - type: object
+              patternProperties:
+                  "^[a-z]{2}$":
+                      allOf:
+                          - type: string
```
GeoSander, line 453: Good initiative, @tomkralidis!

However, I would call the type `i18n_string` or something. Technically, the type is language-aware, but does not need to be multi-lingual if the user only sets a single text string and not an object. Unless we decide that this type should always be an object with at least 1 language tag: string key-value pair in it.

Furthermore, the suggested pattern is fine for basic ISO 639-1 language codes (with only 2 chars), but it should also be possible to define localized language tags (e.g. `fr-CA` for Canadian French - see [RFC5646](https://www.rfc-editor.org/rfc/rfc5646.txt)), because the `l10n` module can handle this too. Basically, any code that can be parsed by `Babel` should be valid here, so we should not restrict the pattern too much. The same applies to the array type below, of course.

tomkralidis: Thanks @GeoSander.  I've renamed the types accordingly.  For patterns, should we allow `^.*$` then or is that too wide?

GeoSander: Here is a [complete list of all supported tags](http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry). I think that we can support the majority of the  tags by using `^[a-zA-Z]{2,3}([-_][a-zA-Z0-9]{2,3})?$` perhaps? In other words: 2-3 chars, optionally followed by a dash/underscore and 2-3 digits/chars.

tomkralidis: Thanks @GeoSander.  Updated with new regexes.

kalxas (approved):

GeoSander (approved):

PR #725, opened 2021-07-10
add translation system to pygeoapi

pvgenuchten:
this pr adds a translation system to pygeoapi, see locale/readme.md on how to set up

- extract latest keys in .pot file with: `pybabel extract -F babel-mapping.ini -o locale/messages.pot ./`
- then update .po language file: `pybabel update -d locale -l fr -i locale/messages.pot`
- update translations and compile .mo file: `pybabel compile -d locale -l fr`

Run pygeoapi and switch language to french ?lang=fr, notice some hardcoded strings translated.


pygeoapi/util.py
```diff
@@ -341,7 +346,9 @@ def render_j2_template(config, template, data, locale_=None):
     env.filters['filter_dict_by_key_value'] = filter_dict_by_key_value
     env.globals.update(filter_dict_by_key_value=filter_dict_by_key_value)
 
-    # TODO: insert Babel Translation stuff here
+    translations = Translations.load('locale', [locale_])
+    env.install_gettext_translations(translations)
```
GeoSander, line 350: We might need to check how efficient this is (loading all translations each time when `render_j2_template` is called). If the locale has not changed, there's no need to reload. But perhaps `install_gettext_translations` already has such a mechanism in place.

pygeoapi/util.py
```diff
@@ -44,7 +44,8 @@
 
 import dateutil.parser
 # from babel.support import Translations
```
GeoSander, line 46: Please remove this line :)

GeoSander (approved):
Good start @pvgenuchten and happy to see that it was easy to put the translations in place!
For the links sections (as generated by `api.py`) we should have a look how to properly build in the Babel/gettext stuff there, but it might be better to do this after the API refactor, don't you agree @tomkralidis?

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,325 @@
...
+msgid "Exception"
+msgstr "Exception"
+
+#: pygeoapi/templates/landing_page.html:25
+msgid "Terms of service"
+msgstr "Conditions d'utilisation"
+
+#: pygeoapi/templates/landing_page.html:35
+msgid "License"
+msgstr "License"
```
alexandreleroux, line 56: License in French is spelled `Licence`, thanks

pygeoapi/util.py
tomkralidis: > Good start @pvgenuchten and happy to see that it was easy to put the translations in place!
> For the links sections (as generated by `api.py`) we should have a look how to properly build in the Babel/gettext stuff there, but it might be better to do this after the API refactor, don't you agree @tomkralidis?

+1

locale/readme.md
```diff
@@ -0,0 +1,32 @@
+# Howto set up language files
```
tomkralidis, line 1: Rename file to `locale/README.md`

tomkralidis (changes requested):
Great work here @pvgenuchten !  Minor comments.  I would also add a sentence or two in https://docs.pygeoapi.io/en/latest/language.html

tomkralidis (approved):

GeoSander:
@tomkralidis is it okay to merge this (after rebase)?

tomkralidis:
@GeoSander +1, let me know when rebased.

pvgenuchten:
just did a rebase

PR #726, opened 2021-07-12
Add missing ports to first docker-compose example

aripollak:
This was present in the second example but should also be needed for the first one.


tomkralidis:
CI failures are unrelated.

PR #747, opened 2021-07-27
Modified the postgres provider to allow fields to be specified

### LINKED ISSUE #291: Provide a way to allow an implementer to map provider column names to provider names

jemrobinson:
Added new 'fields' option to pygeoapi-config YAML to allow the user to specify which fields (and associated types) to expose

This is an alternative solution to #303 which seems to have been closed without merging.

Closes #291


tomkralidis:
Added as part of #784 .

PR #749, opened 2021-08-03
Create SECURITY.md

tomkralidis:
FYI, per https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository. Thanks to @jmckenna for the [inspiration](https://lists.osgeo.org/pipermail/mapserver-dev/2021-July/016572.html).

Interested in folks' feedback here about the contact point.  I put myself only because we don't have a dedicated security/private email address.  Should we?  Or is this too much overhead?  I also thought of using [OSGeo's security-priv](https://trac.osgeo.org/osgeo/ticket/1512), but non-members won't be able to send email.  I would also prefer someone else as the contact (or at least two PSC members).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,17 @@
+# pygeoapi Security Policy
+
+## Supported Versions
+
+Security/vulnerability reports **should not** be submitted through GitHub issues or public discussions, but instead please send your report 
+to **tomkralidis nospam @ gmail.com** (remove the blanks and ‘nospam’).  
+
+Please follow the [contributor guidelines](https://github.com/geopython/pygeoapi/blob/master/CONTRIBUTING.md) when submitting a vulnerability report.
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,17 @@
+# pygeoapi Security Policy
+
+## Supported Versions
+
+Security/vulnerability reports **should not** be submitted through GitHub issues or public discussions, but instead please send your report 
+to **tomkralidis nospam @ gmail.com** (remove the blanks and 'nospam').
```
francbartoli, line 6: You can add me as alternate contact so we can reach the minimum number in my opinion. However, it's likely a unique mailing list address dedicated to the security reports would be better

francbartoli (changes requested):

SECURITY.md
tomkralidis: @francbartoli I've updated the PR with a more generic email address.  We can merge once the address is setup.

francbartoli (approved):

Issue #728, opened 2021-07-19
admin GUI


ksonda:
**Is your feature request related to a problem? Please describe.**
Some users have difficulty configuring due to yml syntax issues, and the general length + complexity of the yml when many resources are configured.

**Describe the solution you'd like**
A GUI to configure pygeoapi, perhaps just a simple HTML form, perhaps in React. 

@webb-ben and I are pursuing this on our own to serve some organizations that have expressed interest. If there is interest in this being a core feature, can open a draft PR.

justb4:
+1 . My personal pref would be using VueJS over React. IMO the `pygeoapi` config structure could first be made more 'composable'/'loosely coupled', by allowing e.g. Collections configured as per-collection-file-yaml's in a directory. 
`ldproxy`, another OGC API server, has a 'manager' UI doing something similar with also a directory-based config convention:
https://github.com/interactive-instruments/ldproxy/tree/master/ldproxy-manager

webb-ben:
@justb4 If this is something that is wanted enough, I would be interested in implementing that sort of behavior for the config.  Probably would make sense to happen in a different branch/issue that an admin GUI, right?

krishnaglodha:
Hi @ksonda , I'm thinking of creating a config yml generator to help users use GUI form to generate files. Please let me know if you are working on it, otherwise I can then spin up an issue.

Thanks

ksonda:
@webb-ben can you show where current draft and any feedback is? You're welcome to iterate on that or try a different implementation if you want. Our work on this is not immediate for the time being.

webb-ben:
@krishnaglodha @ksonda cc: @tomkralidis 

This is implemented in [cgs-earth/pygeoapi:admin](https://github.com/cgs-earth/pygeoapi/tree/admin). Simultaneous to the GUI, working is being done to build out an [admin api](https://github.com/wmo-im/wis2box-api/tree/main/wis2box_api) that is in line with [OGC API - Features - Part 4](http://docs.ogc.org/DRAFTS/20-002.html).

krishnaglodha:
HI @webb-ben , I cloned the repo and it's running but when I do `/admin` I'm getting error, can you please tell how can I access admin panel ?

webb-ben:
@krishnaglodha Could you tell me the error you are getting? `/admin` should be redirecting to `/login` if credentials have not been provided.

[...]

PR #756, opened 2021-08-20
Admin Panel

webb-ben:
This is a work in progress implementation of (#728) for an Admin GUI and form of the `/pygeoapi/local.config.yaml` file. As per @ksonda's idea, this represents the begining of such interface implemented with vue/vueify.js and flask-login. The admin page is a recursive table based GUI with related forms to allow for yaml configuration from a docker-compose version of pygeoapi with minimal direct text editing. The workflow has been implemented except for the re-upload of the JSON/dictionary-like representation configuration file into pygeoapi.

I hope I have implemented enough of this to be able to receive more feedback. This is my first front-end project, so I am eager to hear any and all thoughts. I tried to be as lightweight as possible & still plan to clean up a lot of this, specifically the resources configuration section (eventually adding all CRUD actions) as well as the scope of each component. 

For those interested in exploring themselves, the docker-compose file in inside of of the [docker example](https://github.com/webb-ben/pygeoapi/tree/flask-admin/docker/examples/admin), is the best place to get started. The login requirement to access the admin page is currently disabled for ease of debugging, but can be added back by commenting in [line 93](https://github.com/webb-ben/pygeoapi/blob/ebc0ed5c8e099d63dc9564fb6efbc6a71e989f36/pygeoapi/flask_admin.py#L93) of flask_admin.py. Let me know if you have any questions!


pvgenuchten:
Hi Ben, this looks really nice!

image

What are your thoughts on persisting the changed configuration (requires an api method?) and effectuating it on a running instance. Does pygeoapi already has some 'reload-configuration' method?

I know this is early, but some comments:
- close button on edit panel didn't work for me (but clicking outside the panel closes it)
- is it possible to create a 'add resource' button to the resource-listing
- editing the resource list returns [object] as string

image

webb-ben:
Paul, Thanks for your thoughts!

To my knowledge, there is no reload-configuation within the api. My tentative plan was to have the gunicorn server watch the local.config.yaml file and reload the server on an update. That being said, I believe that @justb4 expressed some interest in reformating the configuration file to be more loosely coupled.

All three of your comments were on my to-do list, but I am glad you mentioned them because I think some of them highlight a couple of questions I have myself. I am specifically interested in the last point surrounding the editable sections of the form. I am not exactly sure how l10n translation works and how to best code for it. The GUI blocks should be softcoded, but I started hardcoding the editable block, which is why the resource list returns a form of object Objects.

For better or worse, after the most recent merge to pygeoapi, I am having trouble passing the config to the Vue app. Which I believe is because I am trying to pass the untranslated configuration file. This worked so long as the config only had one language configured. How are you envisioning this to play out? I am happy to adapt as much as is needed.

Thanks again!

webb-ben:
Having made many of the changes that @ksonda and @pvgenuchten suggested, I would be interested in hearing new thoughts on the state of the admin GUI. Specifically, all of the GUI and form blocks render correctly, new feature collections are able to be added to the configuration, and the changes made are persistent in server config. The capacity to add a resource providers other than feature collections is also on the horizon. I think there are also improvements that can be made in the UI/UX sphere, but am having trouble sussing them out.

tomkralidis:
@webb-ben sorry for the delay here.  I took a quick look -- impressive work here!

Is it possible to have this working as a separate Python package (i.e. pygeoapi-admin) or web front end?  Options:
- via Python/Flask, has write access to local configuration YAML. pygeoapi configuration would have a config option like `admin: true`, which would trigger inclusion of routes and functionality in `pygeoapi_admin`
- pure web frontend (say Vue.js) with only API access.  This would mean extending the pygeoapi API to include management functions, i.e.:
  - `/admin/configuration` (GET/POST/PUT/DELETE).  pygeoapi would need a secure setup and an admin password

webb-ben:
@tomkralidis I think this currently is a bit closer to the former of the two. Adding to the pygeoapi config is smart! At the moment there is a handshake with an encrypted password using flask-login, but i'm pretty sure the password remains unencrypted in the  system environment. The authentication is disabled because there is really bad session protection at the moment which makes navigating the pages difficult. There are definitely better setups for authentication and i am happy to think through some of them.

PR #783, opened 2021-09-20
782: Replaced the literal strings 'x_axis_label' and 'y_axis_label' ...

metmikel:
in the query_params submitted to xarray with the values for the axis labels stored in self._coverage_properties.


tomkralidis:
Thanks @metmikel !

Issue #687, opened 2021-05-13
Enable datetime / sortby / properties queries for Postgres provider


KoalaGeo:
**Is your feature request related to a problem? Please describe.**
We have a lot of data in Postgres, it'd be great to connect this to pygeoapi with the full query functionality the same as an elastic search or mongodb provides

**Describe the solution you'd like**
When using Postgres as a provider for OA-Features be able to query datetime, sortby and properties

**Describe alternatives you've considered**
None

**Additional context**
None

ksonda:
@webb-ben @dblodgett-usgs we are probably going to need to do this for the global river runner anyway.

webb-ben:
I've added to the select_properties, skip_geometry, and sortby to the postgres provider when tinkering around for river runner. I don't know if this is still relevant @ksonda but figured I would loop back to this.

PR #784, opened 2021-09-24
Add postgres provider query parameters

webb-ben:
This is a PR to add the skip_geometry, sortby, select_columns to the postgis provider. This, apart from the datetime query parameter, was brought up in #687. However, the motivation for it was in the creation of my [pygeoapi branch](https://github.com/webb-ben/pygeoapi/tree/river-runner) for [river-runner](https://github.com/ksonda/global-river-runner), where the algorithm required the additional query parameters of the postgis provider [(demo)](https://merit-nldi.internetofwater.app/processes/river-runner/map?lat=15.0150&lng=25.694962&sortby=hydroseq&groupby=streamlev).  I figured I would isolate the postgis specific material and add pytest scripts to verify the functionality. I haven't used a database with a datetime field, but could also add that if it could be easily added to the github workflow.


.github/workflows/main.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         python3 tests/load_es_data.py tests/data/ne_110m_populated_places_simple.geojson geonameid
         python3 tests/load_es_data.py tests/cite/ogcapi-features/canada-hydat-daily-mean-02HC003.geojson IDENTIFIER
         python3 tests/load_mongo_data.py tests/data/ne_110m_populated_places_simple.geojson
-        #gunzip < tests/data/hotosm_bdi_waterways.sql.gz | psql -U postgres test
+        gunzip < tests/data/hotosm_bdi_waterways.sql.gz | psql postgresql://postgres:postgres@localhost:5432/test
```
tomkralidis, line 63: Can we use `${{ secrets.DatabasePassword }}` as the password?

tomkralidis (changes requested):

.github/workflows/main.yml
webb-ben: I switched because the workflow (and thus checks) failed when a secret wasn't set. But I think we could do something like `postgresql password: ${{ secrets.DatabasePassword || postgres }}`. The other part is that the test_postgresql_provider.py also needs access to the database password.
 [here](https://github.com/geopython/pygeoapi/blob/3de5271d0b768d4717ca64f6ae86d3db7da4ec0f/tests/test_postgresql_provider.py#L45-L48):  
```
 'data': {'host': '127.0.0.1',
          'dbname': 'test',
          'user': 'postgres',
          'password': 'postgres',
```
Would that have to be set as an environment variable in the workflow and then retrieved for the test?

webb-ben: @tomkralidis  if `${{ secrets.DatabasePassword || postgres }}` works for you, I have a PR primed with it implemented.

tomkralidis: OK. Or can we wrap the update into this PR?

PR #793, opened 2021-10-07
detect exact match for ES property queries

tomkralidis:
This PR fixes an ES provider issue where OAFeat `prop=value` type queries returned erroneous results when values with spaces were specified (i.e. `prop=foo bar`).  The fix forces an exact match except for when `|` is passed to allow for boolean search.


tomkralidis:
@francbartoli not sure if the ES DSL code needs the same type of updates?

PR #829, opened 2021-12-01
Add support for config file properties to PostgreSQL provider

### LINKED ISSUE #828: Support properties config option in PostgreSQL
**Description**
The [documentation](https://docs.pygeoapi.io/en/latest/configuration.html) states that the `properties` keyword in the config file should work as follows:

```
properties:  # optional: only return the following properties, in order
    - stn_id
    - value
```

This is not currently supported by the `postgres` provider

**Steps to Reproduce**
- Set up a PostgreSQL table
- Use the config to specify a subset of the available columns
- Note that all columns are returned regardless of what this is set to

**Expected behavior**
Only the requested columns should be returned

**Screenshots/Tracebacks**
n/a

**Environment**
 - OS: `macOS 11.5.2`
 - Python version: `3.9.6`
 - pygeoapi version: `0.11.0`

**Additional context**
n/a

jemrobinson:
Add support for the properties keyword from the config file

Closes #828

### DIFF
--- a/pygeoapi/provider/postgresql.py
+++ b/pygeoapi/provider/postgresql.py
@@ -60,7 +60,7 @@ class DatabaseConnection:
      The class returns a connection object.
     """
 
-    def __init__(self, conn_dic, table, columns=[], context="query"):
+    def __init__(self, conn_dic, table, properties=[], context="query"):
         """
         PostgreSQLProvider Class constructor returning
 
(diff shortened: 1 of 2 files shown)


tomkralidis:
Thanks @jemrobinson, cc'ing additional developers for review/comment.  A couple of comments:

- looks like there are some flake8 errors
- can we add to the [PostgreSQL provider tests](https://github.com/geopython/pygeoapi/blob/master/tests/test_postgresql_provider.py)?  Suggest creating another config fixture with properties set and run tests to confirm it works

jemrobinson:
@tomkralidis thanks for pointing those out. I've fixed the `flake8` errors and added a test. Let me know if you'd like me to make any other changes.

tomkralidis:
Thanks @jemrobinson ; looks like there are some flake8 failures in `tests/test_postgresql_provider.py`

jemrobinson:
Sorry! Should be fixed now.

tomkralidis:
Thanks @jemrobinson !

PR #855, opened 2022-02-11
854: Added support for the coverage/metadata path…

metmikel:
… and a simple Jinja2 template for displaying it when requested in HTML format.


metmikel:
Realised that the handling of metadata in the Jinja2 template is hopelessly naïve - it will not handle non-string types or nested objects.
Also need to check what links need to be added to the coverage data (I might need some direction there as there is already a link there to the Coverage "metadata" - the [coverageInfo](https://github.com/opengeospatial/ogcapi-coverages/blob/master/standard/openapi/schemas/coverages-core/coverageInfo.yaml)) & check what else needs to be added - OpenAPI description, unit tests?

PR #857, opened 2022-02-17
Creation of a STAC HATEOAS Provider.

ychoquet:
You will find in this pull request a provider that allows to display a STAC catalog respecting the HATEOAS format as described on the Radiant Earth website (https://github.com/radiantearth/stac-spec).

As mentioned on their site, the STAC specifications define related JSON object types connected by link relations to support a HATEOAS-style traversable interface and a RESTful API providing additional browse and search interfaces.

### DIFF
--- a/docs/source/data-publishing/stac.rst
+++ b/docs/source/data-publishing/stac.rst
@@ -1,13 +1,305 @@
 .. _stac:
 
 Publishing files to a SpatioTemporal Asset Catalog
-==================================================
+**************************************************
 
-The `SpatioTemporal Asset Catalog (STAC)`_ specification provides an easy approach
-for describing geospatial assets.  STAC is typically implemented for imagery and
-other raster data.
+The `SpatioTemporal Asset Catalog (STAC)`_ family of specifications aim to standardize
+the way geospatial asset metadata is structured and queried. A "spatiotemporal asset"
+is any file that represents information about the Earth at a certain place and time.
+The original focus was on scenes of satellite imagery, but the specifications now cover
+a broad variety of uses, including sources such as aircraft and drone and data such as
+hyperspectral optical, synthetic aperture radar (SAR), video, point clouds, lidar, digital
+elevation models (DEM), vector, machine learning labels, and composites like NDVI and
+mosaics. STAC is intentionally designed with a minimal core and flexible extension mechanism
+to support a broad set of use cases. This specification has matured over the past several
+years, and is used in numerous production deployments.
 
-pygeoapi implements STAC as an geospatial file browser through the FileSystem provider,
+pygeoapi has two built-in providers to browse STAC catalogs: `FileSystem Provider`_ and
+`Hateoas Provider`_.
+
+Hateoas Provider
+================
+
+HATEOAS (Hypermedia as the Engine of Application State) is a way of implementing a REST
+application that allows the client to dynamically navigate to the appropriate resources
+by browsing hypermedia links. This type of navigation is similar to WEB navigation
+and requires a very precise data structure that must be respected to allow the HATEOAS
+Provider to behave correctly.
+
+There are three component specifications (Catalog, Collection, Item) that together make
+up the core SpatioTemporal Asset Catalog specification. An Item represents a single
+spatiotemporal asset as GeoJSON. The Catalog specification provides structural elements,
+to group Items and Collections. Collections are catalogs, that add more required metadata
+and describe a group of related Items.
+
+The full catalog structure of links down to sub-catalogs and Items, and their links back to
+their parents and roots, must be done with **relative** URL's for the HATEOAS Provider work
+correctly. The structural *rel* types include *root*, *parent*, *child*, *item*, and
+*collection*. Assets links must be **absolute** URL's. Other links can be absolute, especially
+if they describe a resource that makes less sense in the catalog, like derived_from or even
+license (it can be nice to include the license in the catalog, but some licenses live at a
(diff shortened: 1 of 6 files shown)


tomkralidis:
Fantastic work @ychoquet!  Some comments:
- what is the requirement for the `tests/stac` files?
- we don't need another sample configuration (`example-config.yml`); instead, some documentation can be added on the new plugin in https://github.com/geopython/pygeoapi/blob/master/docs/source/data-publishing/stac.rst

ychoquet:
Hello @tomkralidis ,
Have a look at the doc I made. The example-config.yml is deleted. All information needed to create a stac entry have been moved to the doc.

ychoquet:
Hello @tomkralidis ,
Must do a correction to the doc. I'll tell you when it is done.

ychoquet:
Hello @tomkralidis ,
I'm done now!

docs/source/data-publishing/stac.rst
```diff
@@ -1,13 +1,305 @@
...
+
+3. Sub-Catalogs or sub-Collections must be stored in subdirectories of their parent (and only 1 subdirectory deeper than a document's parent, e.g. .../sample/sub1/catalog.json).
+
+4. Limit the number of Items in a Catalog or Collection, grouping / partitioning as relevant to the dataset.
+
+5. Use structural elements (Catalog and Collection) consistently across each 'level' of your hierarchy. For example, if levels 2 and 4 of the hierarchy only contain Collections, don't add a Catalog at levels 2 and 4.
+
+6. Items must be named <*id*>.json.
+
+7. Items must be stored in subdirectories (1 level deeper) of their parent Catalog or Collection. The subdirectory must have the same name (<*id*>) as the Item without the *.json* extension. This means that each Item are contained in a unique subdirectory.
```
abkfenris, line 59: I'm wondering why there is this limitation? Additionally why have limitations on limits on depth hierarchy at all between catalogs, collections, and items? In the spec it shows catalogs, collections, and items all at the same depth.

For instance, most of the time that I encounter forecast model data, all the run outputs are in the same directory. Instead of moving all the items into their own directory, it would make sense to have them along side the source data, and potentially at the same depth as a collection containing all of those items.

tomkralidis:
@ychoquet thanks.  What is the requirement for the `tests/stac` files?

ychoquet:
What is the requirement for the tests/stac files?

@tomkralidis The tests/stac folder contains an example that shows how to deploy a local STAC Catalog. It follows the same rules as a WEB deployment described in the documentation. The only difference between the two implementation types is shown in the [HATEOAS Configuration](https://github.com/ychoquet/pygeoapi/blob/STAC-HATEOAS-Provider/docs/source/data-publishing/stac.rst#hateoas-configuration) section. More precisely in the data property of the provider.

Not sure I answer your question. Does it?

tomkralidis:
Thanks.  Unless the files are part of the tests, I would rather exclude them from this PR.

ychoquet:
@tomkralidis No, they are not part of the test. I only provided them as an example. Since the documentation shows how to recreate them, we can exclude them from the PR. I have deleted them.

pygeoapi/provider/hateoas.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    Helper function used to load a json file that is located on the WEB (HTTP request)
+    or on the server file system
+
+    :param jsonpath: path to the json file
+
+    :returns: `dict` of JSON item
+    """
+
+    if jsonpath[0:4].upper() == 'HTTP':
```
tomkralidis, line 200: Do we need to account for HTTPS here i.e. options:

- `if jsonpath[0:4].upper().startswith('HTTP'):`
- `if jsonpath[0:4].upper() in ["HTTP', 'HTTPS']:`?

pygeoapi/provider/hateoas.py
```diff
@@ -0,0 +1,205 @@
...
+        jsondata = requests.get(jsonpath).json()
+    else:
+        jsondata = json.load(open(jsonpath))
```
tomkralidis, line 203: Should this be wrapped in a context manager, i.e.:

```python
    else:
        with open(jsonpath) as fh:
            jsondata = json.load(fh)
```
```

tomkralidis (changes requested):

pygeoapi/provider/hateoas.py
ychoquet: The code, as it is, accepts both protocols since jsonpath[0:4].upper() returns the first four characters in uppercase and the first four characters of HTTPS are equal to HTTP. That said, option 1 is equivalent to "if jsonpath[0:4].upper() == 'HTTP':". However, it processes the entire path which is not necessary. Although option 2 is also equivalent to my code, it seems a bit strange because jsonpath[0:4].upper() returns the first 4 characters and HTTPS has 5 characters.

pygeoapi/provider/hateoas.py
ychoquet: If you prefer this syntax, I can change my code to yours.

tomkralidis: Thanks; this is the preferred codepath to reading/writing files.

tomkralidis:
@ychoquet looks like some flake8 errors on `pygeoapi/provider/hateoas.py` that need to be addressed.

ychoquet:
I have corrected the flake8 errors.

tomkralidis:
Excellent work @ychoquet -- thank you for this valuable contribution!

Issue #861, opened 2022-03-04
OGR:WFS Support for CONVERT_TO_LINE


reinij:
**OGR:WFS can't handle CURVEPOLYGON**

e.g.
`CURVEPOLYGON ((9.8195 55.3263,9.8258 55.3209,9.8265 55.3201,9.8294 55.311,9.8298 55.3093,9.8298 55.3082,9.8294 55.3048,9.8285 55.3019,9.8274 55.3009,9.8257 55.3,9.8247 55.2997,9.8165 55.2987,9.8024 55.2965,9.7984 55.2989,9.7949 55.3031,9.7951 55.3036,9.7954 55.3054,9.7951 55.3082,9.7945 55.3104,9.7941 55.3112,9.7932 55.3122,9.7903 55.3143,9.7895 55.3153,9.7896 55.3162,9.7937 55.3238,9.7952 55.3256,9.7983 55.326,9.7993 55.326,9.8035 55.3234,9.8028 55.3225,9.8028 55.3218,9.8052 55.3191,9.806 55.3189,9.8101 55.3191,9.8114 55.3192,9.8118 55.3198,9.8115 55.3222,9.8111 55.323,9.8094 55.3232,9.8085 55.3231,9.8091 55.324,9.8156 55.326,9.8177 55.3262,9.8195 55.3263))
`

gives an error

```
[2022-03-04T11:19:17Z] {\pygeoapi\provider\ogr.py:819} ERROR - Error Number: 6, Type: Failure, Msg: OGR geometry type unsupported as a GeoJSON geometry detected. Feature gets NULL geometry assigned.
[2022-03-04T11:19:17Z] {\pygeoapi\provider\ogr.py:367} ERROR - <built-in function Geometry_ExportToJson> returned a result with an exception set
```

Adding parameter **gdal_ogr_options: CONVERT_TO_LINEAR=YES** (into config.yml) and subsequent code into **ogr.py** would fix an issue.

reinij:
https://github.com/geopython/pygeoapi/pull/863

justb4:
Ok, good catch! Curves show up from time to time. I see that there is a whole bunch:
https://gdal.org/development/rfc/rfc49_curve_geometries.html

[...]

PR #863, opened 2022-03-08
Update ogr.py to enable CURVEPOLYGON

reinij:
Allow to use CURVEPOLYGON from WFS forcing geometry to POLYGON to enable coversion to geoJSON.

### DIFF
--- a/pygeoapi/provider/ogr.py
+++ b/pygeoapi/provider/ogr.py
@@ -457,10 +457,16 @@ class OGRProvider(BaseProvider):
 
     def _ogr_feature_to_json(self, ogr_feature):
         geom = ogr_feature.GetGeometryRef()
+        
+        #convert curvepolygon to polygon as GeoJSON doesn't support curves
+        if geom.GetGeometryType() == osgeo_ogr.wkbCurvePolygon:
+            geom_polygon = osgeo_ogr.ForceToPolygon(geom)
+            ogr_feature.SetGeometry(geom_polygon)
+        
         if self.transform_out:
             # Optionally reproject the geometry
             geom.Transform(self.transform_out)
-
+        
         json_feature = ogr_feature.ExportToJson(as_object=True)
         try:
             json_feature['id'] = json_feature['properties'].pop(self.id_field)


pygeoapi/provider/ogr.py
```diff
@@ -457,10 +457,16 @@ def _get_next_feature(self, layer, feature_id):
...
+        #convert curvepolygon to polygon as GeoJSON doesn't support curves
+        if geom.GetGeometryType() == osgeo_ogr.wkbCurvePolygon:
+            geom_polygon = osgeo_ogr.ForceToPolygon(geom)
+            ogr_feature.SetGeometry(geom_polygon)
+        
         if self.transform_out:
             # Optionally reproject the geometry
             geom.Transform(self.transform_out)
-
+
```
tomkralidis, line 469: This should be removed as it will result in flake8 errors.

pygeoapi/provider/ogr.py
```diff
@@ -457,10 +457,16 @@ def _get_next_feature(self, layer, feature_id):
 
     def _ogr_feature_to_json(self, ogr_feature):
         geom = ogr_feature.GetGeometryRef()
+        
+        #convert curvepolygon to polygon as GeoJSON doesn't support curves
+        if geom.GetGeometryType() == osgeo_ogr.wkbCurvePolygon:
```
tomkralidis, line 462: Curious, do we need to consider other geometry types that may need to be forced into polygon geometries?

tomkralidis (changes requested):

pygeoapi/provider/ogr.py
justb4: Yes, complicated. Hence my question in #861: *I see that there is a whole bunch:
https://gdal.org/development/rfc/rfc49_curve_geometries.html* . I  am no expert on this but:

* is this PR solving only a single use-case (single non-linear geometry)?
* can this be generically solved via an OGR option setting?

francbartoli:
Hey @reinij, what is the status of this PR? Do you like to resolve the comments or get help from us?

pvgenuchten:
interesting topic, also consider that after some discussion on the feat-geom group there is now a proposal to add special geometry types to feat-geom, see https://github.com/opengeospatial/ogc-feat-geo-json/blob/main/proposals/circular-geometry-objects.adoc

tomkralidis:
@reinij can we get a status on this PR?

reinij:
I currently have the feeling that this should be canceled and let it be as it is now, i.e. the geometries are not automatically converted but it is assumed that the GML is a valid GeoJSON geometry.

The future JSON-FG then offers all the necessary geometries.

PR #868, opened 2022-03-10
Add json-ld templating for feature collection item

webb-ben:
Addressing https://github.com/cgs-earth/pygeoapi/issues/1, this PR adds the capacity to render complicated json-ld templates using jinja2 and the default json-ld output from pygeoapi.

- Check for template when generating JSON-LD document for an individual collection item.
- Default `@id` instead of aliasing with `@context`. Makes easier to render jinja2 template with the appropriate fields.
- Rename default `geosparql` to json-ld convention `gsp`
- Include basic json-ld template generation in test workflow.
- Add documentation for templating.


docs/source/configuration.rst
```diff
@@ -335,6 +335,19 @@ This relationship can further be maintained in the JSON-LD structured data using
             ssn: "http://www.w3.org/ns/ssn/"
             Datastream: sosa:isMemberOf
 
+Sometimes, the JSON-LD desired is more complicated than can be achieved by simply aliasing properties using a
```
tomkralidis, line 338: We should articulate that this functionality is triggered for a single feature collection item.

docs/source/configuration.rst
```diff
@@ -335,6 +335,19 @@ This relationship can further be maintained in the JSON-LD structured data using
...
+context. In thise case, it is possible to specify a jinja2 template. When ``template`` is defined for a feature
+collection, the json-ld prepared by pygeoapi will be used to render a jinja2 template. pygeoapi provides a default
+template, which will render the unmodified json-ld document.
+
+.. code-block:: yaml
+
+    template: jsonld/base.jsonld
+    context:
+        - datetime: https://schema.org/DateTime
+
```
tomkralidis, line 348: Add a note that the path can the template path can be relative to the location of the configuration, or absolute.  Also add that one can use environment variables to have better deployment flexibility of template locations

pygeoapi/api.py
```diff
@@ -1953,6 +1953,13 @@ def get_collection_item(self, request: Union[APIRequest, Any],
                 self.config, content, dataset, uri, (p.uri_field or 'id')
             )
 
+            template = self.config['resources'][dataset].get('template')
+            if template:
```
tomkralidis, line 1957: add logging:

```python
if template is not None:
    LOGGER.debug('Rendering JSON-LD template: {}'.format(template))
    ...
```

pygeoapi/templates/jsonld/base.jsonld
```diff
@@ -0,0 +1 @@
+{{ data | to_json | safe }}
```
tomkralidis, line 1: It looks like this file/directory is only used by the tests.  Suggest to remove from `pygeoapi/templates` and store/ref in `tests/data`.

docs/source/configuration.rst
```diff
@@ -335,6 +335,19 @@ This relationship can further be maintained in the JSON-LD structured data using
...
             Datastream: sosa:isMemberOf
 
+Sometimes, the JSON-LD desired is more complicated than can be achieved by simply aliasing properties using a
+context. In thise case, it is possible to specify a jinja2 template. When ``template`` is defined for a feature
+collection, the json-ld prepared by pygeoapi will be used to render a jinja2 template. pygeoapi provides a default
+template, which will render the unmodified json-ld document.
+
+.. code-block:: yaml
+
+    template: jsonld/base.jsonld
```
tomkralidis, line 345: Given the functionality is scoped to a single feature collection item, what about renaming this directive to `item_template`?

Thinking more, what about something like:

```yaml
item_templates:
    - jsonld: /path/to/file.jsonld
    - xml: /path/to/file.xml
```

tomkralidis: Note: whichever way we decide, this optional property will need to be added to the config schema in `pygeoapi/schemas/config/pygeoapi-config-0.x.yml`.

tomkralidis (changes requested):

tomkralidis:
Great work @webb-ben! See change requests. I've also added @alpha-beta-soup to review.

docs/source/configuration.rst
webb-ben: At the moment, `template:  /path/to/file.xml` renders at `.../items/{id}?f=jsonld`. I like your suggestion ... but does that not encroach a bit upon the formatter provider? How does pygeoapi serve the correct `f=__` for defined templates?

pygeoapi/linked_data.py
```diff
@@ -188,16 +188,16 @@ def geojson2jsonld(config, data, dataset, identifier=None, id_field='id'):
         :returns: string of rendered JSON (GeoJSON-LD)
     """
     context = config['resources'][dataset].get('context', []).copy()
+
     defaultVocabulary = {
         'schema': 'https://schema.org/',
-        id_field: '@id',
```
alpha-beta-soup, line 193: Sorry, I've been out of this space for a while, so this may be a dumb comment, but why are we enforcing the use of '@id'? I thought the purpose of this was to allow an existing data source to use whatever property it liked as an effective `@id` without having to literally call it that? Essentially making otherwise naive data somewhat context-aware by laying a minimal LD vocabulary on top without requiring any adjustment to data. Or is this what templating is solving?

pygeoapi/schemas/config/pygeoapi-config-0.x.yml
```diff
@@ -257,6 +257,9 @@ properties:
                           keywords:
                               $ref: '#/definitions/i18n_array'
                               description: list of keywords about the service
+                          item_template:
+                            ype: string
```
alpha-beta-soup, line 261: typo `ype`/`type`

PR #872, opened 2022-03-11
Add unit test for tinydb job manager with parallel requests

MartinPontius:
This PR adds a unit test to test the behaviour of the default tinydb job manager during parallel processing requests using Python's multiprocessing package. The built-in "hello-world" process is used.

### DIFF
--- a/tests/test_tinydb_manager_for_parallel_requests.py
+++ b/tests/test_tinydb_manager_for_parallel_requests.py
@@ -1,0 +1,96 @@
+import os
+
+from werkzeug.wrappers import Request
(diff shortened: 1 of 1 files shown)


tests/test_tinydb_manager_for_parallel_requests.py
```diff
@@ -0,0 +1,96 @@
+import os
```
tomkralidis, line 1: Can you add a source code header?  Example: https://github.com/geopython/pygeoapi/blob/master/tests/test_util.py#L1-L28

tomkralidis (approved):

Issue #796, opened 2021-10-12
URL links drop ending parentheses in HTML


KoalaGeo:
**Description**
When connecting to SensorThingAPI the @iot.selfLink is dropping the final ) from the Frost Server URL - in this case = http://sensors-internal.bgs.ac.uk/FROST-Server/v1.1/Things(4)

In `?f=json` view the URL is correct

**Steps to Reproduce**
Run pygeoapi in docker connecting to a SensorThingsAPI provider

**Expected behavior**
URL link function in html view 

**Screenshots/Tracebacks**
![image](https://user-images.githubusercontent.com/33657876/136968878-3b157738-9cba-4800-866f-2ede4fed6fd5.png)

![image](https://user-images.githubusercontent.com/33657876/136968804-18859aa6-42e9-4da8-a20d-a7cc732d0b6f.png)

**Environment**
 - OS: Docker image
 - Python version: ^
 - pygeoapi version: :latest

**Additional context**
N/A

tomkralidis:
Iis this now fixed, given #876?

Issue #715, opened 2021-06-16
Consider upgrading to Jinja 3.X


webb-ben:
**Description**

As described in pallets/jinja#827, the urlize() function incorrectly removes a trailing close parentheses from URLs when creating a hyperlink in the HTML landing pages.

**Steps to Reproduce**

- To view in pygeoapi with Docker: 
1) Run `docker run -p 5000:80 --rm webbben/pygeoapi-sensorthings:1.2.2` which is an example STA pygeoapi endpoint with many links ending in close parentheses.
2) Navigate to [http://localhost:5000/collections/Datastreams/items/8862231](http://localhost:5000/collections/Datastreams/items/8862231)
![image](https://user-images.githubusercontent.com/40066515/122231953-df2bdd00-ce88-11eb-89f6-d6dee7834b50.png)
As can be seen in the bottom left, and the `@iot.selfLink`, the trailing `)` is dropped from the hyperlink.
- To view in python, running in the virtual environment that pygeoapi is running on:
```
from jinja2.utils import urlize
example_url = u'https://wikipedia.org/wiki/geopython(pygeoapi)'
urlize(example_url)
```
With (default, Jinja2==2.10.1): `urlize(example_url)` -> `'<a ref="https://wikipedia.org/wiki/geopython(pygeoapi"> https://wikipedia.org/wiki/geopython(pygeoapi</a>)'`

**Expected behavior**
![image](https://user-images.githubusercontent.com/40066515/122237351-4350a000-ce8d-11eb-82ef-2b80b77c768c.png)

With (Jinja2==3.0.1): `urlize(example_url)` -> `'<a ref="https://wikipedia.org/wiki/geopython(pygeoapi)"> https://wikipedia.org/wiki/geopython(pygeoapi)</a>'`

**Environment**

 - OS: Any
 - Python version: 3
 - pygeoapi version: master/latest

**Additional context**

This is a problem with Jinja2 that has already been fixed in v3.X which was released in May of 2021. I am not aware of outside reasons to wait before upgrading to the newest version of Jinja2, as v2.10.1 installed as the default dependency with Flask.

tomkralidis:
cc @kalxas 

Thanks for the report.  I think we need to wait until Jinja2 3.x makes its way to Ubuntu/Debian packaging.

tomkralidis:
@webb-ben is this now fixed, given #876?

PR #876, opened 2022-03-27
Pin jinja2 to fix buggy version

### LINKED ISSUE #875: Build is currently broken caused by latest jinja2 version
**Description**
Getting this error in our builds since the latest release 3.1.0 of Jinja2.

`module 'jinja2.utils' has no attribute 'escape'`

More insights [here](https://github.com/sphinx-doc/sphinx/issues/10289)

**Steps to Reproduce**
Run a build

**Expected behavior**
Build shouldn't fail

**Screenshots/Tracebacks**

**Environment**
 - OS:
 - Python version:
 - pygeoapi version:

**Additional context**

francbartoli:
Close #875

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ tinydb
 unicodecsv
 future-annotations
 pydantic
+jinja2==3.0.3


tomkralidis:
FYI Jinja2 has since made a bugfix release (3.1.1) which fixes this issue (we ran into this in pygeometa). We should be okay at this point without having to pin.

francbartoli:
> FYI Jinja2 has since made a bugfix release (3.1.1) which fixes this issue (we ran into this in pygeometa). We should be okay at this point without having to pin.

It is still failing for me https://github.com/geopython/pygeoapi/runs/5716551522?check_suite_focus=true

jaidisido:
Failing for me too

justb4:
Think this is due to Flask. Think we use 1.1.2 and that version does not pin jinja2: see
https://github.com/pallets/flask/issues/4494

justb4:
This is how I solved in GeoHealthCheck: https://github.com/geopython/GeoHealthCheck/issues/423 . So pinning these two:

```
Jinja2==2.11.3
markupsafe==2.0.1
```
`markupsafe` is for Sphinx doc but also used by Jinja (or other way around).

justb4 (commented):
See my comment: https://github.com/geopython/pygeoapi/pull/876#issuecomment-1080778923 . You may try the same fix: 
```
Jinja2==2.11.3
markupsafe==2.0.1

```

francbartoli:
> See my comment: [#876 (comment)](https://github.com/geopython/pygeoapi/pull/876#issuecomment-1080778923) . You may try the same fix:
> 
> ```
> Jinja2==2.11.3
> markupsafe==2.0.1
> ```

@justb4 your changes would impact over two different requirements files since markupsafe comes from sphinx. Pinning to the latest working version is still safely passing the build. Can we merge as it is?

justb4:
Ok, yes that was the GHC fix, where the docs are part of the app. Ok, whatever pinned version works as @francbartoli [indicated that 3.0.3 fails for him](https://github.com/geopython/pygeoapi/pull/876#issuecomment-1080332841). Maybe thus Jinja2 2.11.3. If the Sphinx build does not fail, no need to fix.

webb-ben:
fwiw this would also address #715 and #796

Issue #886, opened 2022-04-14
skipGeometry not implemented for geopackage provider


ksonda:
skipGeometry query parameter is not implemented for geopackage provider

This is a port from https://github.com/internetofwater/geoconnex.us/issues/156 

Compare:

https://demo.pygeoapi.io/master/collections/ogr_gpkg_poi/items?f=json&skipGeometry=true 
https://demo.pygeoapi.io/master/collections/lakes/items?f=json&skipGeometry=true

tomkralidis:
Addressed in #887

tomkralidis:
Now fixed/merged in #887.  @webb-ben can you validate?

ksonda:
It seems to work correctly on /items?f=json&skipGeometry=true

although it does not work on /items/id?f=json&skipGeometry=true

This is fine for our purposes but not sure if skipGeometry is supposed to also work on /items/id endpoint

tomkralidis:
I would leave to `/items` for now, given at the `.../{itemId}` level we are at much smaller payload size (always a single feature).

tomkralidis:
P.S. to add, note that pygeoapi's `skipGeometry` support is at the `.../items` level.  Does this work for you @ksonda ?

ksonda:
Yeah that's fine. Don't need it at /item/id level.

PR #887, opened 2022-04-18
add skip_geometry to remaining feature providers

tomkralidis:
Addresses #886

### DIFF
--- a/docs/source/data-publishing/ogcapi-features.rst
+++ b/docs/source/data-publishing/ogcapi-features.rst
@@ -15,17 +15,17 @@ pygeoapi core feature providers are listed below, along with a matrix of support
 parameters.
 
 .. csv-table::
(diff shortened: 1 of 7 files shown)


francbartoli (commented):
@tomkralidis it looks good to me but I have just one general concern about the validity of GeoJSON. As per the specification https://datatracker.ietf.org/doc/html/rfc7946#section-3.1 it seems that a valid one must have a `geometry` object with a `type`.
Effectively using this linter https://geojsonlint.com/ a `FeatureCollection` with empty `geometry` wouldn't be considered valid.
Thoughts?

francbartoli:
Looking deeper into the specification, in particular in section [2.2](https://geojson.org/geojson-spec.html#feature-objects):

- the first bullet point clarifies my concern!

Thanks @tomkralidis, great issue fixed

PR #889, opened 2022-04-25
Fix SQL injection vulnerability. Fix bug with schema prefixed tables.

bitner:
Make sure that we use sql injection resistant formatting from psycopg2 rather than python string formatting for query to get column information from Postgresql provider.

Make sure that we only get columns from a single table. As the column query is now, if there is a public.tablename table and a myschema.tablename table, the columns list will be all columns that are in either of those tables.


tomkralidis:
Thanks @bitner!

PR #892, opened 2022-05-01
Fix wrong path convertor in starlette routes

francbartoli:
This closes the issue on wrong path convertors in the starlette application which raises the exception described [here](https://github.com/geobeyond/fastgeoapi/issues/5)


francbartoli:
This is not complete because it would break the case when a `collection_id` is composed of hierarchical collections. An example in pygeoapi configuration would be:

```yaml
resources:
    foo/bar/baz/obs:
        type: collection
        title: Observations
        description: My cool observations
```

tomkralidis:
Closing in lieu of #893

PR #894, opened 2022-05-07
Update requirement.txt

Youngestdev:
I have updated the `click` requirement which was originally set to `<7` to `>8`. It is advised to update the requirement as there's a security issue for versions < 8.

This change is needed for the effective functioning of the FastGeoAPI application as the dependency causes the build test to fail.

Signed-off-by: Abdulazeez Abdulazeez Adeshina <youngestdev@gmail.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 Babel
-click<8
+click>8
 Flask
 jsonschema
 pyproj
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 Babel
-7 < click <= 8
+click>7,<=8
 Flask
 jsonschema
 pyproj
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 Babel
-click>8
+7 < click <= 8
 Flask
 jsonschema
 pyproj


tomkralidis:
Thanks @Youngestdev.  Can you point to the security issue here for reference?  Thanks.

Youngestdev:
https://github.com/pallets/click/issues/1752. Here you go, @tomkralidis.

kalxas:
Does this mean that pygeoapi will be available as a package only after Ubuntu 22.04?
https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python3-click&searchon=names

Youngestdev:
> Does this mean that pygeoapi will be available as a package only after Ubuntu 22.04?
> https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python3-click&searchon=names

I’m not sure I understand.

francbartoli:
@kalxas if the Debian requirement can be relaxed then I guess we can merge right? And the pygeoapi Debian package can continue to exist for Ubuntu 20.04?

kalxas:
Will pygeoapi actually work with previous click version? If not, relaxing the requirement in Debian would not make a difference as the package will be broken

francbartoli:
@tomkralidis @kalxas does it make sense to have `7 < click <= 8` to support the current packaging and even a safer one for those upstream projects which would like to pin the secure version of that? @Youngestdev thoughts?

Youngestdev:
I don’t know if this’ll work. I’ll test it.

francbartoli:
@kalxas I have relaxed the click requirement to support both versions including the safer one. Might we approve now?
I left my +1, thanks

francbartoli:
@kalxas I have relaxed the click requirement to support both versions including the safer one. Might we approve now?
I left my +1, thanks

francbartoli:
@kalxas I have relaxed the click requirement to support both versions including the safer one. Might we approve now?
I left my +1, thanks

francbartoli (approved):

tomkralidis (approved):

kalxas (approved):

Issue #901, opened 2022-05-15
return html if no accept encoding is requested


pvgenuchten:
Somehow i missed the aspect that google these days does not fetch any structured content from demo.pygeoapi.io since mid 2020, it seems related to the fact that google crawler does not send an accept header when fetching content in combination with pygeoapi returning json, if no accept header is send.

The yandex structured data validator has the same 'challenge'.

Suggestion would be to return html when no accept-header or ?f={encoding} is provided.

ksonda:
I agree

pvgenuchten:
a potential workaround if people are not enthousiastic on this:
- add a sitemap.xml which references each of the enpoints with explicit ?f=html (or better xxx.html)
- the html would indicate xxx.html as canonical url

PR #902, opened 2022-05-15
set F_HTML as default if no &f={enc} or accept header is provided

pvgenuchten:
resolves issue #901
plz consider that there may be regression for clients which do not send accept header and expect json to be returned

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -38,7 +38,7 @@ The ``server`` section provides directives on binding and high level tuning.
         host: 0.0.0.0  # listening address for incoming connections
         port: 5000  # listening port for incoming connections
     url: http://localhost:5000/  # url of server
-    mimetype: application/json; charset=UTF-8  # default MIME type
+    mimetype: application/json; charset=UTF-8  # default MIME type (if no accept header is provided)
     encoding: utf-8  # default server encoding
     language: en-US  # default server language
     gzip: false # default server config to gzip/compress responses to requests with gzip in the Accept-Encoding header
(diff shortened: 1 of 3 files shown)


GeoSander:
Hi @pvgenuchten, I see the need for such a fix (although I still find it odd that Google, Yandex, etc. apparently do not send an `Accept` header), but I am not sure if simply returning `F_HTML` instead of `None` is the right solution. Note that this will affect `APIRequest.is_valid()` too.

Perhaps we should make this configurable in the YAML, i.e. let the admin decide what the `default_format` is if nothing was specified from [one of these values](https://github.com/geopython/pygeoapi/blob/ce425887d0ff0a17212fa56e264396f8bf3874e0/pygeoapi/api.py#L86-L88) (and by default we can set that to html of course). For example, something like:
```yaml
server:
    bind:
        host: 0.0.0.0
        port: 5000
    url: http://localhost:5000
    mimetype: application/json; charset=UTF-8
    encoding: utf-8
    default_format: html
    ...
```

If you don't wish to make this configurable, another option could be to introduce a `requested_format` property in the `APIRequest` class, that may still return `None`. This way, API routes can still decide which format they need: the "interpreted" format or the actual requested one.

@tomkralidis / @francbartoli What do you think?

francbartoli:
@GeoSander +1 to make the default format configurable. For example, in a couple of production instances, which are not internet-facing, having the `HTML` format doesn't make any sense at all

tomkralidis:
+1 to make `server.default_format` a configuration directive.

pvgenuchten:
Updated the PR, the config already has a server.mimetype param, which i now use as default, if not provided it defaults to html.

@GeoSander some advice on MIMETYPE[], if it is an array i can set it at api.init, and use it in apirequest, if it is a basic string var this does not work...

PR #912, opened 2022-05-24
optimise geojson provider if file does not exist

### LINKED ISSUE #908: key error when no/incorrect geojson file
on a docker 0.12.0 instance, got this error

```
  File "/pygeoapi/pygeoapi/openapi.py", line 1077, in generate
    click.echo(yaml.safe_dump(get_oas(s), default_flow_style=False))
  File "/pygeoapi/pygeoapi/openapi.py", line 1033, in get_oas
    return get_oas_30(cfg)
  File "/pygeoapi/pygeoapi/openapi.py", line 471, in get_oas_30
    coll_properties['schema']['items']['enum'] = list(p.fields.keys())
AttributeError: 'NoneType' object has no attribute 'keys'
```

seems related to a missing or bad geojson file (geojson provider)

expected: an error indicating that a geojson (which one) was not found or corrupt

pvgenuchten:
this checks if fields exist before calling keys() or items() on it, i wanted to use get('fields',{}), but the get function is already defined on provider.

this error occurs for example when the geojson file does not exist, although for this case it would be better to make the error more explicit! Should I add it in this PR?

resolves #908

### DIFF
--- a/pygeoapi/provider/geojson.py
+++ b/pygeoapi/provider/geojson.py
@@ -76,14 +76,16 @@ class GeoJSONProvider(BaseProvider):
         :returns: dict of fields
         """
 
+        fields = {}
         LOGGER.debug('Treating all columns as string types')
         if os.path.exists(self.data):
             with open(self.data) as src:
                 data = json.loads(src.read())
-            fields = {}
             for f in data['features'][0]['properties'].keys():
                 fields[f] = {'type': 'string'}
-            return fields
+        else:
+            LOGGER.warning('File {} does not exist.'.format(self.data))
+        return fields
 
     def _load(self, skip_geometry=None, properties=[], select_properties=[]):
         """Load and validate the source GeoJSON file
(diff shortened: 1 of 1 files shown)


GeoSander:
Thinking... if `p.fields` is not a `dict`, then that basically means that someone did not do a very good job on the provider implementation.
The [base provider already sets `p.fields` to `{}`](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/base.py#L63), so if this turns into `None` somehow, it must mean that the provider implementation explicitly set it to that (possibly in the `get_fields()` function override).
So the question here is: do we really want to "hide" this by this workaround?

pvgenuchten:
I agree, let me try and optimize the geojson provider on this aspect

pvgenuchten:
reverted change on api, optimised geojson provider only

GeoSander (approved):
Looks good to me! :+1:

GeoSander (approved):
Thanks for your changes!

Issue #877, opened 2022-03-30
google does not index datasets anymore


pvgenuchten:
i noticed google dataset search does not list any datasets from demo.pygeoapi.io anymore

i wonder if it is related to the fact that rich snippets test finds the json representation

I assume, in case no accept header is given, the json representation is returned by pygeoapi, in my opinion we should return the html representation if no accept header is given (to support this case)

![image](https://user-images.githubusercontent.com/299829/160828469-6c8b0741-52a6-40ba-aa61-391126068bce.png)

pvgenuchten:
The json-ld for search engine is embedded in the html response. No need to request json-ld specifically

ksonda:
I tested using this config https://github.com/ksonda/global-river-runner/blob/main/deploy/cloud/latest/pygeoapi.config.yml

(server mimetype set to text/html) and google does indeed retrieve the json response by default still.

pvgenuchten:
see #902

PR #916, opened 2022-05-30
Embed jsonld snippet directly, without requesting with ajax

pvgenuchten:
Goal is to:
- ggl rich snippets test does not parse the ajax response (not clear if rich snippets test is fully representative for google structured data crawler)
- better performance

This is a work in progress, now implemented for /collections/foo and collections/foo/items/faa

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -42,6 +42,7 @@ from gzip import compress
 import json
 import logging
 import os
+import copy
 import re
 from typing import Any, Tuple, Union
 import urllib.parse
@@ -1135,6 +1136,9 @@ class API:
         if request.format == F_HTML:  # render
             fcm['collections_path'] = self.get_collections_url()
             if dataset is not None:
+                fcm['ld'] = jsonldify_collection(self, copy.deepcopy(fcm),
+                                                 request.locale)
+                fcm['ld']['@context'] = 'https://schema.org/'
                 content = render_j2_template(self.config,
                                              'collections/collection.html',
                                              fcm, request.locale)
@@ -1971,6 +1975,12 @@ class API:
                 content['title_field'] = l10n.translate(p.title_field,
                                                         request.locale)
             content['collections_path'] = self.get_collections_url()
+
+            # embed a json-ld snippet of the item in html
+            content['ld'] = geojson2jsonld(
+                self.config, copy.deepcopy(content),
+                dataset, uri, (p.uri_field or 'id')
+            )
 
             content = render_j2_template(self.config,
                                          'collections/items/item.html',
--- a/pygeoapi/linked_data.py
+++ b/pygeoapi/linked_data.py
@@ -140,7 +140,7 @@ def jsonldify_collection(cls, collection, locale_):
         ),
         "name": l10n.translate(collection['title'], locale_),
         "description": l10n.translate(collection['description'], locale_),
-        "license": cls.fcmld['license'],
+        "license": cls.config['metadata'].get('license', {}).get('url', None),
         "keywords": l10n.translate(collection.get('keywords', None), locale_),
         "spatial": None if (not hascrs84 or not bbox) else [{
             "@type": "Place",
--- a/pygeoapi/templates/_base.html
+++ b/pygeoapi/templates/_base.html
@@ -22,6 +22,11 @@
     {% endfor %}
     {% block extrahead %}
     {% endblock %}
+    {% if data['ld'] %}
+    <script type="application/ld+json">
+      {{ data['ld'] | to_json | safe }}
+    </script>
+    {% endif %}
   </head>
   <body>
   <div class="bg-light sticky-top border-bottom">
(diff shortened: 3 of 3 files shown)


pvgenuchten:
The code uses copy.deepcopy to copy the content item, to be used for jsonld generation. This relates to the fact that the content item seems altered during jsonld conversion

I'm using metadata.license directly from config, as fcmld is not available on html api class

alpha-beta-soup:
Not sure what the search engines are doing (so whether or not this is required for structure data parsing), but the idea of using AJAX to request it separately is already to _improve performance_ by making the request for that data separate from the page load itself. That is, a human can see the page as quickly as possible, and then the search-engine oriented data component can load in the background.

pvgenuchten:
> Why not just use python machinery to embed the result of a URL request for the json-ld. 

No need to make an actual request, this PR actually does the same thing by calling the function that normally outputs the json-ld

pvgenuchten:
> One hopes a search engine is using conneg to preferentially load JSON-LD over HTML, or at least looking at rel=alternate or something.

unfortunately no, both yandex and ggl (haven't tried bing yet) seem to not use conneg or rel=alternate

pvgenuchten:
> Introducing a new way to get JSON-LD representation that is not via conneg seems a little worrying (e.g. "as fcmld is not available on html api class").

Note that this PR does not add a http mechanism to get json-ld representation, it adds json-ld content to an existing http mechanism. I do not know the consequences of "fcmld is not available on html api class"

tomkralidis:
@pvgenuchten @alpha-beta-soup @webb-ben how well is XHR supported in search engine crawlers?  This will help us determine the direction on whether to XHR or embed.

pvgenuchten:
> Why not just use python machinery to embed the result of a URL request for the json-ld. 

Yes, it would work, but I’d prefer not to add another http request, if we can manage fully on python side

webb-ben:
XHR is supported by the google crawler 
image

Think there is a bigger question at hand regarding best practice for SEO and crawlability (see https://github.com/geopython/pygeoapi/issues/877 and https://github.com/geopython/pygeoapi/pull/902). SEO is graded in part based on how long it takes for the initial page to load - to that extent the XHR implementation is beneficial. 

Separately requesting the JSON-LD via XHR request does increase the CPU processing time. This PR pumps the already computed result into the json-ld script tag.

pvgenuchten:
@webb-ben you indicate xhr is supported? however my impression is that XHR is not supported by google to ingest the specific json-ld payload embedded in html, because in that scenario i'd expect the [rich results test](https://search.google.com/test/rich-results/) to identify a schema.org/dataset on https://demo.pygeoapi.io/master/collections/obs?f=html, which it does not:

image

webb-ben:
When I try to crawl https://demo.pygeoapi.io/master/collections/obs?f=html ([ref](https://search.google.com/test/rich-results/result?id=OhZ-0xr9RR2wYRNPTOGepg)), I get that it is blocked by [robots.txt](https://demo.pygeoapi.io/robots.txt).

However, crawling https://reference.geoconnex.us/collections/nat_aq?f=html running 0.15.dev0 ([ref](https://search.google.com/test/rich-results/result?id=ssVsACYqsPOol-l84BPTkA))  and crawling https://features.internetofwater.dev/collections/NPDES?f=html running 0.16.dev0 ([ref](https://search.google.com/test/rich-results/result?id=ckCA8lJ4C4N7vZ31H0N-0Q)) appear to work as expected

webb-ben:
The Google Rich Result test as a test needs to be better clarified in this conversation. For instance, I was not able to find a version of pygeoapi landing page with any detect rich text. So I went through the OGC API feature hypermedia pattern:

| API Endpoint                   | Rich Result Test URL                                                              | Result |
|-------------------------------|----------------------------------------------------------------------------------|--------|
| `/`                           | [Test URL 1](https://search.google.com/test/rich-results/result?id=vRTFYyf_fPBrFyBvck8MGA)      | ❌     |
| `/collections`                | [Test URL 2](https://search.google.com/test/rich-results/result?id=MhfWIiFmiBksckYWJIR7vg)      | ✅    |
| `/collections/NPDES`          | [Test URL 3](https://search.google.com/test/rich-results/result?id=7uZ_gqX_LcyRou9wKr1lJA)      | ✅    |
| `/collections/NPDES/items`    | [Test URL 4](https://search.google.com/test/rich-results/result?id=CiYyUAlFnLDjXQ6m54pX_Q)      | ❌     |
| `/collections/NPDES/items/04R10I001` | [Test URL 5](https://search.google.com/test/rich-results/result?id=KsJ5c72MNfzP-LyNCbp7dw) | ❌     |

These all have valid structured data in the [schema.org validator](https://validator.schema.org/#url=https%3A%2F%2Ffeatures.internetofwater.dev%2F%3Ff%3Dhtml) that google provides a redirect to. That is in part because some of these pages do not constitute an acceptable [webpage type](https://developers.google.com/search/docs/appearance/structured-data/search-gallery) and thus is ignored by the crawler.

pvgenuchten:
Thanx @webb-ben that comment helped me to figure out what happens

apparently we have on demo 

```
Disallow: *f=json-ld*
```

so the json ld is not read by google, it explains why it works for you, not us

PR #923, opened 2022-06-10
when title-field has a multilingual configuration, use it

### LINKED ISSUE #921: title-field language annotation seems not possible
#608 describes a practice of defining a title-field in multiple languages
however when i try it, i get this artifact on the list view
image
and 'undefined' in the map view
image
if it is fixed/added, we should also update docs

pvgenuchten:
resolves #921

an example added to docs:

```
  providers:
    - type: feature
      name: GeoJSON
      title_field: 
        en: name-en 
        fr: name-fr
        de: name-de
```

?lang=en
image
?lang=it
image

### DIFF
--- a/docs/source/language.rst
+++ b/docs/source/language.rst
@@ -148,6 +148,20 @@ If you wish to make these text values available in English and French, you could
 In other words: each plain text value should be replaced by a dictionary, where the language code is the key and the translated text represents the matching value.
 For lists, this can be applied as well (see ``keywords`` example above), as long as you nest the entire list under a language key instead of each list item.
 
+A similar concept can be applied to the ``title-field`` property of the provider in a collection configuration. If a dataset contains multiple columns each representing the title 
+element in a specific language, you can configure the title-field accordingly.
+
+.. code-block:: yaml
+
+  providers:
+    - type: feature
+      name: GeoJSON
+      data: tests/data/ne_110m_lakes.geojson
+      title_field: 
+        en: name-en 
+        fr: name-fr
+        de: name-de
+
 Note that the example above uses generic language tags, but you can also supply more localized tags (with a country code) if required.
 pygeoapi should always be able find the best match to the requested language, i.e. if the user wants Swiss-French (`fr-CH`) but pygeoapi can only find `fr` tags,
 those values will be returned. However, if a `fr-CH` tag can also be found, that value will be returned and not the `fr` value.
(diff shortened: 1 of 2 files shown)


docs/source/language.rst
```diff
@@ -148,6 +148,20 @@ If you wish to make these text values available in English and French, you could
...
+.. code-block:: yaml
+
+  providers:
+    - type: feature
+      name: GeoJSON
+      data: tests/data/ne_110m_lakes.geojson
+      title_field: 
+        en: name-en 
+        fr: name-fr
+        de: name-de
```
GeoSander, line 163: Your example title-fields `name-en`, `name-fr`, and `name-de` were a bit confusing to me at first (I thought the title field had to use some locale label/construct).
Probably clearer to use `name_eng`, `name_fre`, `name_deu` or something completely different, so that it's clear that the use does not need to include language tags in the field name?

pygeoapi/api.py
```diff
@@ -1540,9 +1540,8 @@ def get_collection_items(
             if p.uri_field is not None:
                 content['uri_field'] = p.uri_field
             if p.title_field is not None:
-                content['title_field'] = p.title_field
-                content['id_field'] = p.title_field
```
GeoSander, line 1544: Haven't tested it, but are you sure you can delete `content['id_field']` just like that?

GeoSander (approved):
Good minor addition 👍. Only some minor things.

pygeoapi/api.py
pvgenuchten: @tomkralidis good comment by @geosander, i found setting the id_field awkward, because 5 lines before we already set it, it could make sense to set it, if it were still empty. I tried to git blame to find the history of the the line, but it didn't give much clarity, do you remember/know?

tomkralidis: @pvgenuchten yes, that's the logic (to override).  We use `id-field` in the template so as not to duplicate the visual in the HTML and its associated table.

pvgenuchten: sorry, i don't understand, why is the id-field overriden with the title-field if it exists? where does id-field duplicate the visual?

tomkralidis: Sorry I was thinking of a downstream application templating use case.

We override the id-field in the items template to display human readable popups when click on a feature.

pvgenuchten: So your suggestion is to restore the override for these cases, let me update the PR

GeoSander:
Looks good to me! Merging it.

PR #927, opened 2022-07-06
add support for non-advertised resources

tomkralidis:
This PR adds support for publishing non-advertised resources, per example below:

```bash
   curl https://example.org/collections  # resource _foo is not advertised
   curl https://example.org/openapi  # resource _foo is not advertised
   curl https://example.org/collections/_foo  # user can access resource normally
```


kalxas (approved):

pvgenuchten:
Hi Tom, what is the user scenario for this development (security through obscurity?)

why did you choose a convention impacting collection name in stead of adding a ‘hidden’ property on the collection

tomkralidis:
I wanted something lightweight without introducing changes to the configuration model. Having said this, we could update to something like visibility: hidden, which also may help with access control at some point. Thoughts?

francbartoli:
@tomkralidis I was wondering the same question about the user scenario. Sorry for being late in reviewing it, anyway `visibility: hidden` looks better to me.

Issue #931, opened 2022-07-06
ES docker-composition is not starting correctly


doublebyte1:
**Description**
The example [ES docker-composition](https://github.com/geopython/pygeoapi/blob/master/docker/examples/elastic/docker-compose.yml) starts with errors.

**Steps to Reproduce**
Start the [docker composition](https://github.com/geopython/pygeoapi/blob/master/docker/examples/elastic/docker-compose.yml) with `docker-compose up`

**Expected behavior**
All containers should start and pygeoapi should be available at http://localhost:5000

**Screenshots/Tracebacks**
the pygeoapi entry script shows the following error:

> pygeoapi_es       | /wait-for-elasticsearch.sh: 40: curl: not found

As it is not able to run curl, it never detects if ES is up and therefore hangs on waiting forever/ never starts the pygeoapi container.

![wait-for-elastic](https://user-images.githubusercontent.com/1038897/177588509-c75cfdbb-10b4-4a30-8be3-9fb0d4db6261.png)

**Environment**
 - OS: Ubuntu 20.04.4 LTS
 - Python version: 3.8.10
 - pygeoapi version:  0.11.0

doublebyte1:
Hopefully addressed in this PR: https://github.com/geopython/pygeoapi/pull/932

tomkralidis:
Merged - thank you!

PR #932, opened 2022-07-06
- added missing curl on pygeoapi container (which was triggering errors)

doublebyte1:
# Overview
pygeoapi [relies on curl to wait for ES](https://github.com/geopython/pygeoapi/blob/master/docker/examples/elastic/pygeoapi/wait-for-elasticsearch.sh#L40), but cur is not installed in that container. That generates an infinite waiting.

# Related Issue / Discussion
https://github.com/geopython/pygeoapi/issues/931

# Additional Information
There is another problem with the docker composition, which I will address on a separate PR.

# Contributions and Licensing

### DIFF
--- a/docker/examples/elastic/ES/Dockerfile
+++ b/docker/examples/elastic/ES/Dockerfile
@@ -31,7 +31,7 @@
 #
 # =================================================================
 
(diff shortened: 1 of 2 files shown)


tomkralidis:
Thanks @doublebyte1!

PR #934, opened 2022-07-08
update hidden resources concept

tomkralidis:
# Overview
Update of hidden resource concept.  Improves #927 

# Related Issue / Discussion
- https://github.com/geopython/pygeoapi/pull/927#issuecomment-1178394548
- https://github.com/geopython/pygeoapi/pull/927#issuecomment-1178043701
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -144,6 +144,7 @@ default.
   resources:
       obs:
           type: collection  # REQUIRED (collection, process, or stac-collection)
+          visibility: default  # OPTIONAL
           title: Observations  # title of dataset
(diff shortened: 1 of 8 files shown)


francbartoli (commented):
@tomkralidis my only question is about the returned response if the hidden collection is called. What’s the returned status code?

tomkralidis:
The HTTP status codes should be the same.  The idea is to publish a collection for demo / test purposes without having it explicitly advertised/indexed/bound to by downstream applications per se.

francbartoli (approved):

Issue #930, opened 2022-07-06
Some test data missing source information


tomkralidis:
As part of #790 and the [Project Provenance Review](https://github.com/geopython/pygeoapi/wiki/ProjectProvenanceReview), the following test data needs its source information listed in https://github.com/geopython/pygeoapi/blob/master/tests/data/README.md

- [x] https://github.com/geopython/pygeoapi/tree/master/tests/data, all `dutch_addresses_*` data (@justb4 @pvgenuchten @GeoSander @jorgejesus)
- [x] https://github.com/geopython/pygeoapi/tree/master/tests/data/tiles/ne_110m_lakes
  - @francbartoli: is this the same NE GeoJSON data we already document, but simply tiled?
- [x] https://github.com/geopython/pygeoapi/blob/master/tests/data/items.geojson @ksonda can you add some source information for this data?

cc @msmitherdc

GeoSander:
@tomkralidis I assume that the Dutch addresses data is derived from https://www.pdok.nl/downloads/-/article/adress-1, but I'm not sure.

Guess that @justb4 knows best, as he is the one who committed this! :)

francbartoli:
@tomkralidis yes it is the same dataset tiled

justb4:
Yes, I am to blame for the Addresses datasets :-). 

* Small areas were directly downloaded from a Dutch Kadaster (PDOK) GeoServer WFS (Dutch National SDI) in various formats: Shape, GPKG etc
* These are address-records derived by Kadaster from the Dutch "Buildings and Addresses" key registry (BAG) to comply with INSPIRE
* License is: CC0 1.0 https://creativecommons.org/publicdomain/zero/1.0/deed.nl
* the URL is (was) https://geodata.nationaalgeoregister.nl/inspireadressen/wfs?request=GetCapabilities&service=wfs
* this *was* the URL as the service [was switched off in June 2022](https://www.pdok.nl/-/oude-url-s-inspire-adressen-uitgefaseerd). 
* Only the WMS is still available: https://service.pdok.nl/kadaster/adressen/wms/v1_0?request=GetCapabilities&service=WMS
* the raw dataset BAG (about 2GB) can always be downloaded via the [Atom Feed](https://service.pdok.nl/kadaster/adressen/atom/v1_0/index.xml)

tomkralidis:
[...]

PR #946, opened 2022-07-27
Update README.md

tomkralidis:
# Overview
Fixes part of #930

# Related Issue / Discussion
#930
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/tests/data/README.md
+++ b/tests/data/README.md
@@ -4,11 +4,11 @@ This directory provides test data to demonstrate functionality.
 
 ## Sources
 
-### `ne_110m_lakes.geojson`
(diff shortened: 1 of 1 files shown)


tests/data/README.md
```diff
@@ -4,11 +4,11 @@ This directory provides test data to demonstrate functionality.
 
 ## Sources
 
-### `ne_110m_lakes.geojson`
+### `ne_110m_lakes.geojson / tiles/ne_110m_lakes`
```
francbartoli, line 7: what about `and` rather `/`? At first look, the whole title seems to be a path in the filesystem

tomkralidis: Good point; updated.

PR #962, opened 2022-08-15
update architecture diagram and HTML tables

tomkralidis:
# Overview
- updates architecture diagram using [C4](https://c4model.com) notation
- add borders to HTML template table output

# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing


francbartoli (changes requested):
Is the C4 model diagram for level 2 (containers) or level 3 (components)? Can we define the level explicitly? Would it make sense to have diagrams for both levels in the long term?

tomkralidis:
Thanks @francbartoli.  +1 to add more diagrams in the future.  I've added a title to the diagram to clarify the C4 level.

PR #967, opened 2022-08-27
Skip CSV row with invalid coordinates

### LINKED ISSUE #958: csv provider fails if a single x/y value is empty
image

the csv had a single row with empty x/y
what can we expect here, a try/catch around https://github.com/geopython/pygeoapi/blob/2ad45120087d50eff9a05102e456d73c8aebe138/pygeoapi/provider/csv_.py#L119, or should all geoms by valid in any data file?

JorgeMartinezG:
# Overview

When a CSV file contains invalid coordinates, skip the row completely and continue with next one. Also, a notification is added to the logs

# Related Issue / Discussion

Closes #958

# Additional Information

# Contributions and Licensing

### DIFF
--- a/pygeoapi/provider/csv_.py
+++ b/pygeoapi/provider/csv_.py
@@ -116,7 +116,8 @@ class CSVProvider(BaseProvider):
                         float(row.pop(self.geometry_y)),
                     ]
                 except ValueError:
-                    LOGGER.error(f"Skipping row with invalid coordinates id = {row.get(self.id_field)}")
+                    msg = f'Skipping row with invalid geometry: {row.get(self.id_field)}'  # noqa
+                    LOGGER.error(msg)
                     continue
                 feature = {'type': 'Feature'}
                 feature['id'] = row.pop(self.id_field)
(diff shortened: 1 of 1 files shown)


pygeoapi/provider/csv_.py
```diff
@@ -110,15 +110,20 @@ def _load(self, offset=0, limit=10, resulttype='results',
                 return feature_collection
             LOGGER.debug('Slicing CSV rows')
             for row in itertools.islice(data_, offset, offset+limit):
+                try:
+                    coordinates = [
+                        float(row.pop(self.geometry_x)),
+                        float(row.pop(self.geometry_y)),
+                    ]
+                except ValueError:
```
tomkralidis, line 118: Suggest:

```python
msg = f'Skipping row with invalid geometry: {row.get(self.id_field)}'  # noqa
LOGGER.error(msg)
continue
```

JorgeMartinezG: Included in commit [c855c7c](https://github.com/geopython/pygeoapi/pull/967/commits/c855c7c4eebdf1737bf783f4b535de7a930e40a4)

pvgenuchten (commented):
Does it make sense to skip a full row in case it contains an invalid geometry? Alternative would be to return a nill geometry with the feature (is this allowed in pygeoapi?)
one use case is that all geometries in a csv are incorrect, in this scrnario an empty collection would be created?

PR #975, opened 2022-09-01
Raise warning instead of error

### LINKED ISSUE #974: Error in the log but everything works as expected
The relevant part of my configuration file:
```
providers:
    - type: feature
      name: OGR
      data:
          source_type: GPKG
          source: filename.gpkg
      id_field: fid
```

Everything works fine but pygeoapi complains about missing `fid` column when I perform a query:

```
[2022-08-31T20:28:16Z] {/pygeoapi/pygeoapi/provider/ogr.py:477} ERROR - 'fid'
```

```
$ ogrinfo -so filename.gpkg filename | grep fid
FID Column = fid
```

What do you think about replacing `LOGGER.error(err)` with `LOGGER.warning` with probably more explicit message [here](https://github.com/geopython/pygeoapi/blob/c0c2ba541fb4b94b98fbb872915a8bfc62533ecc/pygeoapi/provider/ogr.py#L477)?

drnextgis:
# Overview

Fixes #974 

# Contributions and Licensing

- [x] I'd like to contribute to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.

### DIFF
--- a/pygeoapi/provider/ogr.py
+++ b/pygeoapi/provider/ogr.py
@@ -472,10 +472,15 @@ class OGRProvider(BaseProvider):
         if skip_geometry:
             json_feature['geometry'] = None
         try:
-            json_feature['id'] = json_feature['properties'].pop(self.id_field)
-        except Exception as err:
-            LOGGER.error(err)
-            json_feature['id'] = ogr_feature.GetFID()
+            json_feature['id'] = json_feature['properties'].pop(
+                self.id_field, json_feature['id']
+            )
+        except KeyError as err:
+            LOGGER.error(
+                "Cannot use configured id_field nor fid as id, err={}".format(
+                    err
+                )
+            )
 
         return json_feature
 


justb4:
Thanks for this PR @drnextgis . There is still some things that are obscure to me within OGR w.r.t. 'fid' or feature id's in general. e.g. there may be a way to let OGR know and set the key/unique id for features in an OGR Layer with our `self.id_field`. @rouault may enlighten us. For now I would not even log a warning (will give many-many trivial log messages), but use the KeyError as a standard Python 'trap', but still guard the whole with an Exception:

```
try: 
       try:
            json_feature['id'] = json_feature['properties'].pop(self.id_field)
       except KeyError:
            # Assume special name 'fid' configured as id-field
            json_feature['id'] = ogr_feature.GetFID()
except Exception as err:
        LOGGER.error('Cannot use configured id_field nor fid as id, err={}'.format(err))
```

drnextgis:
As I can see FID column is getting translated to JSON's 'id' property upon [ExportToJson](https://github.com/geopython/pygeoapi/blob/c0c2ba541fb4b94b98fbb872915a8bfc62533ecc/pygeoapi/provider/ogr.py#L471), [GDAL](https://github.com/OSGeo/gdal/blob/a21f8c1241c1a07dbe0b5a9241169bc2e5ecac9b/ogr/ogrsf_frmts/geojson/ogrgeojsonwriter.cpp#L707):
```
>>> from osgeo import gdal
>>> ds = gdal.OpenEx("filename.gpkg")
>>> layer = ds.GetLayer()
>>> for f in layer: feature = f
... 
>>> feature.ExportToJson(as_object=True)
{'type': 'Feature', 'geometry': {'type': 'Point', 'coordinates': [-99.216, 20.014]}, 'properties': {'geohash': '9g93y72z'}, 'id': 1}
>>> feature.GetFID()
1
```

Then the code can be simplified to:
```python
  try:
      json_feature["id"] = json_feature["properties"].pop(
          self.id_field, json_feature["id"]
      )
  except KeyError as err:
      LOGGER.error(
          "Cannot use configured id_field nor fid as id, err={}".format(err)
      )
```

justb4:
@drnextgis yes, elegant solution!  We overlooked that internal fid/id magic.

tomkralidis:
@drnextgis +1/agree with simplified update.

drnextgis:
I've updated the PR.

justb4 (changes requested):
Thanks @drnextgis . The CI error has to do with line-length:
```
/pygeoapi/provider/ogr.py:480:80: E501 line too long (82 > 79 characters)
./build/lib/pygeoapi/provider/ogr.py:480:80: E501 line too long (82 > 7[9](https://github.com/geopython/pygeoapi/runs/8203980497?check_suite_focus=true#step:15:10) characters)
```
Best is to run `flake8` at the git root. Yes, quite strict max 79 chars line-length...

drnextgis:
Fixed.

justb4:
Thanks for PR and patience @drnextgis !

Issue #957, opened 2022-08-08
if title_field config is a int property, then crumbs fail


pvgenuchten:
![image](https://user-images.githubusercontent.com/299829/183370194-d0f92757-18b9-4e35-92d0-5ca76a79b10f.png)

tested with:
ogr provider on wfs source

PR #981, opened 2022-09-06
Cast collection title to string

drnextgis:
# Related Issue / Discussion

Fixes https://github.com/geopython/pygeoapi/issues/957

# Contributions and Licensing


drnextgis:
I don't think that the error is caused by the change.

tomkralidis:
Thanks @drnextgis! Test failures unrelated.

Issue #970, opened 2022-08-27
Add Support to Elasticsearch Vector Tiles on OGC API Tiles


doublebyte1:
In recent versions, Elasticsearch is [capable of rendering data directly as vector tiles](https://www.elastic.co/blog/introducing-elasticsearch-vector-tile-search-api-for-geospatial), using a [specific endpoint](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-vector-tile-api.html).

![Screenshot from 2022-08-27 14-54-58](https://user-images.githubusercontent.com/1038897/187031127-8df2e774-3be1-445f-9e22-c4e068f8e96f.png)

This is a high-level roadmap for implementing this change:

1. Upgrade the elasticsearch version on pygeoapi to latest (8.4.0), in order to support this feature (test to make sure that it does not break anything).
2. Update the Dockerfile on examples, to use this version.
3. Add support for rendering OGC API Tiles, using this format.

doublebyte1:
I remove the tilejson metadata, since elasticsearch does not provide that

![Screenshot from 2022-08-27 20-21-18](https://user-images.githubusercontent.com/1038897/187043247-4e9d1246-e3d0-466d-b569-52791df7a6c9.png)

doublebyte1:
The elasticsearch upgrade needs to be tested before doing the PR
https://github.com/doublebyte1/pygeoapi/commit/17f7ec16fc18a70e901ae9f4e64bb636103f2fb1

doublebyte1:
[The elasticsearch client does not support ES=8](https://elasticsearch-dsl.readthedocs.io/en/latest/), but there is a workaround which [is enabling compatibility mode in ES8](https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/migration.html#migration-compat-mode).

PR #985, opened 2022-09-15
Upgrade the docker example to ES=8

doublebyte1:
# Overview
ES8 introduces new capabilities (like the ability to publish data as vector tiles), but we are tied to the elastic-search client, which did not produce a version compatible with 8, yet. See: https://github.com/elastic/elasticsearch-dsl-py/issues/1569
The approach here is to run ES8 in compatibility mode, so that it does not break the client.

# Related Issue / Discussion
https://github.com/geopython/pygeoapi/issues/970

# Contributions and Licensing

- [x] I'd like to contribute the feature to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.

### DIFF
--- a/docker/examples/elastic/docker-compose.yml
+++ b/docker/examples/elastic/docker-compose.yml
@@ -62,6 +62,10 @@ services:
     ports:
       - 9300:9300
       - 9200:9200
+    environment:
+      ELASTIC_CLIENT_APIVERSIONING: 1
+      # Enabling compatibility mode in ES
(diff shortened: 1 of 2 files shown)


docker/examples/elastic/docker-compose.yml
```diff
@@ -62,6 +62,10 @@ services:
     ports:
       - 9300:9300
       - 9200:9200
+    environment:
+      ELASTIC_CLIENT_APIVERSIONING: 1
```
doublebyte1, line 66: running ES 8 in compatibility mode

Issue #820, opened 2021-11-24
Add metadata path property for Tiles provider to config


KoalaGeo:
**Is your feature request related to a problem? Please describe.**
I'm trying using pg_tileserv for Vector Tiles, connected to pygeoapi. It's almost there, see above works , but metadata doesn't come through :
`requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://pg-tileserve-2076-master.kube-idev.bgslcdevops.test/landslides.observation_with_geometry/metadata.json` 

The metadata is at `http://pg-tileserve-2076-master.kube-idev.bgslcdevops.test/landslides.observation_with_geometry.json` but pygeoapi is expecting it at  `http://pg-tileserve-2076-master.kube-idev.bgslcdevops.test/landslides.observation_with_geometry/metadata/landslides.observation_with_geometry.json`

**Describe the solution you'd like**
From Gitter

KoalaGeo @KoalaGeo Nov 10 17:24 - https://gitter.im/geopython/pygeoapi?at=618c005038377967f4a4a2ab
Could I do some sort IF statement at https://github.com/geopython/pygeoapi/blob/8740b23e235c849145c596aac87f61814cf732a0/pygeoapi/provider/mvt.py#L74
For both URL patterns

Francesco Bartoli @francbartoli Nov 10 18:25 - https://gitter.im/geopython/pygeoapi?at=618c0eaad78911028a2f1a25
Yes @KoalaGeo, it’s likely there is a space for improvements there for example when the service is recalling a backend mvt server. Maybe it makes sense to have an additional propetry like a metadata_path in the config file more or less here - https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_master/local.config.yml#L154 

**Describe alternatives you've considered**
Amend pg_tileserv to change the path it writes the metadata.json

KoalaGeo:
Suggest making the metadata path explicit in the config:

```
providers:
    - type: tile
      name: MVT
      data: tests/data/tiles/ne_110m_lakes  # local directory tree
      # data: https://example.org/ne_110m_lakes/{z}/{x}/{y}.pbf
      metadata: https://example.org/ne_110m_lakes/metadata.json # https://example.org/ne_110m_lakes.json
      options:
          metadata_format: raw # default | tilejson
          bounds: [-7.733181,49.863063,1.763249,60.860926]
          zoom:
              min: 0
              max: 5
          schemes:
              - WebMercatorQuad
[...]

PR #986, opened 2022-09-16
- Added support for elasticsearch tiles in the mvt provider

doublebyte1:
# Overview
Version 8 of ES gives support to [publishing data as vector tiles](https://www.elastic.co/blog/introducing-elasticsearch-vector-tile-search-api-for-geospatial). This PR leverages that, to enable another vector tiles backend on the mvt provider.
Because ES does not provide a tilejson, I introduce the option of using a provider with no tile metadata.

# Related Issue / Discussion
https://github.com/geopython/pygeoapi/issues/970

# Additional Information
In order to use this feature, we need ES8, so I introduced another PR for using that in the examples: https://github.com/geopython/pygeoapi/pull/985
As I explain on that PR, ES needs to run on compatibility mode, or it won't work with the python client.

# Contributions and Licensing

### DIFF
--- a/docker/examples/elastic/pygeoapi/docker.config.yml
+++ b/docker/examples/elastic/pygeoapi/docker.config.yml
@@ -150,6 +150,20 @@ resources:
               #Note elastic_search is the docker container of ES the name is defined in the docker-compose.yml
               data: http://elastic_search:9200/ne_110m_populated_places_simple
               id_field: geonameid
+            - type: tile
+              name: MVT
+              data: http://elastic_search:9200/ne_110m_populated_places_simple/_mvt/geometry/{z}/{x}/{y}
+              options:
+                metadata_format: None # default | tilejson
+                bounds: [[-180.0, -90.0],[180.0, 90.0]]
+                zoom:
+                    min: 0
+                    max: 15
+                schemes:
+                    - WebMercatorQuad
+              format:
+                    name: elastic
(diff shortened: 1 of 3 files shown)


KoalaGeo:
@doublebyte1 can you add ES to the provider table & example config in https://github.com/geopython/pygeoapi/blob/master/docs/source/data-publishing/ogcapi-tiles.rst ?

doublebyte1:
Geopython folks chat (pti-xjmc-syg - Sep 23, 2022) 1

Maybe it's an option to refactor to this architecture

KoalaGeo:
> Geopython folks chat (pti-xjmc-syg - Sep 23, 2022) 1
> 
> Maybe it's an option to refactor to this architecture

Could also have PostGres provider like crunchy data pg_tileserv. 

Incidentally Plugging pg_tileserv into pygeoapi doesn't currently work as the metadata is not on the same path pygeoapi expects

doublebyte1:
@KoalaGeo That's why I think it would be a good idea to detach the tile metadata from the data

KoalaGeo:
> @KoalaGeo That's why I think it would be a good idea to detach the tile metadata from the data

👍👍

I raised https://github.com/geopython/pygeoapi/issues/820 previously and started https://github.com/geopython/pygeoapi/pull/821 (some good discussion points) but had to close as my python skills are limited

PR #988, opened 2022-09-23
Sqlalchemy backend

volcan01010:
# Overview

This is a reworking of the PostgreSQL provider to use SQL Alchemy.  The best way to see what it does is to read and run the tests.  As far as I can tell, all existing behaviour has been preserved.


volcan01010:
This was made to the wrong branch in error.

volcan01010:
This code will first be merged into the branch associated with #964 (CQL provision) and then it can all be merged together.

Issue #993, opened 2022-09-26
Fix failing unit tests in GH Workflows


justb4:
**Description**
Many unit tests fail in GH workflow.

**Steps to Reproduce**
Commit/push something to master, watch failing GH Action for 'Build'

**Expected behavior**
Expect to see all tests succeed.

**Screenshots/Tracebacks**
See for example:
https://github.com/geopython/pygeoapi/actions/runs/3131541093

Tried commenting out failing tests in the test code or the GH Workflow file `main.yml`.
There appeared always be a next test failing. So far identified:

Fixed by commenting out:

* tests for 'coverage' failing in `test_api.py` (commented out lines 1094-1110)
* all ESRI Provider tests failing (`test_esri_provider.py` - also locally failing), commented out in `main.yml`.

Still/currently failing:

* test_openapi.py: `get_oas()` , not sure why, but stopped here commenting-out.

**Environment**
 - OS: any
 - Python version: 3.6, 3.7
 - pygeoapi version: latest master

**Additional context**
We need the unit tests to succeed to enable Docker Build/Push since #801. Now this is blocked.

PR #995, opened 2022-09-27
fix xarray provider date handling

tomkralidis:
# Overview
Safeguards fix against newer cftime installations.  Also moves flake8 testing later on in CI.
# Related Issue / Discussion
#993 
# Additional Information

# Contributions and Licensing

### DIFF
--- a/pygeoapi/provider/xarray_.py
+++ b/pygeoapi/provider/xarray_.py
@@ -556,9 +556,15 @@ def _to_datetime_string(datetime_obj):
 
     try:
         value = np.datetime_as_string(datetime_obj)
+    except TypeError as err:
+        LOGGER.warning(err)
+        if datetime_obj.microsecond != 0:
+            value = datetime_obj.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
+        else:
+            value = datetime_obj.strftime('%Y-%m-%dT%H:%M:%SZ')
     except Exception as err:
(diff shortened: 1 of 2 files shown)


pygeoapi/provider/xarray_.py
```diff
@@ -554,11 +555,16 @@ def _to_datetime_string(datetime_obj):
     :returns: str representation of datetime
     """
 
+    LOGGER.warning(f"datetime value {datetime_obj}")
+    LOGGER.warning(f"datetime type {type(datetime_obj)}")
```
justb4, line 559: Do these LOGGER warnings need to be present? Maybe remainders from debugging? Also double quoted str.

justb4 (commented):
Would be great if this fixes the 'get_oas' error in the unit test on GH Workflow. If that is the case, happy to approve/merge.

pygeoapi/provider/xarray_.py
tomkralidis: No, they are for debugging purposes ATM. Will remove once tests pass in this PR.

tomkralidis:
@justb4 @webb-ben CI should now be fixed.  Feel free to review/approve/merge.

justb4 (approved):
Ok, looks good, let's see if this is the last test to be fixed. This should fix the `get_oas()` error as in #993 .

PR #996, opened 2022-09-27
fix for issue #993 - fix ESRI Provider tests

justb4:
# Overview
Fixes ESRI Provider tests.

# Related Issue / Discussion
#993  - concerns several tests.

Strangely the only changes were shifting 2 hours in date-time assertions. Hopefully this is not a local (timezone-)context dependent assertion. Like in The Netherlands we are 2 hours from Zulu (Z) time...

# Additional Information
Still should look into the OGC ESRI tests (still commented out but OGR may not be working in main.yml).

# Contributions and Licensing

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,7 +85,7 @@ jobs:
         pytest tests/test_csv__formatter.py
         pytest tests/test_csv__provider.py
         pytest tests/test_elasticsearch__provider.py
-        #pytest tests/test_esri_provider.py
+        pytest tests/test_esri_provider.py
         pytest tests/test_filesystem_provider.py
         pytest tests/test_geojson_provider.py
         pytest tests/test_mongo_provider.py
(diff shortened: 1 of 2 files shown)


justb4:
Hmm, think this function gives different results depending on one's local timezone setting (?):
```
    def feature_time(r):
        props = r['features'][0]['properties']
        timestamp = props[TIME_FIELD]/1000
        timestamp = datetime.fromtimestamp(timestamp)
        return timestamp.strftime(DATETIME_FORMAT)
```
@webb-ben ?

justb4:
So in my local Python 3.7.1  venv (Amsterdam TZ) the tests only succeed if I set 2 hours ahead. Think the GH Workflow will have Z timezone, so reverting, but we need a more robust fix.

justb4:
Second option, even  as it seems low-level. But it compares the actual returned attribute (guess UTC timestamps in ms, "Unix Time") values without any local formatting. 

btw I am currently focussing on at least get all Workflows "green": all tests passing and Docker build/push. When that is stable will revisit here.

justb4:
Docker build/push now ok, using UTC timestamp `assert`s now.

PR #1000, opened 2022-09-28
Move flake8 test before provider tests

KoalaGeo:
# Overview

This PR moves flake testing before the provider setup & testing, this will catch any python errors before running through the effort of setting up and testing the providers.  

PR = 
- Standalone flake lint job
  - runs on
    - All push, any branch
    - pull requests to master
    - any release
- main.yml 'build'
  -  runs on
    -  lint completing on master (so this shouls catch push, pr & releases from above logic)
- containers.yml 
  - runs on 
    - build completing

# Related Issue / Discussion

None - only noticed testing order in other work. 

# Additional Information

# Contributions and Licensing

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,26 @@
+name: Lint
+
+on:
+  push:
+    paths:
+      - '*.py'
+
+jobs:
+  flake8_py3:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7.4
+          architecture: x64
+      - name: Checkout PyTorch
+        uses: actions/checkout@master
+      - name: Install flake8
+        run: pip install flake8
+      - name: Run flake8
+        uses: suo/flake8-github-action@releases/v1
+        with:
+          checkName: 'flake8_py3'   # NOTE: this needs to be the same as the job name
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,6 +4,16 @@ on:
   push:
     paths:
       - '*.py'
+      - '!**.md'   
+  pull_request:
+    branches:
+      - master
+    paths:
+      - '*.py'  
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,36 @@
+name: Lint
+
+on:
+  push:
+    paths:
+      - '*.py'
+      - '!**.md'
```
tomkralidis, line 7: Do we need this declaration (is `*.py` enough)?

justb4 (approved):
Makes lot of sense to immediately raise the lint/flake8 flag before proceeding with the heavy-load setup. Approving.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  flake8_py3:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7.4
+          architecture: x64
+      - name: Checkout PyTorch
```
tomkralidis, line 27: Are we using PyTorch here?

tomkralidis (changes requested):

.github/workflows/lint.yml
KoalaGeo: It doesn't need to run if only ,md files are changed

.github/workflows/lint.yml
KoalaGeo: Using template from https://github.com/marketplace/actions/flake8-action

.github/workflows/lint.yml
tomkralidis: doesn't the `*.py` path mean "only when this path/pattern is matched"?

KoalaGeo: It does (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths). It'll only run when there are changes to a .py file. 

Although I can see as the build and push are dependent on this job running you would want it to run in more conditions, only not running which changes are to **.md.

KoalaGeo: Fixed in https://github.com/geopython/pygeoapi/pull/1002

PR #1010, opened 2022-10-11
Fix for spaces in postgresql table names

### LINKED ISSUE #890: Postgresql tables including spaces in their names are no more supported
**Description**
Following a recent commit of Monday April 25th 2022 (SHA-1: 15e7deabed7dbcb48a77241d43f1d1ebe62d4ed4), we've been experiencing a new issue connecting to postgresql tables which contain spaces in their names. The commit refers to a fix regarding SQL injection, which is great, but it introduced an issue with postgresql tables containing spaces in their names. Admittedly, having spaces in a database table name isn't very good practice, but Postgresql does support that.

**Steps to Reproduce**
Configure yml to connect to a postgresql table with spaces in their name. Run ```pygeoapi openapi generate [...]``` to get an invalid name syntax error message.

**Expected behavior**
The connection to the postgresql table should be working (it was working before).

**Environment**
 - OS: Windows
 - Python version: 3.9.7
 - pygeoapi version: 0.13.dev0

**Additional context**
I can make it work with spaces by replacing line 130 with this (not too fancy, but proves the issue) line of code which adds quotes around the table name
```
#self.cur.execute(query_cols, (self.table,))
self.cur.execute(query_cols, ('"' + self.table + '"',))
```

Alex-NRCan:
# Overview
Fix to support Postgresql table names with spaces in their names.

# Related Issue / Discussion
Fixes #890 


tomkralidis:
@Alex-NRCan I will revisit this PR once #964 is merged.

PR #1017, opened 2022-10-16
upgrade Elasticsearch support to 8

tomkralidis:
# Overview
This PR upgrades pygeoapi Elasticsearch support to version 8, and drops support for ES 7 and earlier.
# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing


tomkralidis:
@KoalaGeo any idea why the mainline CI did not run here?

KoalaGeo:
> @KoalaGeo any idea why the mainline CI did not run here?

'lint workflow runs on all push and PR to `master`
`build` workflow only runs on the `master` branch
`containers` only runs when `build` completes

webb-ben:
https://github.com/elastic/elasticsearch-dsl-py/issues/1569

pygeoapi/provider/elasticsearch_.py
```diff
@@ -336,13 +307,15 @@ def query(self, offset=0, limit=10, resulttype='results',
...
+                matched = len(results['hits']['hits']) + offset
+                returned = len(results['hits']['hits'])
             else:
-                results = self.es.search(index=self.index_name,
-                                         from_=offset, size=limit,
-                                         body=query)
-                results['hits']['total'] = results['hits']['total']['value']
+                es_results = self.es.search(index=self.index_name,
+                                            from_=offset, size=limit, **query)
+                results = es_results.body
```
doublebyte1, line 316: The es_results structure no longer has a 'body' element:
`results = es_results.body` -> `results = es_results`

doublebyte1, line 316: The es_results structure no longer has a 'body' element:
`results = es_results.body` -> `results = es_results`

pygeoapi/provider/elasticsearch_.py
```diff
@@ -336,13 +307,15 @@ def query(self, offset=0, limit=10, resulttype='results',
...
+                matched = es_results.body['hits']['total']['value']
```
doublebyte1, line 317: The same thing applies to this line:
`matched = es_results.body['hits']['total']['value']` -> `matched = es_results['hits']['total']['value']`

pygeoapi/provider/elasticsearch_.py
```diff
@@ -399,12 +372,12 @@ def get(self, identifier, **kwargs):
             }
 
             try:
-                result = self.es.search(index=self.index_name, body=query)
-                if len(result['hits']['hits']) == 0:
+                result = self.es.search(index=self.index_name, **query)
+                if len(result.body['hits']['hits']) == 0:
```
doublebyte1, line 376: Remove body also here:
` if len(result.body['hits']['hits']) == 0:` -> ` if len(result['hits']['hits']) == 0:`

pygeoapi/provider/elasticsearch_.py
```diff
@@ -399,12 +372,12 @@ def get(self, identifier, **kwargs):
...
                     LOGGER.error(err)
                     raise ProviderItemNotFoundError(err)
                 LOGGER.debug('Serializing feature')
-                feature_ = self.esdoc2geojson(result['hits']['hits'][0])
+                feature_ = self.esdoc2geojson(result.body['hits']['hits'][0])
```
doublebyte1, line 380: ` feature_ = self.esdoc2geojson(result.body['hits']['hits'][0])` ->  `feature_ = self.esdoc2geojson(result['hits']['hits'][0])`

doublebyte1 (commented):
1. Removing the 'body" element, mentioned in the comments should allow these PR to pass the elasticsearch provider tests.
2. If we use the latest elastic client, we no longer need to run the ES container in compatibility mode. This [block](https://github.com/geopython/pygeoapi/blob/master/docker/examples/elastic/docker-compose.yml#L65) can be removed from the docker-compose.

pygeoapi/provider/elasticsearch_.py
tomkralidis: Fixed.

pygeoapi/provider/elasticsearch_.py
tomkralidis: Fixed.

pygeoapi/provider/elasticsearch_.py
tomkralidis: Fixed.

pygeoapi/provider/elasticsearch_.py
tomkralidis: Fixed.

tomkralidis:
FYI this PR is now ready.  I am keeping as WIP to sort out the following issues with Python package dependencies:

- elasticsearch-dsl is still bound to `<8` while elasticsearch supports 8+.  Installing elasticsearch-dsl after elasticsearch will downgrade the latter to `<8`
- PostgreSQL tests fail given SQLAlchemy 2 which is installed via pygeofilter

As a result, in CI, I have done the following as workarounds AFTER the main `requirements-*txt` files are installed in `.github/workflows/main.yml`:
- upgrade `elasticsearch`
- downgrade `SQLAlchemy`

Questions:
- @francbartoli is elasticsearch-dsl something we will support long term?  See https://github.com/elastic/elasticsearch-dsl-py/issues/1569
- @KoalaGeo should we update the PostgreSQL plugin (in another PR)?

cc @kalxas

tomkralidis:
Proposing the following:
- merge PR on approval
- open an issue on elasticsearch-dsl support
- open an issue on SQLAlchemy 2 support

Thoughts @francbartoli @kalxas @doublebyte1 @webb-ben ?

tomkralidis:
Issues filed:
- https://github.com/geopython/pygeoapi/issues/1123
- https://github.com/geopython/pygeoapi/issues/1124

francbartoli (approved):
I would manage the issues separately in the tickets which have been opened

kalxas:
I agree with @francbartoli

kalxas (approved):

francbartoli:
Maybe we could cut a release before merging this, what do you think @kalxas?

webb-ben (approved):
+1 from me. 

Is it worth adding documentation to clarify the sequence to install correct package versions? or is that covered by the pending issues?

docs/source/data-publishing/ogcapi-features.rst
```diff
@@ -73,11 +73,11 @@ Elasticsearch
 ^^^^^^^^^^^^^
 
 .. note::
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
tomkralidis: Good catch. Fixed.

tomkralidis:
Thinking more, the safest bet is cutting a release prior to merging this PR.  We can determine what kind of documentation is needed at that point depending on `elasticsearch-dsl` status and our use of same.

kalxas:
ok, a new release is now done.
Also the SQLAlchemy issue is fixed on master

PR #1024, opened 2022-10-19
Virtualization of API class

Alex-NRCan:
## Overview
This pull request is presented as a work in progress. Please read these as some insights on how we, at National Resources Canada, have integrated pygeoapi to meet our business needs and requirements.
This pull request follows and includes PR #1023.

## Summary
- Adds virtualization in the `API` class to enable further customization on how pygeoapi can be leveraged by providing handles to the developers to override certain methods.
      -> Be able to reload the resources/collections on the fly or periodically without restarting pygeoapi
      -> Be able to reload the openapi swagger ui on the fly or periodically without restarting pygeoapi
      -> Be able to spatially filter the collections returned by the `/collections` endpoint.
      -> Be able to append or adjust some information in the `/collections` endpoint response before the response is sent back to the client.
- Adds support to the http POST method for the `/collections` endpoint

## Details
- Calls a new function named `load_resources` upon initialization of PyGeoAPI. This function loads the `self.config['resources']` object changing the available resources dynamically. The implementation, at the parent level, calls an overridable function named `on_load_resources` which a developer, when inheriting `API` class, can implement their own way based on the realities of their environment. By default, `on_load_resources` simply returns the `resources` object as provided, so that the default behavior of the `API` class isn't altered. At NRCan, `on_load_resources` is overriden to be able to load the resources dynamically from a database of our own. Therefore, when new collections are created/deleted/updated by other systems, we call `load_resources` to hot-reload the collections without the need to update a Yaml file and reboot the PyGeoAPI instance.

- Adds an overridable function named `on_description_filter_spatially` which is called inside the `describe_collections` function. The implementation at the parent level returns the `collections` object as-is so that the default behavior of the `API` isn't altered. At NRCan, `on_description_filter_spatially` is overriden to be able to filter the available collections in PyGeoAPI using a provided geometry. Our spatial filter implementation is dependent on our environment only (we have our own spatial index of the collections offered in PyGeoAPI). I do not know if that sort of spatial filtering on that endpoint is realistic to suggest in PyGeoAPI at large. Maybe abstracting this overridable function further (to accept CQL-like queries for instance!) might make it more interesting for other use cases?

- Adds an overridable function named `on_build_collection_finalize` which is called inside the `describe_collections function`. The implementation at the parent level does nothing so that the default behavior of the `API` isn't altered. At NRCan, `on_build_collection_finalize` is overridden to be able to attach some more attributes on each collection before the response is sent back to the client. Admittedly, this idea probably goes a little bit outside of the OGC API specifications concerning the expected json response, but for NRCan, it addresses some needs (we are adding 1 property, on the fly, per collection in the JSON response). Maybe there is another way we could achieve this too, like having something in the config as the provider level?

- Adds `save_config_pygeoapi`, `save_config_openapi` and `save_configs` functions to be able to save the `API` state back in the `PYGEOAPI_CONFIG` and `PYGEOAPI_OPENAPI` files.
 -> `save_config_pygeoapi` saves the active configuration state back in the PYGEOAPI_CONFIG file
 -> `save_config_openapi` is basically the same thing as running `pygeoapi openapi generate $PYGEOAPI_CONFIG --output-file $PYGEOAPI_OPENAPI`
 -> `save_configs` calls `save_config_pygeoapi` and `save_config_openapi` sequentially

- Adds `get_describe_collections` and `post_describe_collections` which both redirects to the previously existing `describe_collections` function which offers the additions mentioned above. The advantage of this design is that the same core logic is executed whether called from GET or POST.

## Related Comments / Questions / Suggestions:

- Maybe the whole implementation of this virtualization could be using some sort of “core-plugin”, as a 5th kind of plugin (similar to `process`, `provider`, `formatter`, etc) instead of providing virtual functions in the `API` class. That would be another architectural approach, which could be a bit harder to implement, but might align with the plugin concepts as seen in the other plugin patterns. Food for thought.
- The functions `get_collection_items` and `post_collection_items` could probably be merged into a single function with a "GET"/"POST" function parameter instead. That way the same core logic/code is applied in both http methods. The new `read_spatial_filter()` function in PR #1023 is a step in that direction. For an example of this, also see functions `get_describe_collections`, `post_describe_collections` and `describe_collections` in this pull request.

# Contributions and Licensing


pygeoapi/api.py
```diff
@@ -818,11 +1028,47 @@ def conformance(self,
...
+
+        :param request: A request object
+        :param dataset: name of collection
+
+        :returns: tuple of headers, status code, content
+        """
+
+        # Redirect to the common describe_collections method
+        # specifying the GET method was used
+        return self.describe_collections(request, dataset, "GET")
```
francbartoli, line 1047: Why isn't the method addicted to the `APIRequest`?

francbartoli (changes requested):
As a general comment, this PR needs to support the starlette and the django interfaces

pygeoapi/api.py
Alex-NRCan: This can be discussed further. My take is that `self.describe_collections` here refers to `API` logic code, reminiscent of the other get/post handlers for the other endpoints. I think it applies to leave it in `API` class rather than `APIRequest`. To me, `APIRequest` would only have utility methods to handle specifically the request parameters, not so much as the API core logic.

Alex-NRCan:
> As a general comment, this PR needs to support the starlette and the django interfaces

Absolutely agree. Can you detail where you think it might be failing in that regard?

PR #1059, opened 2022-12-05
Helm Chart for pygeoapi to deploy on Kubernetes

BnJam:
# Overview

A basic Helm Chart for deploying `pygeoapi` on Kubernetes

# Related Issue / Discussion

NA

# Additional Information

Helm chart creation and deployment examples in `helm/README.md` 

# Contributions and Licensing

- [ X] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.


helm/charts/pygeoapi/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
+1. Get the application URL by running these commands:
```
tomkralidis, line 1: Can we turn this into `README.md` as well?

tomkralidis (changes requested):

helm/charts/pygeoapi/templates/NOTES.txt
BnJam: I included some extra helper information - let me know if this is appropriate, otherwise it can be removed if we want to keep it as a strict copy to README.md from NOTES.txt

tomkralidis: We should remove `NOTES.txt` and put any relevant information in this `README.md`.

helm/README.md
```diff
@@ -0,0 +1,37 @@
+# Kubernetes `pygeoapi` Helm Chart
```
BnJam, line 1: @tomkralidis Should we move the NOTES.txt->README.md information into this README.md as it's top level to the Helm deployment? Or prefer keeping them separate?

PR #1072, opened 2022-12-14
Stay with Shapely 1.x for a while

kalxas:
# Overview

Shapely 2.0 pulls NumPy as a dependency along with many changes:
https://shapely.readthedocs.io/en/latest/release/2.x.html#version-2-0-0
I believe we should wait/test a bit before moving to this new version

# Contributions and Licensing

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ python-dateutil
 pytz
 PyYAML
 rasterio
-shapely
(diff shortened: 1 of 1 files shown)


tomkralidis:
It should also need noted that NumPy is now a hard dependency of Shapely, which increases the install footprint.

PR #1089, opened 2023-01-07
Removing Python 3.6 from CI and requirements

kalxas:
# Overview

Python 3.6 is EOL and should be dropped

# Contributions and Licensing


kalxas:
We seem to have an issue with rasterio and Python 3.8

tomkralidis (approved):

PR #1105, opened 2023-01-16
Pinned dependencies

### LINKED ISSUE #344: Reproducible build with docker
Using tools like [poetry](https://python-poetry.org/) or [pipenv](https://pipenv.readthedocs.io/en/latest/) the docker build could reach the state of a deterministic artifact beyond the problem to have all requirements pinned (`<some package>==x.y.z`)at the current version. Also, this would allow us to publish and tag them only if they pass the build on travis. 

I would go with [poetry](https://python-poetry.org/docs/cli/) because it is also a good replacement of `setup.py` and it can manage the lifecycle of publishing the package to the pypi repository.
With poetry everything happens in a `pyproject.toml`, even dependencies, so we should adopt it but this might have a high impact.

With poetry:
```Dockerfile
# Upgrade pip
RUN pip install pip --upgrade
# Install poetry:
RUN pip install poetry
# Copy only requirements to cache them in docker layer
# Install pygeoapi
COPY poetry.lock /pygeoapi/poetry.lock
COPY pyproject.toml /pygeoapi/pyproject.toml
# install dependencies from poetry.lock for reproducible builds
RUN POETRY_VIRTUALENVS_CREATE=false poetry install
```

or use tools like [dephells](https://dephell.readthedocs.io/) or [poetrify](https://github.com/kk6/poetrify) to convert from requirements.txt to pyproject.toml:
#### dephell
`dephell deps convert --from=requirements.txt --to=pyproject.toml`
#### poetrify
`poetrify generate -s requirements.txt`

With pipenv:
```shell
#generate Pipfiles from requirements.txt
pipenv install -r requirements.txt
```
In the Dockerfile:

```Dockerfile
# Upgrade pip
RUN pip install pip --upgrade
# Install pipenv
RUN pip install pipenv

# Install pygeoapi
COPY Pipfile /pygeoapi/Pipfile
COPY Pipfile.lock /pygeoapi/Pipfile.lock
# install dependencies from Pipfile.lock for reproducible builds
RUN pipenv install --system --deploy --ignore-pipfile
```

**Additional context**

frafra:
# Overview
- Replace `setup.py` and `requirements*.txt` files with `pyproject.toml`; Python dependencies kept grouped
- Pinned Python dependencies (fix #344)
- Replace packages from UbuntuGIS PPA with Python packages pinned previously
- Build Docker image using Ubuntu 22.04 instead of Ubuntu 20.04 (related:  #1093)
- Smaller Docker image (940 MB instead of 1160 MB) by using multistage builds and adding `.dockerignore` (credits to @justb4)
- Python 3.8 as requirement; Python 3.7.0 cannot satisfy certain dependencies (I figured that out using PDM); 3.7.1 or greater could be used, but some additional dev packages should be installed, since no wheel are provided for older versions
- Fast Docker image build by using parallelization and caching
- Add `gunicorn` group with `gevent`, to avoid having Docker-only dependencies

# Related Issue / Discussion
Related issues: #344, #1093

# Additional Information

- GitHub CI tests are failing
- It would be better to use Docker for GitHub CI as well, to be able to easily reproduce the CI environment locally, thus avoid having to push over and over when working on that.
- Tests image is missing

# Contributions and Licensing

- [X] I'd like to contribute to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.


.dockerignore
```diff
@@ -0,0 +1,9 @@
+.git
+.github
+.pytest_cache
+build
+dist
+**/*.zip
```
justb4, line 6: This needs to be out: .zip files are used in the test-data, i.e. Shapefile .zips for OGR tests, or maybe make an exception for those data files.

docker/entrypoint.sh
```diff
@@ -81,7 +81,7 @@ case ${entry_cmd} in
 	        	echo "Skipping: ${test_py}"
```
justb4, line 81: Also skip the SensorThingsAPI tests (needs local server). See #1093 latest version of this file.

justb4 (commented):
First comments:
* testing: you may want to consider to include `pytest` and `pyld` in standard Toml/Dockerfile builds. Then the standard Image can be used (see `entrypoint.sh` for running most Unit tests).
* (related) see `.github/workflows/containers.yml`: still assumes a separate test image to be built. No need (see first bullet). May need to inlcude changes there: just one build.
* wonder which GDAL version this results in. I guess 3.0.1 from standard Jammy?
* OGR provider tests: some fail for 2 reasons: .zip files (see my .dockerignore comment) lacking and WFS: GeoSolutions WFS URL stale, new URL:                
 'WFS:https://gs-stable.geosolutionsgroup.com/geoserver/wfs' (line 112, test_ogr_wfs_provider.py)
* really wondering what the resulting Image size is! I guess much smaller than the current 1.2GB....

.dockerignore
frafra: Thank you.
I would rather use `COPY`/`ADD` commands on specific files as much as possible and reduce the usage of `.dockerignore`. It is rather easy to end up with additional files when testing/developing locally which then gets added to the locally built image.

docker/entrypoint.sh
frafra: Thanks! Fixed as well.

frafra:
Thank you for your comments.
- I created an argument for the Dockerfile, called `groups`, so users can decide if they want to have the dev/test dependencies or not. I added the dev group as default, for the moment, given that it is required by the entrypoint.
- Fixed, thank you
- GDAL is the one from Jammy indeed, but I wonder if there could be a version mismatch between the Python package and the GDAL library
- Ok! I still have some failures, actually
- `docker inspect pygeoapi | jq .[].Size | numfmt --suffix=B --to=iec-i`: `941MiB`

.dockerignore
justb4: Yes, makes sense!

justb4:
Sorry Jammy has GDAL 3.4.1. Our current `latest` has GDAL 3.4.3, released 2022/04/22 from UbuntuGIS-Unstable. Not a huge difference, but enough for some (WFS) tests to fail. But Jammy+PPA seems problematic as well now via #1093. See [the proposed plan in issue 344](https://github.com/geopython/pygeoapi/issues/344#issuecomment-1383970863).

frafra:
This image is using libgdal30 version 3.4.1, and gdal 3.4.1, so there is no version mismatch. gdal has been pinned to 3.4.1 previously.
I will rebase this PR based on the progress made by the other PR, as agreeded.

pyproject.toml
```diff
@@ -0,0 +1,126 @@
...
+    "Development Status :: 4 - Beta",
+    "Environment :: Console",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Topic :: Scientific/Engineering :: GIS",
+]
+dependencies = [
```
justb4, line 23: Many dependencies are in twice, e.g. `pydantic`, some with and without pinning like `jinja2` and `jinja2==3.0.3`. A copy/paste issue or am I missing something?

.github/workflows/main.yml
```diff
@@ -83,51 +85,47 @@ jobs:
...
+        pdm run pytest tests/test_api.py
+        pdm run pytest tests/test_config.py
+        pdm run pytest tests/test_csv__formatter.py
+        pdm run pytest tests/test_csv__provider.py
+        pdm run pytest tests/test_elasticsearch__provider.py
+        pdm run pytest tests/test_esri_provider.py
+        pdm run pytest tests/test_filesystem_provider.py
+        pdm run pytest tests/test_geojson_provider.py
+        pdm run pytest tests/test_mongo_provider.py
+        #pdm run pytest tests/test_ogr_csv_provider.py
```
justb4, line 110: Not sure why all OGR Provider tests are commented out. Ok, they were already in current workflow.
Except for WFS Provider, these should all run. Maybe give it a try.

docker/entrypoint.sh
```diff
@@ -75,13 +75,15 @@ case ${entry_cmd} in
...
+	        tests/test_sensorthings_provider.py)
+	        ;&
 	        tests/test_postgresql_provider.py)
 			;&
 	        tests/test_mongo_provider.py)
 	        	echo "Skipping: ${test_py}"
 	        ;;
 	        *)
-	        	python3 -m pytest ${test_py}
+	        	pdm run python3 -m pytest ${test_py}
```
justb4, line 86: Do you know of a way to signal test failures to the caller, mainly the GitHub Workflow in containers.yml? Ultimately we want to run these tests in that Workflow, and not push the Docker Image on test failures. This issue is already existing. Or mainly return and catch an error exit code?

pyproject.toml
```diff
@@ -0,0 +1,126 @@
...
+    "PyYAML",
+    "rasterio",
+    "requests",
+    "shapely<2.0",
+    "tinydb",
+    "unicodecsv",
+    "gdal==3.4.1",
+]
+license = {text = "MIT"}
+requires-python = ">=3.8"
```
justb4, line 63: So we are sure that Python 3.10 and 3.11 will work? Maybe then also reflect in `main.yml` Python version matrix. Or be conservative `< 3.10`. Is 3.7 obsolete? For me hard to tell. I locally even run *Python 3.7.1 (default, Nov 17 2020, 13:42:41)* for pygeoapi with `pyenv` MacOSX.
The current latest Docker on demo runs:
*Python 3.8.10 (default, Nov 14 2022, 12:59:47)*

pyproject.toml
frafra: I think I converted the project with pdm and then imported all the requirements files again, but the basic one were there already.

.github/workflows/main.yml
frafra: I think we should invoke pytest just once on all available, making the output more readable, and making it simple to handle the tests. We can skip some of them using the `--ignore` flag.

docker/entrypoint.sh
frafra: Sure. We can run the image without pushing it, running the test, and avoid pushing it if there is a failure.

pyproject.toml
frafra: Requiring 3.7.1 or greater would make dependency pinning stuck with some quite old versions, some of which are not even provided as a pre-compiled wheel. Python 3.7 will also reach end of life in a few months, so it shouldn't be a big issue, but I can extend compatibility to that too.

`main.yml` does not reflect that, since this PR still has some failing tests and is a work in progress, but I will take care of that too.

pyproject.toml
```diff
@@ -0,0 +1,126 @@
...
+[project.urls]
+Homepage = "https://pygeoapi.io"
+
+[project.scripts]
+pygeoapi = "pygeoapi:cli"
+
+[project.optional-dependencies]
+django = [
+    "Django",
+    "pydantic",
```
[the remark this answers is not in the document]
frafra: The initial `requirements-django.txt` list pydantic. Having all the dependencies listed and grouped in the same file makes it easier to spot duplicates :)

.github/workflows/main.yml
```diff
@@ -35,21 +35,23 @@ jobs:
 
   main:
     needs: [flake8_py3]
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         include:
-          - python-version: 3.6
-          - python-version: 3.7
```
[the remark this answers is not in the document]
frafra: We can keep it; I dropped it just because it was easier, and it allows having more recent version of the pinned dependencies.

pyproject.toml
frafra: @francbartoli I think that it would be the safest call, but not necessary, probably, since the dependencies are solved taking into account the Python version requirement. If the dependencies are package properly, the risk is reduced.
We can even test all the supported major Python release using a container, actually.

frafra:
Oh, that looks way better, even if some tests are failing :)

```
============================= test session starts ==============================
platform linux -- Python 3.8.15, pytest-7.2.1, pluggy-1.0.0
rootdir: /home/runner/work/pygeoapi/pygeoapi, configfile: pytest.ini
plugins: cov-4.0.0, env-0.8.1, anyio-3.6.2
collected 261 items

tests/test_api.py ..........................F.....F...........           [ 16%]
tests/test_config.py ..                                                  [ 17%]
tests/test_csv__formatter.py .                                           [ 18%]
tests/test_csv__provider.py ...                                          [ 19%]
tests/test_elasticsearch__provider.py .........                          [ 22%]
tests/test_esri_provider.py ......                                       [ 24%]
tests/test_filesystem_provider.py .                                      [ 25%]
tests/test_geojson_provider.py ......                                    [ 27%]
tests/test_l10n.py ...........                                           [ 31%]
tests/test_mongo_provider.py FFFFFFF                                     [ 34%]
tests/test_ogr_csv_provider.py .........                                 [ 37%]
tests/test_ogr_esrijson_provider.py .......                              [ 40%]
tests/test_ogr_gpkg_provider.py ...................                      [ 47%]
tests/test_ogr_shapefile_provider.py .................                   [ 54%]
tests/test_ogr_sqlite_provider.py ..........                             [ 58%]
tests/test_ogr_wfs_provider.py .............F..F....                     [ 66%]
tests/test_openapi.py ....                                               [ 67%]
tests/test_postgresql_provider.py .................................      [ 80%]
tests/test_rasterio_provider.py .....                                    [ 82%]
tests/test_sensorthings_provider.py ...                                  [ 83%]
tests/test_socrata_provider.py .....                                     [ 85%]
tests/test_sqlite_geopackage_provider.py .............                   [ 90%]
tests/test_tinydb_catalogue_provider.py ....                             [ 91%]
tests/test_tinydb_manager_for_parallel_requests.py F                     [ 92%]
tests/test_util.py ...........                                           [ 96%]
tests/test_xarray_netcdf_provider.py ....                                [ 98%]
tests/test_xarray_zarr_provider.py .....                                 [100%]

=================================== FAILURES ===================================
```

pyproject.toml
justb4: But still I don't understand why many packages are in `dependencies[]`  twice even pinned and unpinned (viz `jinja2`. The order is also not significant, usually alphabetic, ignoring upper/lowercase. I mean like Babel, pydantic etc. we need these only once, or is the order somehow here significant?

.github/workflows/main.yml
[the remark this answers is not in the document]
justb4: If 3.7 still runs all unit tests, why not. 
From what I got from conversations here think: 3.8, 3.9 at least. Jammy has 3.10.(4) so need 3.10. And if 3.11 runs all unit tests why not? There seems to have been enormous performance improvements added in 3.11... 
For all of this: may be a huge penalty in CI/CD performance. This is typically a PSC/core devs decision...

justb4:
Ha, new insight: also with GDAL 3.4.3 the 'Utah' deegree WFS gives errors! Seen on the demo server where in a few moments the new Image from #1093 will be deployed. So will look how to fix. 

This is the 'old' image/container (Python 3.8 + GDAL 3.4.3), bashing in:

```
root@b36f1b1e75ce:/pygeoapi# python3
Python 3.8.10 (default, Nov 14 2022, 12:59:47) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from osgeo import gdal as osgeo_gdal
>>> 
>>> osgeo_gdal.VersionInfo('VERSION_NUM')
'3040300'
```

francbartoli:
Hi @frafra, what is the status of this PR? Are you still working on this?

frafra:
> Hi @frafra, what is the status of this PR? Are you still working on this?

I forgot about it actually :) I can rebase and move forward next month.

francbartoli:
Ok, thanks. Let me know if I can help somehow

PR #1113, opened 2023-01-23
support paths in item IDs

tomkralidis:
# Overview
Adds support for item identifiers with paths (i.e. `/collections/{collectionId}/items/my%2Fidentifier`)
# Related Issue / Discussion

# Additional Information

# Contributions and Licensing


francbartoli (changes requested):
@tomkralidis what do you think to explicitly add an example in the documentation? Just to avoid it's an hidden support

francbartoli (changes requested):
Is it affecting the `Django` route too?

tomkralidis:
> @tomkralidis what do you think to explicitly add an example in the documentation? Just to avoid it's an hidden support

Docs updated.

francbartoli (approved):

PR #1119, opened 2023-01-26
Edr Providers can raise Request Entity Too Large HTTP status

### LINKED ISSUE #1118: EDR provider plugins should be able to fail with Request Entity Too Large
An EDR API is allowed to return [HTTP status code 413](https://docs.ogc.org/is/19-086r5/19-086r5.html#toc31), Request Entity Too Large. For someone writing a provider plugin, it would be nice to have the ability to return this error if the backend for whatever reason cannot handle a valid request.

A new provider exception like the ones found in the base provider module `pygeoapi.provider.base` that can be raised, and which results in an HTTP status code 413.

**Additional context**
A new exception could be handled in the code [here](https://github.com/geopython/pygeoapi/blob/5e7015c2c767046af03d003a45bf4baff2518fd9/pygeoapi/api.py#L3284-L3293).

petergarnaes:
# Overview

Allows an EDR provider plugin to raise an exception that results in an [HTTP status code 413 - Request Entity Too Large](https://docs.ogc.org/is/19-086r5/19-086r5.html#toc31). 

The standard states that the response should:

> ... the implementation should return a message explaining the query limits imposed by the server implementation.

This is why the provider must initialize the provider exception with the `request_limit_explanation` argument, which will be propagated to the HTTP response.

# Related Issue / Discussion

Should resolve #1118

# Contributions and Licensing

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -68,8 +68,8 @@ from pygeoapi.plugin import load_plugin, PLUGINS
 from pygeoapi.provider.base import (
     ProviderGenericError, ProviderConnectionError, ProviderNotFoundError,
     ProviderInvalidDataError, ProviderInvalidQueryError, ProviderNoDataError,
-    ProviderQueryError, ProviderItemNotFoundError, ProviderTypeError)
-from pygeoapi.provider.base_edr import EdrProviderRequestEntityTooLarge
+    ProviderQueryError, ProviderItemNotFoundError, ProviderTypeError,
+    ProviderRequestEntityTooLarge)
 
 from pygeoapi.provider.tile import (ProviderTileNotFoundError,
                                     ProviderTileQueryError,
@@ -3631,7 +3631,7 @@ class API:
             return self.get_exception(
                 HTTPStatus.INTERNAL_SERVER_ERROR, headers, request.format,
                 'NoApplicableCode', msg)
-        except EdrProviderRequestEntityTooLarge as err:
+        except ProviderRequestEntityTooLarge as err:
             msg = 'query error (check logs)'
             return self.get_exception(
                 HTTPStatus.REQUEST_ENTITY_TOO_LARGE, headers, request.format,
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -3632,10 +3632,9 @@ class API:
                 HTTPStatus.INTERNAL_SERVER_ERROR, headers, request.format,
                 'NoApplicableCode', msg)
         except ProviderRequestEntityTooLarge as err:
-            msg = 'query error (check logs)'
             return self.get_exception(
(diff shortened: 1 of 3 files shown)


pygeoapi/api.py
```diff
@@ -69,6 +69,7 @@
     ProviderGenericError, ProviderConnectionError, ProviderNotFoundError,
     ProviderInvalidDataError, ProviderInvalidQueryError, ProviderNoDataError,
     ProviderQueryError, ProviderItemNotFoundError, ProviderTypeError)
+from pygeoapi.provider.base_edr import EdrProviderRequestEntityTooLarge
```
tomkralidis, line 72: Move to `pygeoapi.provider.base.ProviderRequestEntityTooLargeError`.

pygeoapi/provider/base_edr.py
```diff
@@ -95,3 +95,10 @@ def query(self, **kwargs):
             return getattr(self, kwargs.get('query_type'))(**kwargs)
         except AttributeError:
             raise NotImplementedError('Query not implemented!')
+
+
+class EdrProviderRequestEntityTooLarge(ProviderGenericError):
```
tomkralidis, line 100: Define in `pygeoapi.provider.base.ProviderRequestEntityTooLargeError`

pygeoapi/provider/base_edr.py
```diff
@@ -95,3 +95,10 @@ def query(self, **kwargs):
...
+    """provider request entity too large error"""
+
+    def __init__(self, request_limit_explanation: str):
+        self.request_limit_explanation = request_limit_explanation
```
tomkralidis, line 104: Do we need `request_limit_explanation`?  At runtime we would be able to get the same error message via:

```python
if 1 != 2:
    raise ProviderRequestEntityTooLargeError('Requested area too large')
```

tomkralidis (changes requested):

petergarnaes:
@tomkralidis requested changes now fixed

pygeoapi/provider/base.py
```diff
@@ -313,3 +313,8 @@ class ProviderVersionError(ProviderGenericError):
 class ProviderInvalidDataError(ProviderGenericError):
     """provider invalid data error"""
     pass
+
+
+class ProviderRequestEntityTooLarge(ProviderGenericError):
```
tomkralidis, line 318: Rename to `ProviderRequestEntityTooLargeError`

tomkralidis (changes requested):

pygeoapi/provider/base.py
petergarnaes: Ups, missed that part, sorry!

tomkralidis:
CI failures are unrelated.

tomkralidis (approved):

PR #1138, opened 2023-02-10
Fix CSV provider for a single item

webb-ben:
# Overview
The CSV Provider only searches through rows `[offset: offset+limit]` for a single item. That means the CSV provider does not work for CSVs with more than 10 rows. Perhaps the CSV provider could just be a geopandas or numpy wrapper for a CSV  file to provide the wider range of OGC API - Features query patterns - but for now this is CI testing for larger CSV files and this situation.

# Related Issue / Discussion

# Additional Information

# Contributions and Licensing


tomkralidis:
Thanks @webb-ben.  For the CSV support we will keep with minimal tooling/low barrier.

Issue #1130, opened 2023-02-03
pygeoapi not recognising the limit query parameter on the /processes path


ghobona:
**Description**

The response to the following request should only list 1 process, as constrained in the limit query parameter. However, pygeoapi appears to not recognise the limit parameter.

https://demo.pygeoapi.io/master/processes?limit=1 

**Steps to Reproduce**

Call https://demo.pygeoapi.io/master/processes?limit=1 

**Expected behavior**

The number of processes listed should not exceed the number specified by the limit parameter.

See Abstract Test A.12 of [OGC 18-062r2](https://docs.ogc.org/is/18-062r2/18-062r2.html).

**Screenshots/Tracebacks**

![Screenshot 2023-02-03 at 16 52 42](https://user-images.githubusercontent.com/18121038/216660949-c1b29d36-acf0-4af3-8410-d3b5a02ea377.png)

**Additional context**

![Screenshot 2023-02-03 at 16 51 41](https://user-images.githubusercontent.com/18121038/216661006-f5093811-1cb4-4cad-82c5-b1dcf49ef87d.png)

tomkralidis:
@ghobona this should be fixed in master branch (thanks @webb-ben).

PR #1147, opened 2023-02-16
Add limit query arg to processes

webb-ben:
# Overview
Add limit query arg to processes endpoint and update corresponding pytest.

# Related Issue / Discussion
https://github.com/geopython/pygeoapi/issues/1130

# Additional Information

# Contributions and Licensing

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -3096,11 +3096,23 @@ class API:
 
         if processes_config:
             if process is not None:
-                relevant_processes = [(process, processes_config[process])]
+                relevant_processes = [process]
             else:
-                relevant_processes = processes_config.items()
+                try:
+                    limit = int(request.params.get('limit'))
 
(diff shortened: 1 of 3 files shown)


pygeoapi/api.py
```diff
@@ -3096,11 +3096,23 @@ def describe_processes(self, request: Union[APIRequest, Any],
 
         if processes_config:
             if process is not None:
-                relevant_processes = [(process, processes_config[process])]
+                relevant_processes = [process]
             else:
-                relevant_processes = processes_config.items()
+                try:
+                    limit = int(request.params.get('limit'))
```
tomkralidis, line 3102: We should type check as well.  See pattern in https://github.com/geopython/pygeoapi/blob/master/pygeoapi/api.py#L1355-L1372

tomkralidis (changes requested):

Issue #1128, opened 2023-02-02
Support OGC API - Features - Part 2: Coordinate Reference Systems by Reference


justb4:
**Is your feature request related to a problem? Please describe.**
At the moment, pygeoapi only returns features as GeoJSON, which has a default WGS84 coordinate reference system.
This implies that we will implement support for bbox, bbox-crs and crs.
The OGC standard [OGC API - Features - Part 2: Coordinate Reference Systems by Reference](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html) describes a modular extension for that support.
(No support for JSON-FG is needed yet, which is still under development (and not part of the standard mentioned here).

**Describe the solution you'd like**
Support the facilities in pygeoapi as outlined in the above OGC standard, "Coordinate Reference Systems by Reference".

For pygeoapi queries, this means that users can specify a bounding box in a chosen projection with their request. For pygeoapi responses, this means that feature geometries will be returned in the requested CRS, where the CRS definition is set using the HTTP response header "Content-Crs". Supported CRSs can be configured and will be returned in the (Collection) metadata.
The layered architecture of pygeoapi and the already-included PROJ library required for projection transformations will make these adjustments reasonably straightforward.

We will also study existing pygeoapi developments (branches) for this Standard and the same facilities as already used in other pygeoapi API Services. Any suggestions or collaborations welcome!

**Describe alternatives you've considered**
This issue is about adhering to all facilities specified in the standard. There are really no alternatives.

**Additional context**
The context for the planned development   the Geonovum Tender invitation: [Aligning OGC-API-Features tooling with OGC-standards, the Dutch API Design Rules and INSPIRE](https://www.geonovum.nl/themas/invitation-to-tender), for which a consortium of `pygeoapi` core developers were selected to make the pygeoapi enhancements. The details of the requirements in the Tender can be found [in this PDF document](https://www.geonovum.nl/documents/invitation-to-tender-ogc-api-features-tooling-adjustment-inspirev20pdf).
The enhancement (CRS-support) as specified in this issue is one of three enhancements planned. The others will get separate pygeoapi issues assigned.

justb4:
It appears there are two versions of the "Part 2" standard
https://docs.ogc.org/is/18-058/18-058.html (from Geonovum Tender) and
https://docs.opengeospatial.org/is/18-058r1/18-058r1.html

Diff seems only "editorial" though, comparing through
https://www.copyscape.com/compare.php

[...]

PR #1149, opened 2023-02-18
WIP Ogcapi features part 2 - Support for crs query parameter

MTachon:
# Overview

This PR provides support for the *crs* query parameter, specified in the [OGC API - Features Part 2](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html). It provides an implementation on the API side, which also takes care of the *Content-Crs* header in the body of the response, and provides support for the *crs* for all providers of type 'feature'. The coordinates transformation between CRSs is implemented using the native backend for the providers `PostgreSQLProvider` and `OGRProvider`. For other providers, the transformation is taken care of by the decorator `pygeoapi.util.crs_transform`, that decorates both the `get` and the `query` methods of the providers. For the `query` method, the decorator loops through the `Features` of the returned `FeatureCollection`, and transform their coordinates before returning them to the `api.py`'s side. This implementation is not optimal performance-wise, as it loops through all `Features` once more. Even if the native backend of a given provider does not support coordinates transformation between different CRSs, an implementation that transform the `Features` while looping through them the first time (within the `query` method) would be slightly more efficient, especially if there are many of them.

# Related Issue

https://github.com/geopython/pygeoapi/issues/1128

# Additional Information

Some of the code is written twice in the `API` class, once for `get_collection_items`, and once for `get_collection_item`. I thought about writing a new method for the `API` class for this part of the code, but I see that the current design is that every method matches an end-point, and should certainly be kept that way.

# Contributions and Licensing


justb4:
Crossing implementations: I was just working, since last week on a branch for the same issue: https://github.com/justb4/pygeoapi/tree/1128-oapif-ogc-crs no PR yet, but planned. May need some coordination here...My deadline is March 16, 2023, hence was not in a hurry...

MTachon:
I started playing with pygeoapi for the last couple of weeks. We, at the Norwegian Mapping Authority, are maintaining many services that follow the older OGC Web Service Standards (WMS, WFS, WCS...). We now want to move towards more modern standards, such as these *new*  the OGC APIs to replace these services. Mainly looking at *pygeoapi* and *ldproxy* in our team. I am myself more familiar with Python than Java, so here I am.

Most of the data we distribute are in a projected/cartesian coordinate system (mainly UTM projections) with ETRS89 as datum. Transforming everything to  WGS84-based CRSs is unfortunate and not option, we really need support for other CRSs. *ldproxy* has already some support for CRS transformation and distributing data in another CRS than WGS84. We would very much like to have some services, based one the *new* OGC APIs, up and running in the coming weeks. I already started on working on this implementation when I saw your https://github.com/geopython/pygeoapi/issues/1128 in the issues' list. As I did not see any pull request I felt like continuing working on it, but I should definitely have contacted you.

Fortunately, when comparing https://github.com/justb4/pygeoapi/tree/1128-oapif-ogc-crs and this PR, I do not see much overlap regarding which parts of the code were updated. Or maybe you have some unpublished changes? We definitely need some coordination!

justb4:
@MTachon no sweat, there is much activity (PRs) around pygeoapi. That is good! To add to this, there is https://github.com/geopython/pygeoapi/pull/1023 by @Alex-NRCan as well. More challenges/coordination.

Like said I just started  https://github.com/justb4/pygeoapi/tree/1128-oapif-ogc-crs . Added only the metadata/config parts, not actual CRS/CRS-bbox handling, not even StorageCRS. The config schema isn't actually used (yet) in standard pygeoapi. And  I added the mandatory/default CRS-es in api.py. Plus I work test-driven, always writing unit tests (under `tests/`) for each facility implemented. Like #1023 are you targeting only PostgreSQL Providers?
I have little time until about beginning of march, lots of travel. But we should somehow make a best-of-three now. 

My main design issue, that you may have solved/though through: how much of the bbox-crs and esepcially CRS, i.e. Feature reprojection handling is to be performed in the generic part like `api.py` and how much in the backend Providers? For example some Providers like CSV or GeoJSON do not have native projection handling, others like PostGIS and OGR can defer CRS handling to the native backend. We don't want to duplicate functionality in each Provider. I was thinking of an implementation where the generic "api" part interrogates the Provider for CRS support, possibly comparing CRS params with StorageCRS. If the Provider supports CRS natively it will defer to the Provider, otherwise handle CRS itself. Another, typical Java-inspired, implementation is via Interceptors: the generic api-part always passes any CRS parameters to the Provider, but some Providers have an Interceptor to handle generically. In Python we may use something like @crs @crs_bbox. Thinking out loud here...

MTachon:
Yes, I have seen https://github.com/geopython/pygeoapi/pull/1023 , which is focused on supporting the *bbox-crs* parameter and will be needed for implementing the standard [OGC API - Features Part 2](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html) . However, for this PR, I think it would be nice for the pygeoapi project to give support for the *crs* parameter to all providers. Eventually, pygeoapi will need to support the *bbox-crs* parameter for all providers as well.

I have started with the PostgreSQL provider as most of our data are stored in PostGIS databases. I also started with the OGR provider as it is very flexible and can be used with multiple data sources.

When I started working on this PR, I first thought that one should find out what the *storageCrs* for the collection was, from the generic `api.py`'s side, and if the CRS specified by the *crs* parameter was different, creating an object that could transform the coordinates of Shapely objects. This transformation object, essentially an instance of `pyproj.Transformer` *combined* the `shapely.ops.transform` function (see https://shapely.readthedocs.io/en/stable/manual.html#other-transformations), could further be passed as parameter to the `get` and `query` methods of the corresponding provider. Both `pyproj` and `shapely` are already listed in the `requirements.txt` file, so no extra dependency. I felt like using shapely was a good idea for its ability to load/dump WKT/WKB geometries, and it is generally my library of choice when working with geometrical features. However, looking at several providers, too many of them do not use shapely (OGR with OGRFeature objects, GeoJSON with plain JSON objects...) and it would not be very efficient to convert every feature's geometry to a shapely geometrical object just for coordinates transforming, and dumping it back to JSON...

I then thought of extracting the CRSs information (both the *storageCRS* from the config file and the *crs* query parameter) on the `api.py` side, generating the WKT representations of the source and target CRS for the coordinates transformation of the output features' geometry, and passing these further to the right provider instances. It would then be up to the providers to implement their own transformation functions using these CRS WKT representations, and which will work with the geometry object types they "deal with". I thought WKT representations as they are one of the "preferred ways" to describe CRSs (https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems).

It is true that some of the providers can handle coordinates transformation natively (such as the PostgreSQL and OGR providers you mentioned), and we should not duplicate functionality in all providers. However, it is not always straightforward to use the native backend of these providers. Take the PostgreSQL provider for example: using `sqlachemy` and `geoalchemy2` is nice and more pythonic when interacting with a PostGIS database, but right now the entire table is queried in the `get` and `query` methods of the provider. One would need to re-write some of the code there to run SQL queries against the same database table, apply the `ST_Transform` function only on the right geometry column, and return the other columns unchanged. Whenever the coordinates transformation can be handled natively by the backend, and is straightforward, I think it should be used. The `Transform` method of `OGRFeature` instances is easy to use while iterating through the returned features on the provider's side, to name one.

By interceptors, do you refer to Python's decorators? That would indeed be more elegant, and could possibly reduce code duplication. I am not sure how easy it would be to implement. I feel like the most "efficient" way is to perform the transformation while one iterates through the features the first time, which is often a few lines in the middle of a function. A decorator could wrap the `get` and `query` providers' methods, and transform (if supported natively by the used backend) the output features' geometry before returning them back to the `api.py`'s side. But this would require to iterate through the features once more.

Alternatively, the transformation logic could be written once on the `api.py`'s side (in `get_collection_items` and `get_collection_item`), and transform the coordinates of the returned geometries stored in the JSON content, which was returned by the `get` and `query` methods of the provider. Less code (one function to transform the coordinates of the geometries stored in a JSON/`dict` object and no use of the providers' backend), but again, this would mean that we must iterate through all features once more. I always try iterate through sequences as few times as possible, but the performance boosts may not be worth it in that case.

As for https://github.com/geopython/pygeoapi/pull/1023, @Alex-NRCan , have you considered to transform the coordinates of the *bbox* on the `api.py`'s side? The providers could then use the *bbox* directly, as passed by `API` class, completely ignoring the *bbox-crs* parameter (the coordinates would already in be the right CRS), and no need to `ST_Transform`/`OGRFeature.Transform`/`pyproj` transform ... on the providers' side. No need for the `Find_SRID` function either. The [storageCRS](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html#_storage_crs) section also states: *"If all features in a spatial feature collection are stored using a particular CRS then the property storageCrs SHALL be specified in the collection object of the spatial feature collection to indicate the identifier for this storage CRS."* . The *storageCRS* can, and should, be written in the config file for each collection, if all features are stored in the same CRS, which is often the case (e.g. same column in a PostGIS database table). Then, there would be no need to infer it from the data source.

But as always, there is never only one way to do things, suggestions and ideas are more than welcome!

Alex-NRCan:
> As for #1023, @Alex-NRCan , have you considered to transform the coordinates of the _bbox_ on the `api.py`'s side? The providers could then use the _bbox_ directly, as passed by `API` class, completely ignoring the _bbox-crs_ parameter (the coordinates would already in be the right CRS), and no need to `ST_Transform`/`OGRFeature.Transform`/`pyproj` transform ... on the providers' side. No need for the `Find_SRID` function either. The [storageCRS](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html#_storage_crs) section also states: _"If all features in a spatial feature collection are stored using a particular CRS then the property storageCrs SHALL be specified in the collect_

If the crs should be specified,  as per specs, per collection, in the configuration, we can go that route instead of the Find_SRID(), indeed. I'm not opposed to it at all (I had it like this for a while too on my fork :)).

As for placing the transformation of coordinates in api.py or let them go through to the providers and let the providers handle it in their best respective methods, I'm also easy on this. I can see benefits either ways.

MTachon:
Yes, both works, we just need to decide which way to go :) @justb4 , do you have any strong opinion?

I just pushed some code changes that show an implementation (*crs* parameter only) for the OGR provider. It did not require too much work, and I guess it could be done for other providers in a similar way.

Regarding the `OGRProvider`, @justb4 , I see that the configuration file allows for a consistent coordinates transformation with its *source_srs* (which to me looks like the *storageCRS* parameter) and *target_srs* fields in the providers/data section. I understand that this provided some kind of mechanism for coordinates transformation of the stored features. But now that we want to support both the *bbox-crs* and the *crs* query parameters, should the `OGRProvider` still support the *source_srs* and *target_srs* config approach? We can easily keep supporting this functionality (*storageCRs* can replace *source_srs*) if wanted, and at the same time make the *bbox-crs* and *crs* parameters override these inward/outward transformations if specified in the query.

justb4:
Lots of ground to cover/respond. Until march 3, I am sparsely available, mostly underway in travel. So from my understanding the #1023 has focus on `bbox-crs` and #1149 (this PR) on `crs`. So each PR is partially implementing [OGC API - Features Part 2](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html). I have just opened PR #1155, also as to easily see the changes (not many at the moment). My PR title is prefixed with 'WIP' (Work In Progress) as to indicate, well, WIP. This is a convention in pygeoapi development workflows, e.g. to prevent eager merging. Once we feel the PR is ready, we cut out the WIP prefix. You can do the same on  #1023 and #1149.

I read many useful ideas and suggestions above:

* my aim is to have all changes forward-compatible using defaults for CRS and StorageCRS Collection metadata
* handling `bbox-crs` in `api.py` i.s.o. each Provider is my preference
* CRS and StorageCRS config changes in `pygeoapi-config-0.x.yml` should be done at Provider level see #1155, not the Collection Level IMHO and only for the Feature Provider, as they should only appear there only in the (Feature) response metadata.
* CRS (config) defaults according to spec should be http://www.opengis.net/def/crs/OGC/1.3/CRS84 and http://www.opengis.net/def/crs/EPSG/0/4326 not http://www.opengis.net/def/crs/OGC/1.3/CRS84h 
* OGR `source/targetCRS` was a sort of 'poor-man's' support for transforms, we may keep it for now. Some deployments are using this. 
* yes with `Interceptors` I meant `Python Decorators` but these may make implementations less transparent
* please write also tests (under `/tests`). Will take some time, but we need this to get PRs merged and helps you with any kind of refactoring, e.g. moving things from `api.py` to Providers.

Main design issue remaining is: strategy for CRS transforms: at `api.py` level, or in Providers or both? Following a 'lean' strategy is to implement in `api.py`. It is mentioned that this may have performance penalties as each Provider is already looping through all Features. But closer inspection of especially the JSON encoding learns that also in (functions called in) `api.py` again a loop through all Features is made:

* at the end of `get_collection_items()`: `return headers, HTTPStatus.OK, to_json(content, self.pretty_print)`
* `to_json()` calls `json.dumps(dict_, default=json_serial, indent=indent)` . `json_serial()` is an encoder/interceptor function in `util.py` that calls on each `obj` for local formatting.
* maybe we can do any optional CRS transforms there without a need for extra looping in `api.py`

MTachon:
First of all, thanks for the feedback and interesting discussion!

I thought having the PR with *Draft* status was enough, but I have now updated the title of the PR and prefixed it with "WIP" and will remember it for future PRs :)

*crs* and *storageCRS* at the provider level instead of at the collection level may be a better idea, especially if pygeoapi provides support for different providers within the same collection? In [pygeoapi's documentation](https://docs.pygeoapi.io/en/latest/configuration.html#resources), I read *"The providers block is a list of 1..n providers with which to operate the data on."*. I am too much of a newcomer in pygeoapi's environment to have seen any example of multiple providers for a given collection. I thought about it when making these changes, and ended up with the collection level because of the wording ("collection object") in [Storage CRS](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html#crs-discovery). Anyway, if we can have multiple providers (with different data types for example), representing the same collection, and using different data sources with a (possibly) different *storageCRS*, both *crs* and *storageCrs* should definitely be at the provider level. 

Supporting http://www.opengis.net/def/crs/EPSG/0/4326 as well in the default CRS list is fine for me. I read in the specs ([Discovery - CRS identifier list](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html#_crs_identifier_list)) that either http://www.opengis.net/def/crs/OGC/1.3/CRS84 or http://www.opengis.net/def/crs/OGC/1.3/CRS84h must be in the list, depending on whether the data is with or without (ellipsoidal) height. But from my understanding, http://www.opengis.net/def/crs/EPSG/0/4326 really is the same as http://www.opengis.net/def/crs/OGC/1.3/CRS84 with a different axis order (lat/lon instead of lon/lat). Many tools regard them as the same, but most often the easting/northing coordinate order is expected. In addition, many users are more familiar with http://www.opengis.net/def/crs/EPSG/0/4326 than with http://www.opengis.net/def/crs/OGC/1.3/CRS84, so the former should also be in the list of default CRSs. But you can take care of this part in https://github.com/geopython/pygeoapi/pull/1155 .

I will update the code for the `OGRProvider` to keep the `source_srs`/`target_srs` functionality, we do not want to break anything :)

I can also update the tests to cover code changes from this PR.

Thanks for pointing out the `to_json` call in the return statement, I have not thought about it. `to_json` is called when JSON is the request format. The `write` (if *csv* is the request format) method of the `CSVFormatter` also loops through all features once more. Not sure about the *html* request format, I will look into it.

justb4:
@MTachon yes you are right about http://www.opengis.net/def/crs/OGC/1.3/CRS84h. I overlooked.

I was a bit too optimistic on `to_json()`: the `default= json_serial`: that function is only called if `json.dumps()` cannot format the object, e.g. a `date` object. To really get callbacks one needs to override the Encoder via `cls=MyEncoder` as per the manpage: https://docs.python.org/3/library/json.html and use `**kw` for passing any Transform info. When a Provider has transformed that `**kw` be empty. Broad idea, not yet in depth investigated. True that HTML, CSV format does not follow that route.

justb4:
@MTachon Good progress! Like said, I have little time to go through all details. Below based on quick look.

See you started the 'Decorator` approach. Good! Feature Providers may now choose to optimize .

I though that just using `pyproj` would do the transforms, but guess only for Points, not for structures like Polygons etc. So Shapely needed? For bounds (bbox transforms?) I see a Transformer function [transform_bounds](https://pyproj4.github.io/pyproj/stable/api/transformer.html), but ok not part of this PR.

Yes, config at Provider level has to do with pygeoapi allowing multiple different OGC APIs (Standards) for a single Collection, e.g. both OGC API Features and -Tiles for a single Vector Collection. All CRS-config pertains only to Features, I see you have taken care.

Think an existing Unit test fails since you have added a new Collection to the test config. That test simply counts the number of collections, now 8, so change: ` assert len(collections['collections']) == 7` to ` assert len(collections['collections']) == 8`.

MTachon:
@justb4 yes, the decorator can now be used for all other providers which backend do not support natively the coordinates transformation. It should work for all functions that accept a `crs_transform_wkt` keyword-argument (`CrsTransformWkt` instance) and that return a GeoJSON-like `dict`. The decorator is, as mentioned earlier, not optimal as it loops again through all returned features, but that should not be a real performance problem for up to thousands/tens of thousands returned features...

Using the `pyproj.Transformer.transform` function is nice to transform the coordinates of single points or a list of points. When working with `shapely` geometrical objects, combining both the `pyproj.Transformer.transform` and the `shapely.ops.transform` function is more efficient. Working with `shapely`'s geometrical objects makes it easy to load/dump geometries from/to a GeoJSON-like `dict`.

Regarding the `bbox` and `bbox-crs` , @Alex-NRCan , I guess that the following function would do fine:

``` python
def transform_bbox(bbox, bbox_crs, storage_crs):
    transform_func = pyproj.Transformer.from_crs(bbox_crs, storage_crs).transform
    n_dims = len(bbox) / 2
    return tuple(transform_func(bbox[:n_dims])) + tuple(transform_func(bbox[n_dims:]))
    
```

The function treats the bounding box as two points with coordinates `(min_x, min_zy[, min_z])` and `(max_x, max_y[, max_z])`. But this functionality is more something to integrate to PR https://github.com/geopython/pygeoapi/pull/1023

Yes, I guess that why the test failed, thanks for pointing it out, I will update it.

MTachon:
@justb4 , the `get` and `query` methods of all providers other than `PostgreSQLProvider` and `OGRProvider` are now all *decorated*, and the `crs` query parameter is now supported for these as well. A per-provider implementation can still be a better solution for performance reasons, albeit less *clean* and *maintainable*...

pygeoapi/provider/postgresql.py
```diff
@@ -310,7 +317,7 @@ def _reflect_table_model(self, engine):
 
         return TableModel
 
-    def _sqlalchemy_to_feature(self, item):
+    def _sqlalchemy_to_feature(self, item, crs_transform_out):
```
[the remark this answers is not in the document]
MTachon: It could, but does not need to. If no transformation needs to be done, `crs_transform_out` is already assigned `None` (return value of the `_get_crs_transform` method if `crs_transform_wkt` is `None`). Same goes for the `crs_transform_wkt` parameter of the `_get_crs_transform` method. There, the default value is set to `None`, but it does need to, as its value is inherited from the same named parameter of either `get` or `query`.

tests/pygeoapi-test-config.yml
```diff
@@ -149,6 +149,36 @@ resources:
...
+              name: CSV
+              data: tests/data/norwegian_urban_areas.csv
+              id_field: id
+              geometry:
+                  x_field: easting
+                  y_field: northing
+              crs:
+                - http://www.opengis.net/def/crs/EPSG/0/4258
+                - http://www.opengis.net/def/crs/EPSG/0/25833
+              storage_crs: http://www.opengis.net/def/crs/EPSG/0/25833
```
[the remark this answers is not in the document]
MTachon: The [config schema](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/schemas/config/pygeoapi-config-0.x.yml) is already updated https://github.com/geopython/pygeoapi/pull/1155 , did not want to make duplications there.

pygeoapi/provider/postgresql.py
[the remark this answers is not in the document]
francbartoli: OK, then it's up to you to decide for the explicit/implicit way to go

tests/pygeoapi-test-config.yml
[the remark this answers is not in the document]
francbartoli: So if I understand well, your plan is to rebase with master after the merge of #1155. Can you confirm?

justb4:
@MTachon @francbartoli @Alex-NRCan @tomkralidis as you may know there are 3 PRs now in WIP for issue #1128. With some overlap: #1149 (by @MTachon) #1155 (by me) and #1023 (by @Alex-NRCan). 
Rougly speaking: #1155 does generic parts like config, metadata and conformance, #1149 CRS parameter and #1023 CRS-BBOX parameter (for PG Provider only and with some extra features like POST support). 

Makes an interesting integration! I now have some more time (back from road-trip Southern Spain to The Netherlands), and #1149 has progressed a lot.  Subsequently integrating into master branch may be tricky. We sort of need to be able to test "the whole". I propose to create an "integration branch" from `master` like `1128-oapif-part2`  where at least we integrate #1149 and #1155 (I have some doubt for #1023 as we (I) plan to do generic bbox-crs handling based on StorageCRS in `api.py` plus I see POST support as a separate PR), but #1023 or a follow-up may benefit from generic bbox-crs. For performance that should not make a significant difference IMO: `pyproj` or `ST_Transform`.

@tomkralidis @francbartoli would you agree with the integration-branch-approach as you (and I) have the rights? It is like a `dev` branch we used to have in other projects. Could even run on demo server.

MTachon:
The *Install and run SensorThingsAPI* job crashes when building the database. Anyone knows why?

tests/pygeoapi-test-config.yml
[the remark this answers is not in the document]
justb4: Should have answered here, see [my reply below](https://github.com/geopython/pygeoapi/pull/1149#issuecomment-1450193209).

MTachon:
> @MTachon @francbartoli @Alex-NRCan @tomkralidis as you may know there are 3 PRs now in WIP for issue #1128. With some overlap: #1149 (by @MTachon) #1155 (by me) and #1023 (by @Alex-NRCan). Rougly speaking: #1155 does generic parts like config, metadata and conformance, #1149 CRS parameter and #1023 CRS-BBOX parameter (for PG Provider only and with some extra features like POST support).
> 
> Makes an interesting integration! I now have some more time (back from road-trip Southern Spain to The Netherlands), and #1149 has progressed a lot. Subsequently integrating into master branch may be tricky. We sort of need to be able to test "the whole". I propose to create an "integration branch" from `master` like `1128-oapif-part2` where at least we integrate #1149 and #1155 (I have some doubt for #1023 as we (I) plan to do generic bbox-crs handling based on StorageCRS in `api.py` plus I see POST support as a separate PR), but #1023 or a follow-up may benefit from generic bbox-crs. For performance that should not make a significant difference IMO: `pyproj` or `ST_Transform`.
> 
> @tomkralidis @francbartoli would you agree with the integration-branch-approach as you (and I) have the rights? It is like a `dev` branch we used to have in other projects. Could even run on demo server.

Regarding the performance between transformations with `pyproj` or with PostGIS (`ST_Transform`), I do not think the end-user would notice any difference, a *bbox* has just a few points. If handling the `bbox-crs` in `api.py`, I think a function similar to the `transform_bbox` function from a previous comment above can do that.

justb4:
> The _Install and run SensorThingsAPI_ job crashes when building the database. Anyone knows why?

Looks like the build  is requesting a version that does not exist (Step 1/2 : FROM ghcr.io/baosystems/postgis:14-3.2): 14-3.2 (Postgres 14 + PostGIS 3.2. While the [Workflow file specifies PostGIS 3.3](https://github.com/baosystems/docker-postgis/blob/multiarch/.github/workflows/multiarch.yml#L26) but the error message is a bit cryptic so could be something else like a limit exceedence?.

webb-ben:
> The _Install and run SensorThingsAPI_ job crashes when building the database. Anyone knows why?

[GHCR is down?](https://www.githubstatus.com/)

Update - We have identified the issue with Packages and are working to restore service.  We will continue to keep users updated on progress.
Mar 01, 2023 - 14:07 UTC

justb4:
> > The _Install and run SensorThingsAPI_ job crashes when building the database. Anyone knows why?
> 
> [GHCR is down?](https://www.githubstatus.com/)
> 
> Update - We have identified the issue with Packages and are working to restore service. We will continue to keep users updated on progress. Mar 01, 2023 - 14:07 UTC
> 
> @justb4 this is postgres database for [FROST-Sensorthings](https://github.com/FraunhoferIOSB/FROST-Server) which runs on 14.3-2.

Yes, I understand, and 14-3.2 is indeed still available in ghcr.io (thought it was missing)and Docker Hub. It is harder to see tags in ghcr.io, but found:
docker run --rm quay.io/skopeo/stable list-tags docker://ghcr.io/baosystems/postgis
(via [StackOverflow](https://stackoverflow.com/questions/68770946/how-to-see-not-download-docker-image-from-ghcr-io). So temporary hickup like we had earlier with pygeoapi Docker builds.

webb-ben:
Seems to be back up. @MTachon I'm going to rerun action for your most recent commit

justb4:
@MTachon :
> Regarding the performance between transformations with `pyproj` or with PostGIS (`ST_Transform`), I do not think the end-user would notice any difference, a _bbox_ has just a few points. If handling the `bbox-crs` in `api.py`, I think a function similar to the `transform_bbox` function from a previous comment above can do that.

Agree, as #1155 is already implementing StorageCRS support I can add this in. Though not all Feature Collections may have specified/configured StorageCRS and some providers like (PG) can supply from internal (table) metadata I may add a Provider function `get_storage_crs()` that in the base class returns what is configured or a (WGS84) default. Specialized Providers like PG may return from internal metadata such as a PostGIS Column def. One reason/goal (for Part 2)  is to remain forward compatible if possible.

MTachon:
@webb-ben :

> Seems to be back up. @MTachon I'm going to rerun action for your most recent commit

Looks like it is working again, thanks!

justb4:
> A `get_storage_crs()` function could first check for `storage_crs` in the configuration file. If not found, try to extract it from internal metadata, if the provider supports it. If the `storage_crs` still cannot be obtained (no provider support), the function could fallback on the default (WGS84).

Ok, slight complication: in `describe_collections()` no (Feature) Providers are instantiated (with `load_plugin()`). Maybe they should as we may have more cases. I see that EDR  does this already to get field names from an instantiated EDR Provider. Don't know the performance impact, for example an ElasticSearch Provider will try to contact the remote ES Server in `_init_`.
I leave this for now: relying on a configured StorageCRS value or else the default.

justb4:
Also in `openapi.py` will use the same convention @MTachon used: 
```
OPENAPI_YAML = {
    'oapif-1': 'https://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml',  # noqa
    'oapif-2': 'http://schemas.opengis.net/ogcapi/features/part2/1.0/openapi/ogcapi-features-2.yaml', # noqa
.
```
Will make integration easier.

justb4:
@MTachon @Alex-NRCan just pushed code  for bbox-crs support in #1155. Needs more tests and doc.

justb4:
At second thought: we do not need the integration branch: if we merge PR #1155 first, then merge master into #1149 en merge PR #1149 into `master` we should have the Part 2 implementation complete. Plus we'll have a testable Docker image which can be run without significant changes (some config only) on the pygeoapi demo server. #1155 will do the "groundwork" metadata/config and also now implements CRS-BBOX so strictly #1023 is not needed. 
It makes all OGR unit tests working so, handly for testing, also CRS queries. Will still need to add documentation, thinking of a dedicated section on "CRS Support" which documents the entire Part 2 support.
@tomkralidis @MTachon what do you think? I hope to finalize #1155 tue/wed this week for final review. We then remove the "WIP" Prefix from PR title to indicate ready-for-review (and possibly ping reviewers).

It may be added later as PG optimization and POST support.

MTachon:
I think it sounds like a good idea. Merging master into https://github.com/geopython/pygeoapi/pull/1149 should be relatively straightforward, even after merging https://github.com/geopython/pygeoapi/pull/1155 into master.

Wrote a few lines in the documentation to illustrate `crs`/`storage_crs` configuration and some data access examples, but a dedicated section on "CRS support" surely sounds better to me!

pygeoapi/api.py
```diff
@@ -1429,6 +1434,57 @@ def get_collection_items(
                 HTTPStatus.INTERNAL_SERVER_ERROR, headers, request.format,
                 'NoApplicableCode', msg)
 
+        crs_transform_wkt = None
```
justb4, line 1437: The lines 1437-1487 and 2189-2238 are identical, or almost. Please make a utility function. #1155 introduces a function `util.get_supported_crs_list` that may be called within at later (merging) stage.

pygeoapi/api.py
```diff
@@ -2102,26 +2160,93 @@ def get_collection_item(self, request: Union[APIRequest, Any],
...
         # Get provider language (if any)
         prv_locale = l10n.get_plugin_locale(provider_def, request.raw_locale)
 
         try:
             LOGGER.debug(f'Fetching id {identifier}')
-            content = p.get(identifier, language=prv_locale)
+            content = p.get(
+                identifier,
+                language=prv_locale,
+                crs_transform_wkt=crs_transform_wkt,
```
justb4, line 2248: may leave out ',' .

pygeoapi/util.py
```diff
@@ -533,3 +566,155 @@ def get_envelope(coords_list: List[List[float]]) -> list:
...
+    :rtype: `callable`
+    """
+    crs_transform = pyproj.Transformer.from_crs(
+        crs_in, crs_out, always_xy=True,
+    ).transform
+    return partial(shapely.ops.transform, crs_transform)
+
+
+def crs_transform(func):
+    """Decorator that transform the geometry's/geometries' coordinates of a
```
justb4, line 652: transform --> transforms

pygeoapi/util.py
```diff
@@ -533,3 +566,155 @@ def get_envelope(coords_list: List[List[float]]) -> list:
...
+    :returns: Decorated function.
+    :rtype: `callable`
+    """
+    @functools.wraps(func)
+    def get_geojsonf(*args, **kwargs):
+        crs_transform_wkt = kwargs.get('crs_transform_wkt')
+        result = func(*args, **kwargs)
+        if crs_transform_wkt is None:
+            return result
+        else:
```
justb4, line 685: matter of taste: I usually don't use `else` as the function only proceeds when `crs_transform_wkt ` `is not None`. Saves indentation as well.

tests/test_api.py
```diff
@@ -1101,6 +1165,74 @@ def test_get_collection_item(config, api_):
...
+    rsp_headers, code, response = api_.get_collection_item(
+        req, 'norway_pop', '1015',
+    )
+
+    assert code == HTTPStatus.BAD_REQUEST
+
+    # Supported CRSs
+    storage_crs = 'http://www.opengis.net/def/crs/EPSG/0/25833'
+    crs_4258 = 'http://www.opengis.net/def/crs/EPSG/0/4258'
+    supported_crs_list = [storage_crs, crs_4258]
```
justb4, line 1191: Suggestion: also add http://www.opengis.net/def/crs/OGC/1.3/CRS84, as that CRS is default, should always be supported (IMHO according to the Standard), even if not specified in config. Also without the `crs` parameter the Features should be returned in that CRS (default standard GeoJSON).

justb4 (changes requested):
I've added line-based comments per file most minor, except to bundle the logic in get() and query() for getting the optional Transform WKT structure in a util function.

Other tests I think need to be there for more full code/test coverage:

* testing `crs` for OGR Provider in combination source/target SRS
* PostgreSQL Provider which does internal Transform

MTachon:
Thanks for the feedbacks @justb4 ! Will look into it today.

pygeoapi/api.py
MTachon: I have been using *Black* for the formatting of previous projects, and got used to the trailing comma. [Black's code style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) points out that *"Such formatting produces smaller diffs; when you add or remove an element, it’s always just one line."* But I can remove it if it makes the code style more consistent with pygeoapi's code style.

pygeoapi/api.py
MTachon: Yes, I have mentioned this redundancy in the **Additional Information** section of the PR's description. The lines covering the generation of the `supported_crs_list` is already covered by https://github.com/geopython/pygeoapi/pull/1155 's function `util.get_supported_crs_list`. In addition, it does need to be done twice (processing the `bbox-crs` parameter already needs it). I thought about remove it when merging master into https://github.com/geopython/pygeoapi/pull/1149.

For the rest however, I am not sure what's best. A utility function should take as parameter a `APIRequest` object, the `provider_def` `dict`, the `API` object itself (for error handling, when raising exceptions with `self.get_exception()`), the `headers` object (both for the `self.get_exception()` call and for setting `Content-Crs`). This utility function should return a `CrsTransformWkt` instance, or `None` if no transformation should be done.

I feel like many of these parameters are very specific to both the `API` class and the `APIRequest` to pass them to a function of the generic module `util.py`. I thought that having a dedicated method the `API` class would be a good solution (removes duplication and the need to pass the `API` instance as parameter, `self`), but then it would break the current design of the `API` class: one method for one end-point, or at least the method should be very generic and usable by many other methods, not just specific to two methods and only related to OGC API - Features...

Maybe a helper function in the same `api.py` module would be a better fit? There are already such functions (e.g. `validate_bbox`, `validate_datetime`...).

MTachon:
@justb4 :

I have made some changes, and commented some of your suggestions.

Yes, the tests should cover `crs` for the `OGRProvider` in combination source/target SRS, and `crs` for the `PostgreSQLProvider`, as they both have their own implementation. Will look into it now.

pygeoapi/api.py
justb4: Ok!

tests/test_api.py
justb4: Yes, the intention for 'Part 2' support is that it should not affect existing deployments, so configuring `crs/storage_crs` is not required,  but still published in Collection metadata.

pygeoapi/api.py
MTachon: Just pushed some changes to make both `get_collection_items` and `get_collection_item` more compact. I used your approach with two new `@staticmethod`s. I have also re-used your `get_supported_crs_list` util function from https://github.com/geopython/pygeoapi/pull/1155. Does this correspond to what you had in mind?

pygeoapi/util.py
```diff
@@ -537,3 +570,184 @@ def get_envelope(coords_list: List[List[float]]) -> list:
...
+    :type uri: str
+
+
+    :raises `CRSError`: Error raised if no CRS could be identified from the
+        URI.
+
+    :returns: `pyproj.CRS` instance matching the input URI.
+    :rtype: `pyproj.CRS`
+    """
+    uri_pattern = re.compile(
```
justb4, line 616: I made a small optimization in my version: promoted this RE to global var such that RE compilation happens only once.

justb4:
> @justb4 :
> 
> I have made some changes, and commented some of your suggestions.
> 
> Yes, the tests should cover `crs` for the `OGRProvider` in combination source/target SRS, and `crs` for the `PostgreSQLProvider`, as they both have their own implementation. Do you think the tests should cover more than checking for a correct coordinates transformation if `crs`is used? I feel like the behaviour of pygeoapi regarding the `Content-Crs` header, and `crs`/`storage_crs` configuration are already covered by the test functions `test_get_collection_items_crs` and `test_get_collection_item_crs`.

Yes, makes sense. `Content-Crs` header, and `crs`/`storage_crs`  is generic, API-level, provider independent, no need to test for each Provider.

pygeoapi/api.py
justb4: Yes, good!

justb4:
So after talk/review with @tomkralidis we decided to merge both #1155 and #1149 into a single branch [crs-features-ogc](https://github.com/geopython/pygeoapi/tree/crs-features-ogc) from which we will create a single PR, such that the entire implementation can be reviewed/adapted before merging into `master`.

pygeoapi/util.py
MTachon: Yes, I see, good idea!

MTachon:
OK. Is this OK if I merge [crs-features-ogc](https://github.com/geopython/pygeoapi/tree/crs-features-ogc) into https://github.com/geopython/pygeoapi/pull/1149 instead of rebasing https://github.com/geopython/pygeoapi/pull/1149 onto [crs-features-ogc](https://github.com/geopython/pygeoapi/tree/crs-features-ogc)? Lots of identical merge conflicts that I have to solve again and again here...

justb4:
Sorry yes, I meant merge [crs-features-ogc](https://github.com/geopython/pygeoapi/tree/crs-features-ogc) into #1149, just like we have been doing from `master`.

MTachon:
Just found out about [git rerere](https://git-scm.com/book/en/v2/Git-Tools-Rerere). But will do a merge if that's fine.

justb4:
Ok!  git is eternal learning :-). We have a testserver that runs from the [crs-features-ogc](https://github.com/geopython/pygeoapi/tree/crs-features-ogc) branch, will give link in DM.

MTachon:
Yes :) OK, thanks. I have now merged [crs-features-ogc](https://github.com/geopython/pygeoapi/tree/crs-features-ogc) into [ogcapi-features-crs](https://github.com/MTachon/pygeoapi/tree/ogcapi-features-crs) from https://github.com/geopython/pygeoapi/pull/1149.

justb4:
Yes, I see! The CI error, strange, could that be due to a Proj version used in GH Workflow? Looks a bit like: https://github.com/PDAL/PDAL/issues/3423

justb4:
Maybe using another projection may help? [EPSG:32631](https://epsg.io/32631) is quite common though. I usually use EPSG:28992 (Dutch Projection) with common projections like EPSG:4258 4326 3857. I mean we are not to test Proj itself. But then again it is hard to assess what the problem is. Another possibility is, that I merge your branch into [crs-features-ogc](https://github.com/geopython/pygeoapi/tree/crs-features-ogc) and fix from there. Let me know.

justb4 (approved):

justb4:
So I merged this PR into the branch [crs-features-ogc](https://github.com/geopython/pygeoapi/tree/crs-features-ogc) and created new PR #1174 .
See comments there why the OGR Shapefile Provider test is failing. It is not the specific EPSG code, but related to the combination of GDAL, Proj, Pyproj versions. On my local system te test succeeds...

Issue #1134, opened 2023-02-09
Support adherence to specific API design rules


GeoSander:
**Is your feature request related to a problem? Please describe.**
`pygeoapi` currently does not provide advanced configuration options to better adhere to specific API design rules that may apply.
Governmental organizations offer more and more public API's. While this is a positive development, and those individual API's aim to be developer-friendly and easy to implement, the design patterns and conventions might differ between those API's based on the technology (e.g. REST framework) that was used to build them. In an attempt to harmonize the API's for the consumer, governments or other entities may define design guidelines, which API developers have to implement wherever they can. 

More specifically, we've received a request by [Geonovum
](https://www.geonovum.nl/about-us) to make the necessary changes to `pygeoapi` so that it can adhere to the [Dutch API Design Rules (ADR)](https://publicatie.centrumvoorstandaarden.nl/api/adr). 
Geonovum has run an OGC API testbed in the past where `pygeoapi` was selected as one of the tools to be tested. The findings were published [online](https://geonovum.github.io/OAPIF-PDOK-INSPIRE/#pygeoapi-versus-inspire-requirements). As an example related to the Dutch API Design Rules, one of the things that `pygeoapi` currently does not support is [API versioning](https://publicatie.centrumvoorstandaarden.nl/api/adr/#versioning), where the major version is reflected in the API URI.

**Describe the solution you'd like**
`pygeoapi` already does a pretty good job at following global REST API conventions (e.g. OAS).
However, we would like to add more options to the existing `pygeoapi` configuration file (YAML) so that the API does a better job at adhering to local design rules, if required. _This must stay optional at all times._

The challenge will be to solve this mostly on the application side, even though some problems might better be solved on the network/proxy level (e.g. Traefik). An example is this rule, which we wish to address here:
- [ ] [API-48: Leave off trailing slashes from URI](https://publicatie.centrumvoorstandaarden.nl/api/adr/#api-48)

Related to API versioning, the most pressing issues that we wish to solve are:
- [ ] [API-20: Include the major version number in the URI](https://publicatie.centrumvoorstandaarden.nl/api/adr/#api-20)
- [ ] [API-57: Return the full version number in a response header](https://publicatie.centrumvoorstandaarden.nl/api/adr/#api-57)

The issues above obviously depend on the following rules, which would be great to implement as well:
1. [API-55: Publish a changelog for API changes between versions](https://publicatie.centrumvoorstandaarden.nl/api/adr/#api-55)
[...]

PR #1152, opened 2023-02-20
API design rule support

GeoSander:
# Overview
This PR adds an optional `api_rules` object to the `server` section of the pygeoapi configuration (YAML). For an example, please check the updated documentation [here](https://github.com/GeoSander/pygeoapi/blob/adr/docs/source/configuration.rst#server). Explanation about the available options can be found [here](https://github.com/GeoSander/pygeoapi/blob/adr/docs/source/configuration.rst#api-design-rules).

The suggested changes affect the API at the highest level (Flask, Starlette, Django). Because the current tests only test the lower level methods in `api.py` but not the routes leading towards them, I have added a [mock client](https://github.com/GeoSander/pygeoapi/blob/adr/tests/util.py#L79) for Flask  and Starlette so we can test API rule adherence or any routing-related things for that matter.

Note that I have also replaced the `@app.route()` decorators in favor of `Route()` in [starlette_app.py](https://github.com/geopython/pygeoapi/pull/1152/files#diff-22b78f4f30605039c367873f702cce35081fb85fe0590b04a9f432c0b43f0db0). This technique was deprecated anyway (it is not even documented anymore in the current Starlette docs) and it proved harder to build in the API rule logic like that.

# Related Issue / Discussion
This PR relates to issue #1134, which contains the rationale behind this change. 

# Additional Information
Since this work is part of a funded [tender](https://www.geonovum.nl/themas/invitation-to-tender) in which [Geonovum](https://www.geonovum.nl/about-us) emphasized that it was very important to them that the work would be incorporated into the core, and also because the suggested changes are optional and non-disruptive, we hope that the community is willing to approve this PR. 

# Contributions and Licensing

- [x] I'd like to contribute feature "API design rule support" to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -18,6 +18,8 @@ pygeoapi configuration contains the following core sections:
 - ``metadata``: server-wide metadata (contact, licensing, etc.)
 - ``resources``: dataset collections, processes and stac-collections offered by the server
 
+The full configuration schema with descriptions of all available properties can be found `here <https://github.com/geopython/pygeoapi/blob/master/pygeoapi/schemas/config/pygeoapi-config-0.x.yml>`_.
+
 .. note::
    `Standard YAML mechanisms <https://en.wikipedia.org/wiki/YAML#Advanced_components>`_ can be used (anchors, references, etc.) for reuse and compactness.
 
@@ -30,6 +32,7 @@ Reference
 ^^^^^^^^^^
 
 The ``server`` section provides directives on binding and high level tuning.
+Please find more information related to API design rules (the property at the bottom of the example below) :ref:`further down<API Design Rules>`.
 
 .. code-block:: yaml
 
@@ -60,6 +63,12 @@ The ``server`` section provides directives on binding and high level tuning.
         name: TinyDB  # plugin name (see pygeoapi.plugin for supported process_manager's)
         connection: /tmp/pygeoapi-process-manager.db  # connection info to store jobs (e.g. filepath)
         output_dir: /tmp/  # temporary file area for storing job results (files)
+
+    api_rules:  # optional API design rules to which pygeoapi should adhere
+        api_version: 1.2.3  # omit to use pygeoapi's software version
+        strict_slashes: true  # trailing slashes will not be allowed and result in a 404
+        url_prefix: 'v{api_major}'  # adds a /v1 prefix to all URL paths
+        version_header: X-API-Version  # add a response header of this name with the API version
 
 
 ``logging``
@@ -286,6 +295,58 @@ Examples:
    curl https://example.org/collections  # resource foo is not advertised
    curl https://example.org/openapi  # resource foo is not advertised
    curl https://example.org/collections/foo  # user can access resource normally
+
+
+API Design Rules
+----------------
+
+Some pygeoapi setups may wish to adhere to specific API design rules that apply at an organization.
+The ``api_rules`` object in the ``server`` section of the configuration can be used for this purpose.
+
+Note that the entire ``api_rules`` object is optional. No rules will be applied if the object is omitted.
+
+The following properties can be set:
+
+``api_version``
+^^^^^^^^^^^^^^^
+
+If specified, this property is a string that defines the semantic version number of the API.
+Note that this number should reflect the state of the *API data model* (request and response object structure, API endpoints, etc.)
+and does not necessarily correspond to the *software* version of pygeoapi. For example, the software could have been
+completely rewritten (which changes the software version number), but the API data model might still be the same as before.
+
+Unfortunately, pygeoapi currently does not offer a way to keep track of the API version.
+This means that you need to set (and maintain) your own version here or leave it empty or unset.
+In the latter case, the software version of pygeoapi will be used instead.
+
+``strict_slashes``
+^^^^^^^^^^^^^^^^^^
+
+Some API rules state that trailing slashes at the end of a URL are not allowed if they point to a specific resource item.
+In that case, you may wish to set this property to ``true``. Doing so will result in a ``404 Not Found`` if a user adds a ``/`` to the end of a URL.
+If omitted or ``false`` (default), it does not matter whether the user omits or adds the ``/`` to the end of the URL.
+
+``url_prefix``
+^^^^^^^^^^^^^^
+
+Set this property to include a prefix in the URL path (e.g. `https://base.com/<my_prefix>/endpoint`).
+Note that you do not need to include slashes (either at the start or the end) here: they will be added automatically.
+
+If you wish to include the API version number (depending on the `api_version`_ property) in the prefix, you can use the following variables:
+
+- ``{api_version}``: full semantic version number
+- ``{api_major}``: major version number
+- ``{api_minor}``: minor version number
+- ``{api_build}``: build number
+
+For example, if the API version is *1.2.3*, then a URL prefix template of ``v{api_major}`` will result in *v1* as the actual prefix.
+
+``version_header``
+^^^^^^^^^^^^^^^^^^
+
+Set this property to add a header to each pygeoapi response that includes the semantic API version (see `api_version`_).
+If omitted, no header will be added. Common names for this header are ``API-Version`` or ``X-API-Version``.
+Note that pygeoapi already adds a ``X-Powered-By`` header by default that includes the software version number.
 
 
 Validating the configuration
--- a/pygeoapi/config.py
+++ b/pygeoapi/config.py
@@ -40,6 +40,14 @@ LOGGER = logging.getLogger(__name__)
 THISDIR = Path(__file__).parent.resolve()
 
 
+def load_schema() -> dict:
+    """ Reads the JSON schema YAML file. """
+    schema_file = THISDIR / 'schemas' / 'config' / 'pygeoapi-config-0.x.yml'
+
+    with schema_file.open() as fh2:
+        return yaml_load(fh2)
+
+
 def validate_config(instance_dict: dict) -> bool:
     """
     Validate pygeoapi configuration against pygeoapi schema
@@ -48,14 +56,8 @@ def validate_config(instance_dict: dict) -> bool:
 
     :returns: `bool` of validation
     """
-
-    schema_file = THISDIR / 'schemas' / 'config' / 'pygeoapi-config-0.x.yml'
-
-    with schema_file.open() as fh2:
-        schema_dict = yaml_load(fh2)
-        jsonschema_validate(json.loads(to_json(instance_dict)), schema_dict)
-
-        return True
+    jsonschema_validate(json.loads(to_json(instance_dict)), load_schema())
+    return True
 
 
 @click.group()
--- a/pygeoapi/models/config.py
+++ b/pygeoapi/models/config.py
@@ -1,0 +1,103 @@
+# ****************************** -*-
+# flake8: noqa
+# =================================================================
+#
+# Authors: Sander Schaminee <sander.schaminee@geocat.net>
+#
+# Copyright (c) 2023 Sander Schaminee
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation
+# files (the "Software"), to deal in the Software without
+# restriction, including without limitation the rights to use,
+# copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following
+# conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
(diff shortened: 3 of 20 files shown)


pygeoapi/process/manager/tinydb_.py
```diff
@@ -27,7 +27,12 @@
 #
 # =================================================================
 
-import fcntl
+try:
+    import fcntl
+except ModuleNotFoundError:
+    # When on Windows, fcntl does not exist and file locking is automatic
```
GeoSander, line 33: There really was no need to include this fix in this PR, but it made my life a lot easier (yes, I'm on Windows...), because the `ModuleNotFoundError` was raised on every API test.
For quicker testing, I do not wish to spin up a Linux Docker container all the time...

pygeoapi/util.py
```diff
@@ -30,29 +30,29 @@
...
-from shapely.geometry import Polygon
+from urllib.request import urlopen
 
 import dateutil.parser
-from jinja2 import Environment, FileSystemLoader, select_autoescape
-from babel.support import Translations
 import yaml
+from babel.support import Translations
+from jinja2 import Environment, FileSystemLoader, select_autoescape
+from shapely.geometry import Polygon
```
GeoSander, line 51: I merely re-ordered the imports of built-ins and installed libs here. Looked a bit chaotic...

tests/pygeoapi-test-config-apirules.yml
```diff
@@ -0,0 +1,345 @@
+# =================================================================
+#
+# Authors: Tom Kralidis <tomkralidis@gmail.com>
+#
+# Copyright (c) 2019 Tom Kralidis
```
GeoSander, line 5: Hmm... this should probably mention my name here... 🤷

GeoSander, line 5: Hmm... I should probably mention my own name instead here... 🤷

docs/source/configuration.rst
```diff
@@ -60,6 +63,12 @@ The ``server`` section provides directives on binding and high level tuning.
         connection: /tmp/pygeoapi-process-manager.db  # connection info to store jobs (e.g. filepath)
         output_dir: /tmp/  # temporary file area for storing job results (files)
 
+    api_rules:  # optional API design rules to which pygeoapi should adhere
+        api_version: 1.2.3  # omit to use pygeoapi's software version
+        strict_slashes: true  # trailing slashes will not be allowed and result in a 404
+        url_prefix: 'v{api_major}'  # adds a /v1 prefix to all URL paths
```
tomkralidis, line 69: In theory, can this be any prefix (i.e. not starting with `v`)?

pygeoapi/config.py
```diff
@@ -40,6 +40,14 @@
 THISDIR = Path(__file__).parent.resolve()
 
 
+def load_json_schema() -> dict:
```
tomkralidis, line 43: `s/load_json_schema/load_schema`?

tomkralidis (changes requested):
See change requests.  Adding @francbartoli given the significant changes to starlette.

docs/source/configuration.rst
GeoSander: @tomkralidis Correct. This is just an example. The docs elaborate on this.

pygeoapi/config.py
GeoSander: @tomkralidis Not sure what you mean by this?

I just put the code responsible for the actual schema loading out of `validate_config` and into its own function so I can reuse it (i.e. in the tests).

GeoSander:
Thanks for looking @tomkralidis. 
As for the "significant changes to Starlette" (@francbartoli), it's really not that significant except for the changes that were also done for Flask and Django. I mostly refactored the routing to follow a more fashionable approach, so I could better fit in the actual changes I needed to do.

pygeoapi/config.py
francbartoli: @GeoSander I guess he's asking to change the function name to just `load_schema`

tests/pygeoapi-test-config-apirules.yml
```diff
@@ -0,0 +1,347 @@
...
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# =================================================================
+
+server:
+    bind:
+        host: 0.0.0.0
+        port: 5000
+    url: http://localhost:5000/
```
francbartoli, line 36: What does happen if this url has a root context, i.e. `http://localhost:5000/foo` and the `url_prefix` is set? Can you add a specific test for this?

tests/util.py
```diff
@@ -67,5 +71,43 @@ def mock_request(params: dict = None, data=None, **headers) -> Request:
...
     environ.update(headers)
     request = Request(environ)
-    request.args = ImmutableMultiDict(params.items())
+    request.args = ImmutableMultiDict(params.items())  # noqa
     return request
+
+
+@contextmanager
+def mock_client(config_file: str = 'pygeoapi-test-config.yml',
+                use_cookies: bool = True, **kwargs) -> FlaskClient:
```
francbartoli, line 80: Is it reasonable to have the same mock client and test for the other supported frameworks?

francbartoli (changes requested):

pygeoapi/config.py
GeoSander: Understood :)
Will rename it!

pygeoapi/models/config.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    strict_slashes: bool = Field(
+        False,
+        description="If False (default), URL trailing slashes are allowed. "
+                    "If True, pygeoapi will return a 404."
+    )
+
+    @staticmethod
+    def create(**rules_config) -> 'APIRules':
+        """ Returns a new APIDesignRules instance for the current API version
```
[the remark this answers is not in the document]
GeoSander: Yes, that docstring occurrence should be renamed.

tests/pygeoapi-test-config-apirules.yml
GeoSander: Should be fine, but I will (try to) add a test 👍

tests/util.py
GeoSander: I thought it was too much work for now (I'm on a budget...) so I focused on the default one. Tbh, I was actually disappointed that no one ever bothered to add mock clients for any framework. So I'd say this already is a win 😉

docs/source/configuration.rst
GeoSander: See [docs](https://github.com/geopython/pygeoapi/pull/1152/files#diff-e662b6389537d2a309e93ca2065f311d3f00156764a4835671300f6c4e95ec1dR274-R287) for what you can fill in here.

docs/source/configuration.rst
```diff
@@ -60,6 +63,12 @@ The ``server`` section provides directives on binding and high level tuning.
...
+        version_header: X-API-Version  # add a response header of this name with the API version
```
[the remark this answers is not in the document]
GeoSander: Again, this is just an example. People may configure it to anyway they like (by default it won't be configured).
In the case of the Dutch API Design Rules (which I need to adhere to for my client Geonovum), they actually need it to be `API-Version`, but I preferred the `X-...` prefix here, which is what pygeoapi also does for the `X-Powered-By` header.

GeoSander:
Thanks for reviewing @francbartoli!

tests/util.py
GeoSander: Thinking.. perhaps I misunderstood you. You mean use the mock FlaskClient to also test Django and Starlette? For Starlette it might work (since that also uses Werkzeug, from which the FlaskClient inherits its mock client), for Django I'm not so sure.
We would also have to make any supported framework a dependency/requirement for the test and install that framework prior to running the tests.

francbartoli: Well, actually I mean separate StarletteClient and DjangoClient

GeoSander: Ah okay, so I did understand you correctly the first time 🙂 
I can see what I can do in the time available, focusing on Starlette first. 
Are these mock clients for all frameworks a "deal breaker" for you?

GeoSander: Ok, I'll have a look later this week!

tests/pygeoapi-test-config-apirules.yml
GeoSander: @francbartoli I'm having trouble testing this. The mock client ignores the `url` setting in the YAML config completely.

In fact, if I set `url` to `http://localhost:5000/api` for example and spin up a Docker container, pygeoapi _does_ start successfully, but it actually runs at http:/localhost:5000 instead. All the links however do have the `/api` suffix, but if I click any of the links, none of them work. I have to remove the suffix to make them work.

The `url` property is only used in api.py and other code to create absolute URLs (which is not very dynamic actually...). Aside: I actually realize that I need to fix all those URL's too if a prefix applies...

**Q:** have **you** ever successfully set a root context using _only_ the `url` property?

In the case of Flask, I could make use of the built-in [`url_for`](https://flask.palletsprojects.com/en/2.2.x/api/?highlight=url_for#flask.url_for) function, which resolves the URL correctly, provided that I set the [`SERVER_NAME`](https://flask.palletsprojects.com/en/2.2.x/config/#SERVER_NAME) and [`APPLICATION_ROOT`](https://flask.palletsprojects.com/en/2.2.x/config/#APPLICATION_ROOT) correctly in the mock client, according to what's defined in pygeoapi's configuration.

requirements-dev.txt
```diff
@@ -9,7 +9,11 @@ pytest
 pytest-cov
 pytest-env
 coverage
-pyld
```
francbartoli, line 12: I guess this dependency is used in `tests/test_api.py`, the test is failing indeed https://github.com/geopython/pygeoapi/actions/runs/4396701156/jobs/7699332233

francbartoli (changes requested):

requirements-dev.txt
GeoSander: Yes, I saw that it failed. Did not delete that dependency though. Or at least I wasn't aware that I did...
Anyway, will fix it of course 👍

pygeoapi/util.py
```diff
@@ -135,6 +134,23 @@ class EnvVarLoader(yaml.SafeLoader):
...
+
+    :param config:  Current pygeoapi configuration (dictionary).
+    :returns:       An APIRules instance.
+    """
+    rules = config['server'].get('api_rules') or {}
+    rules.setdefault('api_version', __version__)
+    return config_models.APIRules.create(**rules)
+
+
+def get_base_url(config: dict) -> str:
```
GeoSander, line 148: Added this function so we can get the complete base URL (with optional URL prefix) everywhere using the same code.

pygeoapi/util.py
```diff
@@ -519,7 +534,7 @@ def url_join(*parts: list) -> str:
     :returns: str of resulting URL
     """
 
-    return '/'.join([p.strip().strip('/') for p in parts])
+    return '/'.join([p.strip().strip('/') for p in parts]).rstrip('/')
```
GeoSander, line 537: While writing and testing `get_base_url()` above, I noticed that this function returned paths with trailing slashes if you are concatenating empty parts or parts that strip down to nothing.
The `rstrip('/')` should prevent this.

GeoSander:
Hi @tomkralidis / @francbartoli, I added the final changes (Starlette mock client, more tests) as per your request.

For Starlette, I ran into an interesting problem. Flask has a `strict_slashes=True` setting, which will **prevent** it from redirecting (308) to the URL without a trailing slash if the user added one and throws a 404 instead. Exactly what I needed. Starlette however has the `redirect_slashes` setting, but that only prevents the redirection (307) from happening and does not enforce strictness: adding a trailing slash will simply result in a 200 in that case, and in some cases [it still redirected anyway](https://github.com/encode/starlette/issues/869)...

So I have now added a [middleware class](https://github.com/geopython/pygeoapi/pull/1152/files#diff-22b78f4f30605039c367873f702cce35081fb85fe0590b04a9f432c0b43f0db0R472-R501) that tackles this problem. Note that the middleware is only installed when the `strict_slashes` API rule applies.

GeoSander:
Changed behavior of landing page requests according to OGC spec [here](https://docs.opengeospatial.org/is/17-069r4/17-069r4.html#_api_landing_page) if the `strict_slashes` API rule is in effect. For the landing page only, having a trailing slash is actually preferred (or mandatory even). 
Note that this actually contradicts the Dutch API Design Rules as discussed [here](https://gitlab.com/commonground/don/adr-validator/-/issues/5), which simply states that trailing slashes are never allowed.

GeoSander:
Update: rebased again to incorporate latests changes to `collection_items` route in starlette_app.py (added `OPTIONS` HTTP method).

PR #1155, opened 2023-02-22
Full implementation of OGC API - Features Part 2 (CRS and CRS-BBOX)

justb4:
# Overview
Full implementation [OGC API - Features Part 2](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html).

# Related Issue / Discussion
#1128 and PR #1149

# Additional Information

Full implementation of [OGC API - Features Part 2](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html) entails:

* metadata support: conformance clause in Landing Page, CRS and StorageCRS in Collection Metadata
* support  CRS and StorageCRS in pygeoapi config
* request handling
* documentation
* unit tests
* official CITE tests
* demo service on https://github.com/geopython/demo.pygeoapi.io (separate issue+PR there)

And possibly more. This PR will basically follow [OGC API - Features Part 2](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html) more or less line-by-line, adapting pygeoapi until conformant to the spec.

Two other PRs are doing similar work, we are currently discussing and  coordinating:

* #1023 with focus on `bbox-crs` in PostgreSQL Provider 
* #1149 (this PR) with focus on `crs` parameter (not `bbox-crs`).

Our coordinated strategy is to first merge this PR #1155 and then #1149.  These 2 PRs together will completely implement the "Part 2" support. 
   
# Contributions and Licensing


pygeoapi/api.py
```diff
@@ -1429,6 +1452,39 @@ def get_collection_items(
...
+                    HTTPStatus.BAD_REQUEST, headers, request.format,
+                    'NoApplicableCode', msg)
+
+            if len(bbox_crs) == 0:
+                msg = 'bbox-crs specified but is empty'
+                return self.get_exception(
+                    HTTPStatus.BAD_REQUEST, headers, request.format,
+                    'NoApplicableCode', msg)
+
+            supported_crs_list = provider_def.get('crs', DEFAULT_CRS_LIST)
```
MTachon, line 1470: I think that the `supported_crs_list` should include all *CRSs* from `DEFAULT_CRS_LIST`, even if they are not include in the config for the provider (`crs` field). I have included them in https://github.com/geopython/pygeoapi/pull/1149 with:

```python

supported_crs_list = provider_def.get('crs', list())
for uri in DEFAULT_CRS:
    if uri not in supported_crs_list:
        supported_crs_list.append(uri)
```

But as this list of supported CRSs is used for processing both the `bbox-crs` and `crs` parameter, maybe we should move these lines up and out of these `if` *statetements* for re-use?

justb4: A left-over is also to deal with http://www.opengis.net/def/crs/OGC/0/CRS84h (for coordinates with ellipsoidal height). The better to have a single check/append.

MTachon: Yes. According to the standards, the default CRS is either [http://www.opengis.net/def/crs/OGC/0/CRS84](http://www.opengis.net/def/crs/OGC/0/CRS84) (coordinates without height) or [http://www.opengis.net/def/crs/OGC/0/CRS84h](http://www.opengis.net/def/crs/OGC/0/CRS84h) (coordinates with ellipsoidal height). But to know that, one should get at least one `item`/`Feature` from the collection, and see if it has 2D/3D coordinates... Is sending one request through `get_collection_item` too expensive (maybe with `&limit=1`)? The returned item must also have a non-`NULL` geometry...

justb4: Yes that is expensive, at least cumbersome. My suggestion is that in that case one should configure explicitly.

Ha: It appears there are two versions of the "Part 2" standard
https://docs.ogc.org/is/18-058/18-058.html and
https://docs.opengeospatial.org/is/18-058r1/18-058r1.html

Diff seems only "editorial" though, comparing through
https://www.copyscape.com/compare.php

Version 18-058r1 van 2022-05-11 is listed on https://www.ogc.org/standards/ so that is the "official one".

Maybe I was looking at intermediate versions as I thought the default CRSs for the Collection should be:
    http://www.opengis.net/def/crs/EPSG/0/4326
and one of
    http://www.opengis.net/def/crs/OGC/1.3/CRS84 or
    http://www.opengis.net/def/crs/OGC/1.3/CRS84h

But can't find that (http://www.opengis.net/def/crs/EPSG/0/4326) now, looks like the default CRS 'list' is:
one of
    http://www.opengis.net/def/crs/OGC/1.3/CRS84 or
    http://www.opengis.net/def/crs/OGC/1.3/CRS84h

Which makes sense as these have lon, lat axis ordering like GeoJSON default requires..

MTachon: Yes, http://www.opengis.net/def/crs/EPSG/0/4326 is not part of the default CRSs in the standard.

Something I have also thought about, is how pygeoapi handles the coordinates of the `bbox` query parameter. In the part 1 of the standard, [requirement 23 B](https://docs.ogc.org/is/17-069r4/17-069r4.html#_parameter_bbox) use the wording *"coordinate axis1, coordinate axis 2"*. Axis 1 for http://www.opengis.net/def/crs/EPSG/0/4326 is for latitude, while the Axis 1 is longitude for http://www.opengis.net/def/crs/OGC/1.3/CRS84 . I think pygeoapi handles a `bbox` as `min_x, min_y[, min_z], max_x, max_y[, max_z]` even if `bbox-crs` will be set to http://www.opengis.net/def/crs/EPSG/0/4326

justb4: Cannot tell if/how pygeoapi  bbox-handling affects axis ordering. At least it checks if `min_x < max_x` etc, but that is not affected by ordering IMHO. I'll be writing more tests, also with http://www.opengis.net/def/crs/EPSG/0/4326, so that may come out.

I have just changed the "default CRS" handling what I think is according to the Part 2 standard. Having the util-function came in handy!

MTachon: Well, maybe not when handling the `bbox` itself. Just wondering how pygeoapi deals with the different providers (and data sources) that may use different conventions regarding the axis order lat/lon or lon/lat.

Nice work!

pygeoapi/schemas/config/pygeoapi-config-0.x.yml
```diff
@@ -429,6 +429,36 @@ properties:
...
                                           minItems: 1
                                           uniqueItems: true
+                                      crs:
+                                           type: array
+                                           description: |-
+                                               supported coordinate reference systems (CRSs).
+                                               pygeoapi will always provide or add the defaults if not specified.
+                                           items:
+                                               type: string
+                                               format: uri
```
francbartoli, line 439: Does it make sense to add an `enum` property to validate the enumeration of allowed values? Something like

```yaml
items:
    type: string
    format: uri
    enums:
        - http://www.opengis.net/def/crs/OGC/1.3/CRS84
        - http://www.opengis.net/def/crs/EPSG/0/4326
        - etc
```

in this way we can avoid to use unsupported CRSs

francbartoli (changes requested):

pygeoapi/openapi.py
```diff
@@ -727,14 +732,14 @@ def get_oas_30(cfg):
                     'parameters': [
                         items_f,
                         items_l,
-                        {'$ref': f"{OPENAPI_YAML['oapif']}#/components/parameters/bbox"},  # noqa
+                        {'$ref': f"{OPENAPI_YAML['oapif-1']}#/components/parameters/bbox"},  # noqa
                         {'$ref': '#/components/parameters/bbox-crs'}
```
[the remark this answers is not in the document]
justb4: Good catch, yes, to make the ref valid, thought had done global replace, but have come in through a rebase.

But, for l ater: this segment is for Coverages. Not sure if/why we should refer to oapif schemas..
Ask @tomkralidis . Looks like there is no official OGC oacov schema and https://github.com/tomkralidis/ogcapi-coverages-1 is used. In the long term think these defs should belong in a (OGC API) Common Schema IMHO...

pygeoapi/schemas/config/pygeoapi-config-0.x.yml
justb4: I think the default should be - http://www.opengis.net/def/crs/OGC/1.3/CRS84 (WGS84 lon, lat ordering) according to the latest [Part 2 Standard](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html) see 6.2.1. For http://www.opengis.net/def/crs/OGC/1.3/CRS84h, impractically, we should inspect the data. In that case users should explicitly configure.

But apart from that with enums we would need to add the entire list (?). The default is now just http://www.opengis.net/def/crs/OGC/1.3/CRS84.

francbartoli: yep, adding the entire list of supported CRSes we could help the users to configure and validate explicitly the list defined in the pygeoapi configuration

justb4: Let me understand: this is a list of hundreds of URIs including all the height-variants like http://www.opengis.net/def/crs/OGC/1.3/CRS84h ? Plus I expect at some point we may want to add regular EPSG: formats.

pygeoapi/openapi.py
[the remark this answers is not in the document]
justb4: So resilved I think.

pygeoapi/schemas/config/pygeoapi-config-0.x.yml
francbartoli: so you think it is too much?

justb4: Well, I think there are like 22000 projections at least, plus all kinds of variants with height. In `proj` there used to be a file called `epsg`. Can't find the URI list but here an example https://wiki.spatialmanager.com/index.php/Coordinate_Systems_objects_list. 

Like @GeoSander said: we may better migrate the whole pygeoapi config schema-thing to Pydantic.

For now it may be useful to check the CRS-list/array on startup for `(py)proj` support. The 'enum' should really come from there as pyproj will determine if the projection URI is valid/supported. We check already the regular expression for URI in [this util function](https://github.com/geopython/pygeoapi/blob/55ce54a72d38904c87d967e7725eb63cadd3da10/pygeoapi/util.py#L572).

francbartoli: I like the migration of pygeoapi config schema to pydantic. @justb4 @GeoSander can we file this improvement in a new issue?

pygeoapi/openapi.py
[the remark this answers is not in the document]
justb4: This is resolved @francbartoli ?

pygeoapi/schemas/config/pygeoapi-config-0.x.yml
justb4: Ok, yes, not part of this PR.

Issue #1111, opened 2023-01-19
Make Unit tests passing in Docker Container


justb4:
**Description**
Failing Unit tests in Docker Container 

**Steps to Reproduce**
* docker run -p 5000:80 --rm --name pygeoapi geopython/pygeoapi:latest test 
* observe failed tests

**Expected behavior**
All tests passing.

**Screenshots/Tracebacks**
See test output:
[test.txt](https://github.com/geopython/pygeoapi/files/10457130/test.txt)

**Environment**
 - OS: Ubuntu Jammy as in Docker Image
 - Python version: 3.10
 - pygeoapi version: latest on jan 19, 2023

**Additional context**
Actually there are only 3 main failure-causes, most can be explained:

* several tests in [test_api.py](https://github.com/geopython/pygeoapi/blob/master/tests/test_api.py)
Many tests IMO should not be here! Like tests for Postgresql CQL,Coverage, EDR. They should be in their Provider (PG) or Service-specific testfile.  `test_api.py` should perform generic API tests, not specific, and relies on unavailable local servers.

* WFS tests: still the wrong GeoSolutions URL is in, thought we fixed that ages ago...
* `test_async_hello_world_process_parallel` fails, cannot explain

justb4:
Think the Utah WFS has strange behaviour:
This works:
https://demo.deegree.org/utah-workspace/services/wfs?SERVICE=WFS&REQUEST=GetCapabilities

But
https://demo.deegree.org/utah-workspace/services/wfs?SERVICE=WFS&REQUEST=DescribeFeatureType
This fails

<?xml version='1.0' encoding='UTF-8'?>
--
  | <ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd" version="2.0.0">
  | <ows:Exception exceptionCode="OperationNotSupported" locator="request">
  | <ows:ExceptionText>Request type 'DescribeFeatureType' is not supported.</ows:ExceptionText>
  | </ows:Exception>
  | </ows:ExceptionReport>

justb4:
Seeing other output via:
https://demo.deegree.org/utah-workspace/console/client/index.xhtml choosing any DescribeFeatureType :

```
[...]

PR #1160, opened 2023-03-01
Move PostgreSQL API tests out of test_api.py 

justb4:
# Overview
Moving PostgreSQL tests out of  `test_api.py` is needed as these require a local PostgreSQL server to be running.
This has two consequences:

* see issue #1111 :  `test_api.py` should able be run in Docker Container
* `test_api.py` IMHO should be able to run with the local repo test data, like in an IDE, not requiring external services

# Related Issue / Discussion
#1111 (partial fix)

# Additional Information
This is simply moving PG tests from one file (`test_api.py`) to another `test_postgresql_provider.py`, no functional changes.
This makes the test suite more modular. Like one does not need to install PostgreSQL, ES, MongoDB, GDAL etc for generic API tests.

# Contributions and Licensing

- [x] I'd like to contribute this fix to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.


webb-ben (approved):
Insofar as cleaning up `test_api.py` such that it passes out of the box, I appreciate this change.

That being said, the tests that were moved are testing the API's ability to parse CQL requests... on the Postgres provider. As more providers add CQL support it will be important clean up line between generic in `api.py` and provider specific cql parsing. The current CQL parsing in `api.py` is only for the Postgres provider - not abstracted for all providers. I do not know what comes next but this is a good first step.

+1

justb4:
So let's merge then? Ok @francbartoli ?

Issue #1043, opened 2022-11-24
Bad Leaflet bounds in Tiles template


vidhav:
**Description**
In the [HTML template for Tiles](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/templates/collections/tiles/index.html#L131), the Leaflet viewport is set to the bbox of the collection. Leaflet orders the coordinates by latitude-longitude (the bbox uses the opposite order).

**Steps to Reproduce**
Set the collection bbox to be `[4,57,32,72]`. The viewport should place you in the Arabian sea.

**Expected behavior**
Should show all of mainland Norway in the viewport.

**Environment**
 - OS: Windows 10
 - Python version: 3.10.5
 - pygeoapi version: 0.13.0

doublebyte1:
@vidhav the order of the coordinates seems to be correct, according to the [Leaflet documentation](https://leafletjs.com/reference.html#latlngbounds).
I think that the problem is that the bounds being passed to this method correspond to the bounds of the extent, rather than the bounds of the actual data (see screenshot).

![pygeoapi-extent](https://user-images.githubusercontent.com/1038897/204131367-cfc22d24-081e-4430-8b0e-2b5bb5d66d5c.png)

doublebyte1:
If you set the extent of the layer in the [configuration of pygeoapi](https://github.com/geopython/diving-into-pygeoapi/blob/main/workshop/exercises/pygeoapi.config.yml#L227), the map will be correctly fitted.

![tiles-extent](https://user-images.githubusercontent.com/1038897/204133265-217bd67d-485e-4436-91ba-69508613c5e6.png)

doublebyte1:
If one day there is support for getting the bounds of a [Leaflet.VectorGrid](https://github.com/Leaflet/Leaflet.VectorGrid#todo), we could try to fit the map to the data extent, instead. Maybe @IvanSanchez knows something about this...

vidhav:
I did set the extent in the the [config file](https://github.com/geopython/diving-into-pygeoapi/blob/main/workshop/exercises/pygeoapi.config.yml#L227), @doublebyte1. I pass it in as `[west, south, east, north]`, so in my understanding it's `[lng, lat, lng, lat]`...

[...]

PR #1163, opened 2023-03-06
Fixed problem with coordinate order, on leaflet preview of tiles

doublebyte1:
# Overview

Leaflet [latlngbounds](https://leafletjs.com/reference.html#latlngbounds) uses a different order (lat,lng) from [pygeoapi's extent](https://github.com/geopython/pygeoapi/blob/master/pygeoapi-config.yml#L167). This causes the tiles preview to fit to an incorrect bounding box.

The problem does not affect the other HTML previews (e.g.: collection, features): only tiles.

# Related Issue / Discussion

@vidhav had pointed this issue here:
https://github.com/geopython/pygeoapi/issues/1043

# Additional Information

# Contributions and Licensing

### DIFF
--- a/pygeoapi/templates/collections/tiles/index.html
+++ b/pygeoapi/templates/collections/tiles/index.html
@@ -131,7 +131,7 @@
       .addTo(map);
 
     map.on('click', clearHighlight);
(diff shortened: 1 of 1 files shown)


tomkralidis:
CI error is unrelated.

Issue #1169, opened 2023-03-13
INSPIRE OAPIF bulk download link (enclosure)


GeoSander:
**Is your feature request related to a problem? Please describe.**
`pygeoapi`'s predefined example collections currently lack a couple of [INSPIRE metadata elements](https://github.com/INSPIRE-MIF/gp-ogc-api-features/blob/master/spec/oapif-inspire-download.md#metadata-elements-of-the-data-set) and a [bulk download link](https://github.com/INSPIRE-MIF/gp-ogc-api-features/blob/master/spec/oapif-inspire-download.md#84-requirements-class-inspire-bulk-download-), causing it to not fully pass the [INSPIRE validator](https://inspire.ec.europa.eu/validator/test-selection/index.html).

These include:
- [x] The response of the `GET /collections` operation requires a link to the license of the dataset in the `links` property (`rel` = license) and a link to the metadata record for the dataset. This link shall have `rel` link parameter `describedby` and `type` link parameter `application/xml`.
- [x] The response of the `GET /collections` operation should include at least one link where `rel` = `enclosure` that allows requesting a representation of the entire _dataset_ **OR** the response of each `GET /collections/{collectionId}` operation should include at least one `enclosure` link that allows requesting a representation of the entire _feature collection_.

**Describe the solution you'd like**
The upper two links can already be configured if needed and will simply work out of the box. We will configure a demo collection that meets these INSPIRE link requirements. Note that we will add the `license` link to each feature collection as recommended [here](https://docs.opengeospatial.org/is/17-069r3/17-069r3.html#rec_core_fc-md-license) under 10.C. The same applies to each `describedby` link as recommended [here](https://docs.opengeospatial.org/is/17-069r3/17-069r3.html#rec_core_fc-md-descriptions) under 9.C. There is currently no support in pygeoapi for adding these links to the `collections` response (as recommended under 9.A and 10.A). It also wouldn't make much sense looking at how pygeoapi is typically being used, where different licenses and schemas often apply for each feature collection.

Any configured bulk download link(s) (`rel` = `enclosure`) will require some code that actually issue an HTTP HEAD request to the path given in the `href` parameter of that link if no `length` parameter was configured (content size in bytes). If the HEAD request returns a 200 and the `Content-Length` header is bigger than 0, the `length` and `type` (from `Content-Type` if not configured or different) parameters will be set and the download link will be included in the collection item `links` section.

We will add the bulk download link to each `GET /collections/{collectionId}` operation.

[...]

PR #1173, opened 2023-03-15
Verify collection enclosure links

GeoSander:
# Overview
This PR adds a URL prefetcher to pygeoapi that can verify collection links of `rel` type `enclosure`, often used for bulk INSPIRE downloads.
These type of links require a `length` parameter (content size in bytes). 

Maintainers can manually set the `length` if the byte size is known or never changes. In that case, pygeoapi will **not** check the URL, but simply list the enclosure link as-is.

If the `length` is omitted, the prefetcher will issue a HEAD request to the specified `href` to get the actual content length. Because it also receives the content type, it will automatically adjust/update the `type` parameter (MIME type) to match the actual one.

So in theory, if maintainers want to add a download URL from which they know that its content regularly updates, they can set any `type` (this is a required schema parameter) and omit the `length`, and pygeoapi will resolve it.

The [configuration documentation](https://github.com/GeoSander/pygeoapi/blob/inspire-oapif-1169/docs/source/configuration.rst#adding-links-to-collections) was updated with a section about adding collection links.

# Related Issue / Discussion
This PR relates to issue #1169, which contains the rationale behind this change. 

# Additional Information
Since this work is part of a funded [tender](https://www.geonovum.nl/themas/invitation-to-tender) in which [Geonovum](https://www.geonovum.nl/about-us) emphasized that it was very important to them that the work would be incorporated into the core, and also because the suggested changes are optional and non-disruptive, we hope that the community is willing to approve this PR. 

# Contributions and Licensing

- [x] I'd like to contribute feature "Verify collection enclosure links" to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -179,7 +179,7 @@ default.
               - type: feature # underlying data geospatial type: (allowed values are: feature, coverage, record, tile, edr)
                 default: true  # optional: if not specified, the first provider definition is considered the default
                 name: CSV
-                # transactions: DO NOT ACTIVATE unless you have setup access contol beyond pygeoapi
+                # transactions: DO NOT ACTIVATE unless you have setup access control beyond pygeoapi
                 editable: true  # optional: if backend is writable, default is false
                 data: tests/data/obs.csv  # required: the data filesystem path or URL, depending on plugin setup
                 id_field: id  # required for vector data, the field corresponding to the ID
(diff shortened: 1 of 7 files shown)


GeoSander:
Thanks for merging @tomkralidis!

PR #1184, opened 2023-03-22
fix(automap): fixes naming conflicts for PostgreSQLProvider

MTachon:
# Overview
Automapping classes and relationships from database schema throws an error if a column name is the same as a relationship name (see https://docs-sqlalchemy.readthedocs.io/ko/latest/orm/extensions/automap.html#handling-simple-naming-conflicts). This PR solves potential naming conflict problems by changing the relationship name if the original one is the same as existing column name.

# Related Issue / Discussion
Have not seen any related issue/discussion.

# Additional Information
No additional information.

# Contributions and Licensing


francbartoli (changes requested):
Is it possible to add a failing test for a conflict?

MTachon:
I guess a `get_collection_items` call on a minimal dataset with a *conflicting* table structure would do just fine. Will look into it!

MTachon:
I have extended the new test data file from PR https://github.com/geopython/pygeoapi/pull/1185 with two new (empty) tables, which results in pygeoapi throwing a `sqlalchemy.exc.ArgumentError` when attempting to publish the table `naming_conflicts_table`. A simple test was added to `test_postgresql_provider.py` which fails in master and passes in this branch.

francbartoli (approved):

PR #1185, opened 2023-03-22
PostgreSQLProvider - Support for tables with multiple geometry/geography columns

MTachon:
# Overview
This PR aims to enable the `PostgreSQLProvider` to deal with database tables with multiple geometry/geography columns. Before https://github.com/geopython/pygeoapi/pull/1182, publishing data from database tables with multiple geometry/geography columns with `PostgreSQLProvider` would throw an error when attempting to serialize to JSON `Features` with multiple geometries/geographies: `PostgreSQLProvider` attempted to serialize the geometry/geography column(s) not specified in `geom_field` of the config file the same way as any other properties. `geoalchemy2.elements.WKBElement` instances could not be directly serialized to JSON, which resulted in `TypeError: ... type <class 'geoalchemy2.elements.WKBElement'> not serializable` errors. With https://github.com/geopython/pygeoapi/pull/1182, `PostgreSQLProvider` successfully **ignores** the geometry/geography columns not specified in the `geom_field` field of the config file, with the proviso that they are not listed in the `properties` field of the same config file. However, pygeoapi's users may want to publish a `FeatureCollection` from a table with multiple geometry/geography columns, where the served `Features` have a `GeometryCollection` geometry type, instead of publishing different `collections` for the different geometries/geographies of the same database table.

This PR gives support to `PostgreSQLProvider` for returning `Features` with `GeometryCollection` geometries, which aggregate multiple geometry/geography table columns.

Let us assume we have a `buildings` database table representing buildings with two geometry columns:
- *footprint* -> `Polygon` geometry
- *position* -> `Point` geometry

The PostgreSQL provider section of the config file for such database table can then be as follows:

```yaml
# If we want to publish only footprint polygons
providers:
    - type: feature
      name: PostgreSQL
      ...
      table: buildings
      geom_field: footprint

# If we want to publish both footprint polygons AND position points as GeometryCollections
providers:
    - type: feature
      name: PostgreSQL
      ...
      table: buildings
      geom_field:
          - footprint
          - position
```

# Related Issue / Discussion
https://github.com/geopython/pygeoapi/pull/1182

# Additional Information
For tables with multiple geometry/geography columns, where some may have a `GeometryCollection` geometry type, these geometries are effectively *flattened* before being aggregated in one single `GeometryCollection` geometry. As such, output `Features` will not have `GeometryCollection` geometries embedded in their output `GeometryCollection` geometry.

The only thing we need to make sure, is that the different geometry/geography columns have the same `SRID` / `storage_crs` as the end-user will interpret the output `GeometryCollection`'s geometries as being in the same *CRS*. See also PR https://github.com/geopython/pygeoapi/pull/1174. If we still want to support geometry/geography columns with different `SRID`, we will need to make further changes to `PostgreSQLProvider`.

# Contributions and Licensing


francbartoli:
Hi @MTachon, I had the same use case with multiple geometry columns in PostGIS so thanks for looking into this! I'm wondering if this could be extended beyond just the Postgres provider

MTachon:
Hi @francbartoli, which provider(s) do you have in mind? `OGRProvider`? `SQLiteGPKGProvider`? Others? Not all data formats support multiple geometry fields. If the list of providers is long, we may want to implement a more generic solution.

PR #1190, opened 2023-03-28
Tinydb updates

tomkralidis:
# Overview
Updates time and geometry handling given updates in OGC API - Records record model (i.e. no long using `properties.extent`, only GeoJSON geometry primitive).
# Related Issue / Discussion
None
# Additional Information
cc @rouault 
# Contributions and Licensing


rouault:
@tomkralidis There are issues with the spatial filtering. I'm working on a fix

rouault:
extra fix on top of this PR in https://github.com/tomkralidis/pygeoapi/pull/4

PR #1196, opened 2023-03-30
Azure Blob Storage STAC provider

tomkralidis:
# Overview
This PR adds functionality to expose Microsoft Blob Storage as a STAC catalog.

# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/pygeoapi/provider/azure_.py
+++ b/pygeoapi/provider/azure_.py
@@ -28,10 +28,10 @@
 # =================================================================
 
 from datetime import datetime
-import io
-from json import loads
 import logging
 import os
+
+from azure.storage.blob import BlobServiceClient
 
 from pygeoapi.provider.base import (BaseProvider, ProviderConnectionError,
                                     ProviderNotFoundError)
@@ -40,8 +40,8 @@ from pygeoapi.util import file_modified_iso8601, get_path_basename, url_join
 LOGGER = logging.getLogger(__name__)
 
 
-class FileSystemProvider(BaseProvider):
-    """filesystem Provider"""
+class AzureBlobStorageProvider(BaseProvider):
+    """Azure blob storage Provider"""
 
     def __init__(self, provider_def):
         """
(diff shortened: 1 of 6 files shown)


kalxas (approved):

requirements-provider.txt
webb-ben, line 1: Do these installs also need to be included in requirements-docker.txt? I don't see debian distributions for them

pygeoapi/provider/azure_.py
```diff
@@ -0,0 +1,363 @@
...
+# WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# =================================================================
+
+from datetime import datetime
+import logging
+import os
+from urllib.parse import urljoin
```
webb-ben, line 33: use pygeoapi.util url_join?

webb-ben (commented):
Is GitHub Action testing possible (and/or required) for this provider? I don't see any tests for STAC providers...

requirements-provider.txt
tomkralidis: > Is GitHub Action testing possible (and/or required) for this provider? I don't see any tests for STAC providers...

We have tests for the [FileSystemProvider](https://github.com/geopython/pygeoapi/blob/master/tests/test_filesystem_provider.py).  Adding tests for Azure would mean setting up a local Azure instance (azurite) in CI.  If there is enough use/demand on STAC/Azure we can add in the future.

tomkralidis: > 

If you mean the azure dependencies, they are best placed in `requirements-provider.txt`.

pygeoapi/provider/azure_.py
tomkralidis: updated.

requirements-provider.txt
webb-ben: Agree that it _should_ be in `requirements-provider.txt`. 

In practice, `requirements-docker.txt` installs the packages that are in `requirements-provider.txt`, but not available as debian distro.

Should https://github.com/geopython/pygeoapi/blob/9debd89cd832e4155d335001376a49cd4f4755c3/Dockerfile#L125
be 
```
&& pip3 install -r  requirements-provider.txt \
```
?

PR #1198, opened 2023-03-30
[WIP] Normalize and enhance breadcrumbs

GeoSander:
# Overview
Currently, the breadcrumbs (navigational links at the top of the HTML pages) are partially generated (e.g. for STAC) and partially hand-written (i.e. defined in the Jinja HTML templates).
Once #1152 will be merged and **_only_** if API rules have been configured with the `strict_slashes` option set to `true`, this may lead to incorrect `href` values for the links.
In general, most breadcrumb links are relative (starting with `/` or `./`) now, which is inconsistent as most links in pygeoapi are absolute. 

This PR will improve this situation by making the `get_breadcrumbs()` function in `utils.py` more advanced, so that it will return consistent breadcrumbs with absolute URLs.

# Related Issue / Discussion
- #1134 
If API rules state that trailing slashes are **not** allowed, some breadcrumbs in which a trailing slash has now been hard-coded into the HTML template will result in a 404 when clicked.

# Contributions and Licensing

- [x] I'd like to contribute this enhancement to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -3648,7 +3648,7 @@ class API:
                     'result': job_output
                 }
                 content = render_j2_template(
-                    self.config, 'jobs/results/index.html',
+                    self.tpl_config, 'jobs/results/index.html',
                     data, request.locale)
 
         return headers, HTTPStatus.OK, content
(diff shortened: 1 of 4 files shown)


tomkralidis:
@GeoSander #1134 has been merged. Let us know when this is ready for review.

GeoSander:
Yes, I noticed @tomkralidis, thanks!
This is not ready yet (I still need to change the template code so that it uses the enhanced `get_breadcrumbs` function), but I was waiting for the merge.
Will let you know when it's done 👍

PR #1206, opened 2023-04-06
1204 oapi process execution use prefer header

### LINKED ISSUE #1191: BaseProcessor.execute() signature should include a single parameter
**Description**
I'm starting implementation of an alternative process manager for pygeoapi based on [prefect]. While studying pygeoapi's source code I encountered the following discrepancy:

When looking at the base class for processes, as present in

https://github.com/geopython/pygeoapi/blob/466a1df72393e787c4bccde2973f77914d7a1a12/pygeoapi/process/base.py#L51

the signature of the `execute()` method does not show it supporting any parameters.

However, the example `HelloWorldProcessor` defines its `execute()` method as taking one `data` parameter, as shown in:

https://github.com/geopython/pygeoapi/blob/466a1df72393e787c4bccde2973f77914d7a1a12/pygeoapi/process/hello_world.py#L120

Moreover, both built-in process manager implementations in pygeoapi also call a process' `execute()` method by passing it a single parameter, as can be seen in:

- the base process manager (which is used by `pygeoapi.process.manager.tinydb_.TinyDbManager` - https://github.com/geopython/pygeoapi/blob/466a1df72393e787c4bccde2973f77914d7a1a12/pygeoapi/process/manager/base.py#L195
- the dummy process manager (which is used by default by pygeoapi) - https://github.com/geopython/pygeoapi/blob/466a1df72393e787c4bccde2973f77914d7a1a12/pygeoapi/process/manager/dummy.py#L86

In all cases the parameter being passed to a process' `execute()` method is a dictionary with the provided process input data. As such, it seems the signature of `pygeoapi.process.base.BaseProcessor.execute()` should be updated in order to reflect the expectation that a pluggable process manager should try to execute processes by passing them this parameter.

[prefect]: https://www.prefect.io/

### LINKED ISSUE #1204: process execution mode ought to be read from request headers
**Description**
According to the [OAPI - Processes spec, section 7.11.2.3. Execution Mode](https://docs.ogc.org/is/18-062r2/18-062r2.html#toc32):

> A process may be executed synchronously or asynchronously.

> In which of these two modes a server responds is a function of the job control options specified in the [process description](https://docs.ogc.org/is/18-062r2/18-062r2.html#sc_process_description) and the presence or absence of the HTTP Prefer header ([IETF RFC 7240](http://tools.ietf.org/rfc/rfc7240.txt)).

The spec goes on to mention that async execution should be used when `Prefer: respond-async` and sync should be used when `Prefer: wait`. 

Finally, the spec also mentions that the eventual response should also include a `Preference-Applied: [respond-async|wait]` header, to denote which execution mode has been chosen by the server.

However, the current implementation of pygeoapi expects to find the process execution mode in the request body, as the `mode` parameter:

https://github.com/geopython/pygeoapi/blob/48b5b3b163da3e86189cf0c3e065b9ebe6449dac/pygeoapi/api.py#L3439

ricardogsilva:
# Overview
This PR modifies OAPI - Process request handling in order to extract execution mode from the `Prefer` HTTP header of the request, as discussed in #1204.

# Related Issue / Discussion
- fixes #1204
- fixes #1191 - this is a small fix that just adds the parameter to the base class' method, as discussed in the issue

# Additional Information
In addition to getting the requested execution mode from the HTTP request's `Prefer` header, the proposed implementation also delegates decision about which execution mode to use to the process manager.

Additionally, and in compliance with the OAPI - Processes spec, when there is a `Prefer` header in the HTTP request, pygeoapi now replies back with a `Preference-Applied` response header too.

# Contributions and Licensing

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -3549,7 +3549,7 @@ class API:
             execution_result = self.manager.execute_process(
                 process, data_dict, execution_mode=execution_mode)
             job_id, mime_type, outputs, status, additional_headers = execution_result
-            headers.update(additional_headers)
+            headers.update(additional_headers or {})
             headers['Location'] = f"{self.base_url}/jobs/{job_id}"
         except ProcessorExecuteError as err:
             LOGGER.error(err)
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -3546,9 +3546,9 @@ class API:
             execution_mode = None
         try:
             LOGGER.debug('Executing process')
-            execution_result = self.manager.execute_process(
+            result = self.manager.execute_process(
                 process, data_dict, execution_mode=execution_mode)
-            job_id, mime_type, outputs, status, additional_headers = execution_result
+            job_id, mime_type, outputs, status, additional_headers = result
             headers.update(additional_headers or {})
             headers['Location'] = f"{self.base_url}/jobs/{job_id}"
         except ProcessorExecuteError as err:
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -74,7 +74,8 @@ from pygeoapi.provider.tile import (ProviderTileNotFoundError,
                                     ProviderTileQueryError,
                                     ProviderTilesetIdNotFoundError)
 from pygeoapi.models.cql import CQLModel
-from pygeoapi.util import (dategetter, DATETIME_FORMAT, UrlPrefetcher,
+from pygeoapi.util import (dategetter, RequestedProcessExecutionMode,
+                           DATETIME_FORMAT, UrlPrefetcher,
                            filter_dict_by_key_value, get_provider_by_type,
                            get_provider_default, get_typed_value, JobStatus,
                            json_serial, render_j2_template, str2bool,
@@ -3544,18 +3545,17 @@ class API:
         data_dict = data.get('inputs', {})
         LOGGER.debug(data_dict)
 
-        is_async = data.get('mode', 'auto') == 'async'
-        if is_async:
-            LOGGER.debug('Asynchronous request mode detected')
-
-        if is_async and not self.manager.is_async:
-            LOGGER.debug('async manager not configured/enabled')
-            is_async = False
-
+        try:
+            execution_mode = RequestedProcessExecutionMode(
+                request.headers.get("Prefer"))
+        except ValueError:
+            execution_mode = None
         try:
             LOGGER.debug('Executing process')
-            job_id, mime_type, outputs, status = self.manager.execute_process(
-                process, data_dict, is_async)
+            execution_result = self.manager.execute_process(
+                process, data_dict, execution_mode=execution_mode)
+            job_id, mime_type, outputs, status, additional_headers = execution_result
+            headers.update(additional_headers)
             headers['Location'] = f"{self.base_url}/jobs/{job_id}"
         except ProcessorExecuteError as err:
             LOGGER.error(err)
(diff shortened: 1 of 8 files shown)


pygeoapi/api.py
```diff
@@ -3393,12 +3394,6 @@ def execute_process(self, request: Union[APIRequest, Any],
                 HTTPStatus.NOT_FOUND, headers,
                 request.format, 'NoSuchProcess', msg)
 
-        if not self.manager:
-            msg = 'Process manager is undefined'
-            return self.get_exception(
-                HTTPStatus.INTERNAL_SERVER_ERROR, headers,
-                request.format, 'NoApplicableCode', msg)
-
```
ricardogsilva, line 3401: Removed this bit because it does not apply - as per this class' `__init__()` method, there is always a manager, so it is no longer relevant to test if `self.manager` exists

pygeoapi/api.py
```diff
@@ -3435,19 +3430,17 @@ def execute_process(self, request: Union[APIRequest, Any],
...
-        if is_async and not self.manager.is_async:
-            LOGGER.debug('async manager not configured/enabled')
-            is_async = False
-
+        try:
+            execution_mode = RequestedProcessExecutionMode(
+                request.headers.get("Prefer"))
+        except ValueError:
+            execution_mode = None
```
ricardogsilva, line 3437: requested execution mode is now gotten from the `Prefer` HTTP header, if it exists

pygeoapi/api.py
```diff
@@ -3435,19 +3430,17 @@ def execute_process(self, request: Union[APIRequest, Any],
...
         try:
             LOGGER.debug('Executing process')
-            mime_type, outputs, status = self.manager.execute_process(
-                process, job_id, data_dict, is_async)
+            execution_result = self.manager.execute_process(
+                process, job_id, data_dict, execution_mode=execution_mode)
+            mime_type, outputs, status, additional_headers = execution_result
+            headers.update(additional_headers or {})
```
ricardogsilva, line 3443: We now pass the requested execution mode to the process manager and let it decide whether to honor the request or not, based on its own configuration and that of the process being executed

pygeoapi/process/base.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, processor_def: dict, process_metadata: dict):
         self.name = processor_def['name']
         self.metadata = process_metadata
 
-    def execute(self) -> Tuple[str, Any]:
+    def execute(self, data: dict) -> Tuple[str, Any]:
```
ricardogsilva, line 51: this is a fix for #1191

pygeoapi/util.py
```diff
@@ -493,6 +493,16 @@ def get_provider_default(providers: list) -> dict:
...
 
 
+class ProcessExecutionMode(Enum):
+    sync_execute = "sync-execute"
+    async_execute = "async-execute"
+
+
+class RequestedProcessExecutionMode(Enum):
+    wait = "wait"
+    respond_async = "respond-async"
```
ricardogsilva, line 503: These are taken from the OAPI - Processes spec

ricardogsilva:
@tomkralidis 

bump :sweat_smile: - any chance I could get this reviewed soon?

tomkralidis (changes requested):
Minor change requests -- good work!

We also need updates to the [documentation](https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-processes.html#processing-examples) as part of this PR given the breaking change.

tomkralidis (approved):

Issue #1207, opened 2023-04-06
process manager should be responsible for generating a job id


ricardogsilva:
Pygeoapi identifies process execution jobs by their respective `job_id`. The current implementation allows a job id to be generated by one of two ways:

- Either it is supplied by the client, when requesting a process execution, which does not seem to be supported by the [OAPI - Processes spec, section 7.11.2.1](https://docs.ogc.org/is/18-062r2/18-062r2.html#toc32), and therefore would likely be eligible for removal

- Or it is generated by `pygeoapi.api.API.execute_process()` method.

Both these options are visible in:

https://github.com/geopython/pygeoapi/blob/48b5b3b163da3e86189cf0c3e065b9ebe6449dac/pygeoapi/api.py#L3434

These are painful to work with when implementing an alternative process manager. A process manager that delegates to an external system (such as prefect, which is my current use case) may not have the liberty to specify a job id to the third-party system and may instead be given one by this third-party upon trying to execute a process. 

As such, since the process manager is already responsible for:

- Executing a process, _i.e._ creating a job
- Retrieving details about existing jobs

It would be be nice if the process manager would also be the entity that is responsible for providing job ids to pygeoapi's outer API layer.

This would allow the process manager to better encapsulate job-related operations, including their creation.

PR #1209, opened 2023-04-07
Process manager generate job ids

ricardogsilva:
# Overview
This PR modifies the pygeoapi process manager to be responsible for generating job ids, as discussed in #1207.

The proposed implementation simply makes the generation of a job's id part of the process manager's set of responsabilities, which means that the job id is now returned from the process manager's `execute_process()` method, rather than it being an input to this method.

# Related Issue / Discussion
- #1207 

# Contributions and Licensing

### DIFF
--- a/pygeoapi/process/manager/base.py
+++ b/pygeoapi/process/manager/base.py
@@ -283,7 +283,7 @@ class BaseManager:
         else:
             LOGGER.debug('Asynchronous execution')
             result = self._execute_handler_async(p, job_id, data_dict)
-        return job_id, *result
+        return (job_id, *result)
 
     def __repr__(self):
         return f'<BaseManager> {self.name}'
--- a/pygeoapi/process/manager/base.py
+++ b/pygeoapi/process/manager/base.py
@@ -27,16 +27,16 @@
 #
 # =================================================================
 
+from datetime import datetime
 import json
 import logging
-import uuid
-from datetime import datetime
 from multiprocessing import dummy
 from pathlib import Path
 from typing import Any, Tuple
+import uuid
 
-from pygeoapi.util import DATETIME_FORMAT, JobStatus
 from pygeoapi.process.base import BaseProcessor
+from pygeoapi.util import DATETIME_FORMAT, JobStatus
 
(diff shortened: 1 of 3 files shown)


ricardogsilva:
@tomkralidis 

The failing test reported by CI seems to be unrelated to my changes - I've tracked it down to this failed request

https://demo.mapserver.org/cgi-bin/msautotest?version=1.3.0&service=WMS&request=GetMap&bbox=-90.0%2C-180.0%2C90.0%2C180.0&crs=EPSG%3A4326&layers=world_latlong&styles=default&width=500&height=300&format=image%2Fpng&transparent=TRUE

Indeed demo.mapserver.org is returning `502`

justb4:
Yes MapServer demo is temporarily down due to upgrades, see also:
https://lists.osgeo.org/pipermail/mapserver-users/2023-April/083093.html

pygeoapi/process/manager/base.py
```diff
@@ -27,9 +27,10 @@
 #
 # =================================================================
 
-from datetime import datetime
```
tomkralidis, line 30: Can we keep the order of imports:

- standard library (alphabetical order)
- blank line
- 3rd party imports (PyPI)  (alphabetical order)
- blank line
- local imports (alphabetical order)

pygeoapi/process/manager/dummy.py
```diff
@@ -28,6 +28,7 @@
 # =================================================================
 
 import logging
+import uuid
```
tomkralidis, line 31: Same comment here, RE: import ordering

tomkralidis (changes requested):

pygeoapi/process/manager/base.py
ricardogsilva: I thought this was in line with keeping the order? 

I usually put `from` imports below straight ones because it is also how [isort](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html#controlling-how-isort-sections-from-imports) and other tools do it, and because it is sorted alphabetically. Anyway, no biggie I'll switch it up :wink:

PR #1215, opened 2023-04-12
Refactored process manager to know about processes

### LINKED ISSUE #1201: job manager to be responsible for creating an instance of a process
In pygeoapi:

```python
```

We'd have:

```python
```

### LINKED ISSUE #1213: process manager to be given config of known processes during initialization

ricardogsilva:
# Overview
This PR modifies the base process manager class in order to have it accept the configuration of known processes during initialization.

With the proposed implementation the process manager know becomes aware of existing processes and can be treated as the source of truth whenever there is a need to retrieve process-related details. Additional implementation details:

- Added the `pygeoapi.process.manager.get_manager(pygeoapi_config)` factory function, which becomes the standard way of getting the manager object - this function reads the pygeoapi config and gets both the manager and whatever processes defined therein, and then calls `pygeoapi.plugins.load_plugin()`.
- Added a `get_processor(process_id)` method to the base manager class, which can be used whenever there is a need to get a process instance. This is a fix for #1201 and contributes to making the process manager fully responsible for managing processes and jobs

# Related Issue / Discussion
- fixes #1213
- fixes #1201

# Contributions and Licensing

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -62,26 +62,26 @@ from pygeoapi.formatter.base import FormatterSerializationError
 from pygeoapi.linked_data import (geojson2jsonld, jsonldify,
                                   jsonldify_collection)
 from pygeoapi.log import setup_logger
-from pygeoapi.process.base import (
-    ProcessorExecuteError,
-    ProcessorGenericError,
-)
-from pygeoapi.process.manager import get_manager
+from pygeoapi.models.cql import CQLModel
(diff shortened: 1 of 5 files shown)


ricardogsilva:
@tomkralidis , can you please review this after #1206 , and hopefully be able to merge both soon? Thanks in advance and sorry for bugging

francbartoli:
I guess we can close this in favor of #1222, can you confirm @ricardogsilva?

tomkralidis:
+1.

PR #1222, opened 2023-04-23
Refactor of process-related functionality

### LINKED ISSUE #1193: Calling the process execution endpoint via HTTP GET shows erroneous error message
**Description**
When calling a process' execute URL via HTTP GET I'm getting an error response with this misleading content:

```
http -v localhost:5000/processes/hello-world/execution
GET /processes/hello-world/execution HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: localhost:5000
User-Agent: HTTPie/2.6.0



HTTP/1.1 404 NOT FOUND
Connection: close
Content-Language: en-US
Content-Length: 74
Content-Type: application/json
Date: Thu, 30 Mar 2023 17:22:36 GMT
Server: Werkzeug/2.2.3 Python/3.10.6
X-Powered-By: pygeoapi 0.14.0

{
    "code": "NoSuchProcess",
    "description": "Identifier not found"
}


```

**Steps to Reproduce**
- start pygeoapi
- make a GET request to the endpoint of the `hello-world` process, which is configured by default

**Expected behavior**
I would expect the server to reply with a message stating that I am using the wrong HTTP method, as process execution is supposed to be triggered via HTTP POST and not GET, as mentioned in [the OAPI - Processes spec](https://docs.ogc.org/is/18-062r2/18-062r2.html#toc32)

### LINKED ISSUE #1199: Use a pydantic model to define the schema of processes
The various schemas for working with processes are defined in the OAPI - Processes spec but pygeoapi has a loose implementation, which makes it a bit painful to implement custom processes and managers. 

It would be nice if pygeoapi would provide some pydantic models to work with process-related entities (process, inputs, etc). Something similar already seems to be implemented under `pygeoapi.models`.

I guess the alternative is to keep the current implementation, which seems to work too, albeit in a less structured way.

**Additional context**
I think I have already implemented most of the relevant models in my ongoing work with using prefect as a process manager, as shown here (still full of WIP everywhere):

https://github.com/ricardogsilva/pygeoapi-prefect/blob/main/src/pygeoapi_prefect/schemas.py

This is to say that I would be willing to produce a PR with these changes, if the maintainers find it a good idea.

### LINKED ISSUE #1201: job manager to be responsible for creating an instance of a process

### LINKED ISSUE #1213: process manager to be given config of known processes during initialization

### LINKED ISSUE #1217: Add support for job list filter params
According to the [OAPI - Processes spec: Job list Operation](https://docs.ogc.org/is/18-062r2/18-062r2.html#toc49), the job list endpoint shall support some filtering parameters. Namely:

- [ ] `type`
- [ ] `processID`
- [ ] `status`
- [ ] `datetime`
- [ ] `minDuration`
- [ ] `maxDuration`
- [ ] `limit`

pygeoapi does not currently support any of these, but it would be cool if it did :smiley:

ricardogsilva:
# Overview
This PR has a substantial overhaul of the OGC API - Processes - related features.

Main modifications:

- Base process and manager classes derive from Python's [abc.ABC](https://docs.python.org/3/library/abc.html) and use `abstractmethod` to signal those methods that are expected to be implemented by plugins implementing additional processes/manager;
- Schemas for process-related operations are defined using [pydantic](https://docs.pydantic.dev/) models (in analogous way as is done elsewhere in the codebase). These models follow the schemas shown on the OAProc specification;
- The pygeoapi API and process/job manager has been refactored in order to align with the OAProc specification:
  - Support for both `document` and `raw` response types when there are multiple process outputs
  - Support for both `value` and reference` transmission modes for each output of a process
  - Support for job list filtering parameters
- Several enhancements to the manager in order to give it more responsibilities in handling the workflows related to execution and display of job results, including sorting jobs, passing storage locations, etc.
- An additional sample process named `pygeoapi.processes.hello_world.GreeterProcessor`. This process returns an actual JSON object, so it serves as another example.

This is part of an ongoing work to bring OAProc support in pygeoapi inline with the published specification.

This PR is intended for completion and review during the upcoming [OGC code sprint](https://developer.ogc.org/sprints/20/)

# Related issues
- fixes #1217
- fixes #1213
- fixes #1201
- fixes #1199
- fixes #1193

# Contributions and Licensing

### DIFF
--- a/docs/source/data-publishing/ogcapi-processes.rst
+++ b/docs/source/data-publishing/ogcapi-processes.rst
@@ -9,17 +9,52 @@ fashion (inputs, outputs).
 pygeoapi implements OGC API - Processes functionality by providing a plugin architecture, thereby
 allowing developers to implement custom processing workflows in Python.
 
-A `sample`_ ``hello-world`` process is provided with the pygeoapi default configuration.
+Two `sample`_ processes process are provided with the pygeoapi default configuration:
+
+* ``HelloWorldProcessor`` - A simple process that takes one mandatory ``name`` input and an additional optional
+  ``message`` input and produces an ``echo`` output, which is a string with a greeting that includes the provided
+  name and message. The output is of type ``text/plain``.
+
+* ``GreeterProcessor`` - A simple process that takes one mandatory ``num_greetings`` input and produces a ``greetings``
+  output, which is a JSON object with a single ``greetings`` property - this property is a list of generated greetings.
+  The output is of type ``application/json``.
+
 
 Configuration
 -------------
 
+Processes are configured as normal resources in the pygeoapi configuration file. Their configuration follows the
+following pattern:
+
 .. code-block:: yaml
 
-   processes:
-       hello-world:
+   resources:
+       <process-id>:
            processor:
-               name: HelloWorld
+               name: <process-name | process-dotted-path>
+
+Where ``<process-id>`` is replaced by the process identifier and ``processor.name`` and either be:
+
+1. the process name, if it is known to pygeoapi, _i.e._ if this process is part of pygeoapi core
+2. the process dotted path, for processes that are part of third-party Python packages which have already been installed
+
+For example, this is the relevant configuration snippet that is included in pygeoapi's ``pygeoapi-config.yml``:
+
+.. code-block:: yaml
+
+   resources:
+     hello-world:
+       type: process
+       processor:
+         name: HelloWorld
+     greeter:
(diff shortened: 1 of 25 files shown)


ricardogsilva:
@tomkralidis , a gentle bump on this

ricardogsilva:
I'm closing this PR as we've been introducing some related changes in a more granular way.

PR #1234, opened 2023-05-02
custom cli commands

### LINKED ISSUE #1232: Allow plugins to register custom pygeoapi CLI commands
It would be nice if plugin authors had the possibility to register additoinal CLI commands. These would be discoverable by running `pygeoapi plugins` and would be run like this:

```sh
pygeoapi plugins <some-command>
```

I am willing to provide a PR with this functionality.

ricardogsilva:
# Overview
This PR implements the capability of loading additional CLI commands from third-party pygeoapi plugins.

Implementation uses [importlib_metadata](https://importlib-metadata.readthedocs.io/en/latest/index.html), which has become part of Python's standard library since Python 3.8 (but is also backported to earlier versions) to look for relevant entry points and adds them as additional click commands/groups. These become available to run by calling

```sh
pygeoapi plugins <command-name>
```

This PR includes a more thorough example in the docs.

# Related Issue / Discussion
- fixes #1232 

# Contributions and Licensing


tomkralidis (changes requested):
Nice addition @ricardogsilva.  FYI I made a small change to `requirements.txt` to resolve conflicts.

Can you add `python3-importlib-metadata` to `debian/control` as part of this PR?  Thanks.

pygeoapi/__init__.py
```diff
@@ -30,16 +32,47 @@
...
 @click.version_option(version=__version__)
 def cli():
     pass
 
 
+@_find_plugins()
+@cli.group()
+def plugins():
+    """Additional commands provided by third-party pygeoapi plugins"""
+    ...
```
tomkralidis, line 73: Sorry, one last nitpick: can `...` be updated to `pass` instead?

tomkralidis (changes requested):

pygeoapi/__init__.py
ricardogsilva: sure

tomkralidis (approved):

Issue #824, opened 2021-11-24
Cloud Data Access Within Providers


Dylan-Pugh:
I've been experimenting with using AWS S3 as a backing store within a custom provider (extending `XarrayEDRProvider`), but have been running into a few issues. 

I've tried two different approaches so far (opening a NetCDF file stored in S3):

### 1. Retrieving the file with [`boto3`](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) and saving to a [`tempfile`](https://docs.python.org/3/library/tempfile.html)
This approach works, but requires a few helper methods, and I can see it becoming a bit clunky going forward:

```python
if provider_def['data'].startswith('s3://'):
      tmp = tempfile.NamedTemporaryFile(delete=False)
      self.fetch_s3_file(provider_def['data'], tmp.name)
      self.data = tmp.name`
```

code for `fetch_s3_file`:

```python
def fetch_s3_file(self, object_uri, out_file):
        
        bucket_name, key_name = self.split_s3_uri(object_uri)

        s3 = boto3.client('s3')

        try:
            s3.download_file(bucket_name, key_name, out_file)
        except Exception as err:
            LOGGER.warning(err)
            raise ProviderNoDataError(err)
```

### 2. Using [`fsspec`](https://filesystem-spec.readthedocs.io/en/latest/index.html) to handle S3 operations
This is my preferred method, but I've been running into some issues. This code fails, and returns the following error: `pygeoapi.provider.base.ProviderConnectionError: I/O operation on closed file.`

```python
with fsspec.open(S3_URL) as f:
    ds = xr.open_dataset(f)
```

This code (using a work-around mentioned [in this s3fs issue](https://github.com/fsspec/s3fs/issues/337)) does work, but is extremely slow:

```python
self._data = open_func(fsspec.open(provider_def['data']).open())
```

**Describe the solution you'd like**

Ideally I'd like to get fsspec working, and then figure out the best way to integrate/abstract the solution, so we can use data from a number of different cloud services going forward. 

Is anyone working on something similar? 

I've reviewed these issues, which may provide additional context:

https://github.com/geopython/pygeoapi/issues/724

https://github.com/geopython/pygeoapi/issues/807

I'm also happy to provide more detail - thanks!

ksonda:
Not that I have much to contribute atm, but I'm interested/support. Would also like to extend Approach 2 to any s3-compliant provider. However, I can also see the merits of Approach 1 for the simplest providers (csv, geojson)

[...]

PR #1235, opened 2023-05-02
S3 Cloud Storage Access for EDR Provider

sjordan29:
# Overview
Update to allow use of AWS S3 or Open Storage Network pods with S3 access links as backing stores when using the EDR provider. The updates leverage `fsspec` to handle S3 operations. Users provide an S3 link in the data field, and any relevant `storage_options` required within the configuration file.

# Related Issue / Discussion
#824 

# Additional Information

# Contributions and Licensing

- [x] I'd like to contribute S3 cloud access for EDR provider to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.


docs/source/data-publishing/ogcapi-edr.rst
```diff
@@ -58,12 +58,29 @@ The `xarray-edr`_ provider plugin reads and extracts `NetCDF`_ and `Zarr`_ data
...
 
    `Zarr`_ files are directories with files and subdirectories.  Therefore
    a zip file is returned upon request for said format.
 
+.. note::
+   When referencing data stored in an S3 bucket, be sure to provide the full
+   S3 URL. Any parameters required to open the dataset using fsspec can be added
+   to the config file under `storage_options`.
+
+
```
webb-ben, line 83: Because this PR adds S3 support for both OA-EDR and  OA-Coverages, it might be worth adding this documentation to [ogcapi-coverages/xarray](https://github.com/geopython/pygeoapi/blob/master/docs/source/data-publishing/ogcapi-coverages.rst) as well

webb-ben (approved):
+1

webb-ben:
Not sure the implications this has for the S3 capacities of [pygeoapi.util.read_data](https://github.com/geopython/pygeoapi/blob/953b6fa74d2ce292d8f566c4f4d3bcb4161d6e95/pygeoapi/util.py#L574). The function does appears to perform a different task than the capacities provided by s3fs.

aufdenkampe:
@webb-ben, thanks for your review and suggestions!

@tomkralidis, we're moving forward with deploying our fork for USGS, but they would much prefer to deploy from your parent repo. Is there anything we can do to help move this PR toward a merge?

docs/source/data-publishing/ogcapi-edr.rst
aufdenkampe: Looks like @sjordan29 did that with https://github.com/geopython/pygeoapi/pull/1235/commits/ab198109b0f161a785874ca0b029776d0bda1ef8

docs/source/data-publishing/ogcapi-edr.rst
```diff
@@ -58,12 +58,29 @@ The `xarray-edr`_ provider plugin reads and extracts `NetCDF`_ and `Zarr`_ data
...
             mimetype: application/zip
+    
+    providers:
+       - type: edr
+         name: xarray-edr
+         data: s3://power-analysis-ready-datastore/power_901_annual_meteorology_utc.zarr
+         format:
+            name: zarr
+            mimetype: application/zip
+         storage_options:
```
tomkralidis, line 69: Can we rename to `options` for some consistency with other providers?  And then namespace with `s3`?  So:

```
options:
  s3:
    anon: true
    requestor_pays: false
```

pygeoapi/provider/xarray_.py
```diff
@@ -67,8 +68,19 @@ def __init__(self, provider_def):
                     open_func = xarray.open_mfdataset
                 else:
                     open_func = xarray.open_dataset
+            if provider_def['data'].startswith('s3://'):
```
tomkralidis, line 71: Given the `options` proposal in this PR, then would something like the below help?

```python
            if provider_def['data'].startswith('s3://'):
                LOGGER.debug('Data is stored in S3 bucket.')
                if 's3' in provider_def.get('options', {}):
                    s3_options = provider_def['options']['s3']
                else:
                    s3_options = {}

                data_to_open = fsspec.get_mapper(self.data, s3_options)
                LOGGER.debug('Completed S3 open function')
            else:
                data_to_open = self.data
```

tomkralidis (changes requested):
Great work here!  See comments.  Once this is addressed/approved/merged, I can work on a wider implementation so additional providers can benefit.

sjordan29:
Thank you @tomkralidis and @webb-ben - I made the changes you requested. Note that I changed the requirements to `s3fs<=2023.6.0`; as I was testing these updates, I realized that there were releases last week for `s3fs` and `fsspec` that resulted in the following error: `NotImplementedError: Calling sync() from within a running loop.` 

Do you think I should try to resolve this issue with the most recent version of `s3fs` as part of this PR or just open a new issue to flag this should be fixed down the road?

tomkralidis:
Thanks @sjordan29.  Given this is an optional provider dependency and does not affect core functionality, the existing pin is fine.  Thanks for updating this PR!

Issue #1237, opened 2023-05-03
Last page of pagination contains link to next page, although there are no elements on next page


Diuke:
**Description**
When paginating a response (adding offset and limit), the API adds a link to the "next" page. When the last page is reached and the total number of elements is a multiple of the "limit", the "next" link is added, but the next page is empty.

For example, consider a collection of 10 items, with pages of 5 elements (limit=5). The first page should have a link to the next page (limit=5&offset=5), but the next page should not, as the maximum number of elements is reached. Although, the next page does contain a link to the next page and the next page is empty.

**Steps to Reproduce**
Visit https://demo.pygeoapi.io/stable/collections/obs/items?limit=5
Click the button "Next" to go to the next page.
Notice that there are no items in the collection as the next page comes empty.

**Expected behaviour**
The response of the last page should not contain the "next" page link.

**Screenshots/Tracebacks**
No screenshots

**Environment**
 - OS: Ubuntu 22.04
 - Python version: 3.10.6
 - pygeoapi version: master

**Additional context**
-

PR #1240, opened 2023-05-04
Pagination issue

Diuke:
# Overview
As for #1237, I updated api.py to take into account the total number of records (numberMatched) instead of just the limit for appending the "next" link to the list of links of the response.

# Related Issue / Discussion
Issue #1237 

# Additional Information

# Contributions and Licensing

- [X] I'd like to contribute bugfix for pagination to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.


tomkralidis:
@Diuke can you update the unit test that is failing in CI?

totycro:
I've opened PR #1333 which is exactly this PR plus a commit which should fix the failing test.

Issue #1239, opened 2023-05-03
EDR Time Querying Error


sjordan29:
**Description**
When I try to use the datetime query with the EDR provider, I get 
`2023-05-03 11:12:02   pygeoapi-docker-fabric  \| TypeError:   '<' not supported between instances of 'str' and 'Timestamp'`

I have a workaround saved on an [edr-time-query](https://github.com/LimnoTech/pygeoapi/tree/edr-time-query) branch of [LimnoTech's forked pygeoapi](https://github.com/LimnoTech/pygeoapi/tree/master) that I could issue a PR for. However, I'm curious if there's something wrong with how I'm structuring the request.

**Steps to Reproduce**
1. I was testing with a locally stored subset of a NASA POWER dataset (s3://power-analysis-ready-datastore/power_901_annual_meteorology_utc.zarr):
```python
import xarray as xr 
import s3fs 

def s3open(path, storage_options):    
    fs = s3fs.S3FileSystem(**storage_options)
    return s3fs.S3Map(path, s3=fs)

storage_options = {'anon': True, 'requester_pays': False}
nasa = xr.open_zarr(s3open('s3://power-analysis-ready-datastore/power_901_annual_meteorology_utc.zarr', storage_options ))
nasa_sub = nasa[['RH2M', 'GWETROOT']]
```
2.  Use docker to create test build implementation of pygeoapi serving this dataset
3. I attempted the following request for a single datetime: `http://localhost:5000/api/fabric/pygeoapi/collections/s3/position?coords=POINT%28-100%2040%29&datetime=1981-12-31T00%3A00%3A00Z&parameter-name=GWETROOT&f=json`

**Expected behavior**
Expected to receive json results for my request. 

**Screenshots/Tracebacks**
Sample of logs:
```
2023-05-03 11:12:02   pygeoapi-docker-fabric  \|   [2023-05-03T12:12:02Z]   {/mambaforge/lib/python3.9/site-packages/pygeoapi/provider/xarray_.py:158}   DEBUG - Determining rangetype for GWETROOT
--
2023-05-03 11:12:02   pygeoapi-docker-fabric  \|   [2023-05-03T12:12:02Z]   {/mambaforge/lib/python3.9/site-packages/pygeoapi/provider/xarray_.py:158}   DEBUG - Determining rangetype for RH2M
[...]

PR #1247, opened 2023-05-15
Time Querying in EDR Provider

sjordan29:
# Overview
Expanded capabilities for time queries in the EDR provider, including:

- Resolving `TypeError` for single time query with certain datetime formats
- Handling datetime queries with UTC timezone, specified with trailing `Z`
- Querying over bounded and unbounded time intervals

# Related Issue / Discussion
#1239 

# Contributions and Licensing


pygeoapi/provider/xarray_edr.py
```diff
@@ -213,16 +219,65 @@ def cube(self, **kwargs):
...
                           for var_name, var in data.variables.items()}
         }
 
         return self.gen_covjson(out_meta, data, self.fields)
+
+    def _make_datetime(self, datetime_):
+        """
+        Make xarray datetime query
+        :param datetime_: temporal (datestamp or extent)
+        :returns: xarray datetime query
```
webb-ben, line 237: Add empty line before first `:param:` and before `:return:` in function headers. i.e.
```
        Make xarray datetime query

        :param datetime_: temporal (datestamp or extent)

        :returns: xarray datetime query

File: pygeoapi/provider/xarray_edr.py [hunk 2]
```diff
@@ -213,16 +219,65 @@ def cube(self, **kwargs):
...
+        """
+        datetime_ = datetime_.rstrip('Z').replace('Z/', '/')
+        if '/' in datetime_:
+            begin, end = datetime_.split('/')
+            if begin == '..':
+                begin = self._data[self.time_field].min().values
+            elif end == '..':
```
webb-ben, line 244: Although it would be redundant to do `?datetime=../..`, this would throw an error. Changing `elif end == '..':` to `if end == '..':` would prevent.

webb-ben (approved):
+1

pygeoapi/provider/xarray_edr.py
```diff
@@ -28,6 +28,7 @@
 # =================================================================
 
 import logging
+import numpy as np
```
tomkralidis, line 31: Add a blank line between modules from standard library and 3rd party, i.e.:

```python
import logging

import numpy as np

from pygeoapi.provider.base import ProviderNoDataError, ProviderQueryError
from pygeoapi.provider.base_edr import BaseEDRProvider
```

tomkralidis (changes requested):
Minor change requests along with @webb-ben -- great work!

tomkralidis:
Thanks @sjordan29 !

PR #1252, opened 2023-05-23
Dockerfile improvements

totycro:
# Overview

These changes allow developing locally with the dockerfile by making use of image caches. Previously changing any character in the source code would lead to a rebuild time of several minutes. Now it should only be a matter of seconds.

The base image is now fixed to a specific ubuntu jammy version (currently `jammy-20230425`). This should reduce the image size because before, the old packages were included in the image as well as the update. It should also make the builds more reproducible and give control over when libraries are updated.

Also pip cache is now disabled.

In my local builds, the image size is reduced from 1.21 GB to 1.19 GB by this change.

# Contributions and Licensing

- [ X ] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.
- [ X ] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 .git
 .github
 .pytest_cache
+.mypy_cache
 build
 dist
 docs
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@
 #
 # =================================================================
 
-FROM ubuntu:jammy-20230425
+FROM ubuntu:jammy-20231211.1
 
 LABEL maintainer="Just van den Broecke <justb4@gmail.com>"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -102,7 +102,6 @@ ENV TZ=${TZ} \
     ${ADD_DEB_PACKAGES}"
 
 WORKDIR /pygeoapi
-ADD . /pygeoapi
 
 # Install operating system dependencies
 RUN \
@@ -120,22 +119,27 @@ RUN \
     && unzip ./SCHEMAS_OPENGIS_NET.zip "ogcapi/*" -d /schemas.opengis.net \
     && rm -f ./SCHEMAS_OPENGIS_NET.zip \
 
-    # Install remaining pygeoapi deps
-    && pip3 install -r requirements-docker.txt \
-    && pip3 install -r requirements-admin.txt \
-
-    # Install pygeoapi
-    && pip3 install -e . \
-
-    # Set default config and entrypoint for Docker Image
-    && cp /pygeoapi/docker/default.config.yml /pygeoapi/local.config.yml \
-    && cp /pygeoapi/docker/entrypoint.sh /entrypoint.sh  \
-
     # Cleanup TODO: remove unused Locales and TZs
     && apt-get remove --purge -y gcc ${DEB_BUILD_DEPS} \
     && apt-get clean \
     && apt autoremove -y  \
     && rm -rf /var/lib/apt/lists/*
+
+ADD requirements-docker.txt requirements-admin.txt /pygeoapi/
+# Install remaining pygeoapi deps
+RUN pip3 install -r requirements-docker.txt \
+    && pip3 install -r requirements-admin.txt
+
+
+ADD . /pygeoapi
+
+ # Install pygeoapi
+RUN pip3 install -e . 
+
+RUN \ 
+    # Set default config and entrypoint for Docker Image
+    cp /pygeoapi/docker/default.config.yml /pygeoapi/local.config.yml \
+    && cp /pygeoapi/docker/entrypoint.sh /entrypoint.sh 
 
 ENTRYPOINT ["/entrypoint.sh"]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,14 +127,14 @@ RUN \
 
 ADD requirements-docker.txt requirements-admin.txt /pygeoapi/
 # Install remaining pygeoapi deps
-RUN python3 -m pip install -r requirements-docker.txt \
-    && python3 -m pip install -r requirements-admin.txt
+RUN python3 -m pip install --no-cache-dir -r requirements-docker.txt \
+    && python3 -m pip install --no-cache-dir -r requirements-admin.txt
 
 
 ADD . /pygeoapi
 
  # Install pygeoapi
-RUN python3 -m pip install -e . 
+RUN python3 -m pip install --no-cache-dir -e . 
 
 RUN \ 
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -1,6 +1,7 @@
 .git
 .github
 .pytest_cache
+.mypy_cache
```
doublebyte1, line 4: No harm in this being here, but I think it may be specific to your system

Dockerfile
```diff
@@ -33,7 +34,7 @@
 #
 # =================================================================
 
-FROM ubuntu:jammy 
+FROM ubuntu:jammy-20230425
```
doublebyte1, line 37: I think it is a good idea to pin down the version for traceability

Dockerfile
```diff
@@ -121,21 +120,28 @@ RUN \
...
     # Cleanup TODO: remove unused Locales and TZs
+    # NOTE: this tries to remove gcc, but the actual package gcc-11 can't be
+    #       removed because python3-scipy depends on python3-pythran which
+    #       depends on g++
     && apt-get remove --purge -y gcc ${DEB_BUILD_DEPS} \
     && apt-get clean \
     && apt autoremove -y  \
     && rm -rf /var/lib/apt/lists/*
 
+ADD requirements-docker.txt /pygeoapi
```
doublebyte1, line 132: You can avoid this line, by moving line 136 here

doublebyte1 (commented):
Thank you for your contributions @totycro totycro  In my machine this indeed reduced the size from 1.2 GB to 1.18 GB and it is faster to build. However, I think the biggest benefit when making small changes, since you split the long lines into multiple commands.

Dockerfile
totycro: Thanks for the review!

Yes, this would be possible. However then any change in the pygeoapi source would lead to a reinstall of the pip packages.

The use case this optimizes for is when some development happens and the code is changed while `requirements-docker.txt` stays the same. Then a rebuild of the docker image is extremely fast.

doublebyte1 (approved):

francbartoli (commented):
@totycro I'm just wondering if it makes sense to move your changes to a dedicated `Dockerfile.develop`. If those are only relevant for development purposes then I wouldn't keep them in the image for the releases. In the `Dockerfile` I would expect improvements and optimisations for production

totycro:
> @totycro I'm just wondering if it makes sense to move your changes to a dedicated `Dockerfile.develop`. If those are only relevant for development purposes then I wouldn't keep them in the image for the releases. In the `Dockerfile` I would expect improvements and optimisations for production

There are also some benefits for production in the sense that newer versions will share most of the layers of the docker image, so pulling new images during an update is faster.

Generally i like the idea of having development and production systems be as similar as possible to make sure that the code that is being developed actually runs the same way in production. Currently some tests fail in docker (https://github.com/geopython/pygeoapi/issues/1111), so here it's not clear if running the docker image in production would have the same issues.

However if there are some reasons to have a separate `Dockerfile.develop` I'm not really against that either. It would mean some more maintenance overhead for the future, but one reason for it could be to install the test dependencies by default.

Dockerfile
francbartoli: @totycro @doublebyte1 shall we have a multi stage build and two dockerfiles, one for development and one for production?

francbartoli (changes requested):

Dockerfile
```diff
@@ -102,12 +103,10 @@ ENV TZ=${TZ} \
     ${ADD_DEB_PACKAGES}"
 
 WORKDIR /pygeoapi
-ADD . /pygeoapi
 
 # Install operating system dependencies
 RUN \
     apt-get update -y \
-    && apt-get upgrade -y \
```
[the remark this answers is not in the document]
totycro: I think in order to get the latest versions of libraries in containers, the idea is to use the latest base image directly. If updates happen here, then both the old and the new library are included in the image which can make it a lot larger depending on how many updates happen here.

Also when image caches are used, you never know which library versions are being used because an image built with this instruction can be cached for a long time.

There's also the issue of reproducibility, because when an updated package is available, it will be used on the next image build. So it can happen that we verify that pygeoapi is working fine with a certain set of libraries, then we need to release a hotfix, but we can't because a new library version is pulled in here which could create an incompatibility.
This is something that can be avoided by selecting the base image version manually at a time of our choosing.

I realize that going with using the latest image and also doing the upgrades here is a feasible path forward, so i'd also be fine with that if there are good reasons for that.

Dockerfile
totycro: My preference would be to stick to 1 Dockerfile as i've argued here: https://github.com/geopython/pygeoapi/pull/1252#issuecomment-1602187908

Having 2 Dockerfiles would also work, but with some more maintainence overhead. I'm not sure how multi-stage build could be used in a convenient way here.

Dockerfile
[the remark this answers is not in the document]
kalxas: I think we should be using the latest official image available and not run apt upgrade ourselves.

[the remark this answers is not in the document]
francbartoli: @kalxas I have the impression that taking only the latest official image doesn't preserve us to get vulnerable python packages. This might be fixed with the apt upgrade to get the remediated version of a package if any

Dockerfile
tomkralidis: Agree for a single Dockerfile.

tomkralidis:
- +1 to NOT `apt-get upgrade` in lieu of updating the base image
- +1 to keeping a single `Dockerfile`

Will merge later today if there is no further discussion.

tomkralidis (approved):

Dockerfile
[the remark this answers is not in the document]
tomkralidis: I think the onus is on the base image to be updated accordingly.

doublebyte1:
> * +1 to NOT `apt-get upgrade` in lieu of updating the base image
> * +1 to keeping a single `Dockerfile`
> 
> Will merge later today if there is no further discussion.

I agree to not using `apt-get upgrade`, which can bring unpredictable changes.

@francbartoli nothing against having two separate Dockerfiles, apart from the fact that we have to maintain both. Would we then be generating two different docker images?

tomkralidis:
@totycro can you resolve the merge conflicts?

totycro:
It's rebased now, and I've also set the base image to the latest ubuntu jammy version

PR #1277, opened 2023-06-22
Process manager now knows about existing processes

### LINKED ISSUE #1213: process manager to be given config of known processes during initialization

ricardogsilva:
# Overview

This PR makes the process manager aware of existing processes, as discussed in #1213.

It introduces two main changes:

- There is now a `processes` property on the `BaseManager` class. This property holds the configuration of known processes and can be referred to whenever there is a need to know which processes exist.

- The `get_manager()` function is introduced as the way to instantiate the process manager. This function accepts the main pygeoapi config dict as an input and is able to instantiate the process manager and provide it with the configuration of any processes defined in the config dict.

With this PR, the manager is responsible for letting the rest of pygeoapi codebase know which processes exist. The base implementation still reads processes from the main pygeoapi config file, but this is now an implementation detail, which means that custom managers could use an alternative method for specifying processes, if needed.

# Related Issue / Discussion

- fixes #1213 

# Contributions and Licensing

### DIFF
--- a/pygeoapi/process/manager/__init__.py
+++ b/pygeoapi/process/manager/__init__.py
@@ -1,8 +1,10 @@
 # =================================================================
 #
 # Authors: Tom Kralidis <tomkralidis@gmail.com>
+#          Ricardo Garcia Silva <ricardo.garcia.silva@geobeyond.it>
 #
 # Copyright (c) 2019 Tom Kralidis
+#           (c) 2023 Ricardo Garcia Silva
 #
 # Permission is hereby granted, free of charge, to any person
 # obtaining a copy of this software and associated documentation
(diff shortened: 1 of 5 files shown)


pygeoapi/process/manager/__init__.py
```diff
@@ -26,3 +28,36 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 # =================================================================
+
```
francbartoli: I see that other base classes doesn't use to put something in the `__init__`. Maybe, just for consistency, we can move this into the `base.py` file

francbartoli (approved):

PR #1278, opened 2023-06-22
Process manager instantiates processor

### LINKED ISSUE #1201: job manager to be responsible for creating an instance of a process

ricardogsilva:
# Overview

**NOTE: This PR builds upon #1277 - Please review it first**

This PR implements the `get_processor()` method on the `BaseManager` class, thus making the manager responsible for instantiating a processor, as discussed in #1201. The base manager still calls `load_plugin` internally, but this is now an implementation detail, which means that custom managers could use an alternative method for loading processes, if needed. 

# Related Issue / Discussion
- Fixes #1201 

# Contributions and Licensing

### DIFF
--- a/tests/test_manager.py
+++ b/tests/test_manager.py
@@ -1,3 +1,31 @@
+# =================================================================
+#
+# Authors: Ricardo Garcia Silva <ricardo.garcia.silva@geobyond.it>
+#
+# Copyright (c) 2023 Ricardo Garcia Silva
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation
+# files (the "Software"), to deal in the Software without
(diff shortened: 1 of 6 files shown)


ricardogsilva:
@tomkralidis ready for review now

tests/test_manager.py
```diff
@@ -1,8 +1,39 @@
+# =================================================================
+#
+# Authors: Ricardo Garcia Silva <ricardo.garcia.silva@geobyond.it>
```
francbartoli, line 3: There is a typo in the email

francbartoli (changes requested):

francbartoli (approved):

PR #1281, opened 2023-06-22
Guard against invalid job id throwing error

### LINKED ISSUE #1279: Asking for details of invalid job id throws an uncatched error
**Description**
When trying to get details about an invalid job, pygeoapi errors out with an uncatched error and throws `HTTP 500`, returning the full traceback to the client. 

This happens with both:

- `DummyManager` (the default) - Errors with `NotImplementedError`
- `TinyDBManager` - Errors with `TypeError`

Morevoer, the other core manager, `MongoDBManager`, just returns `False` when it cannot find a job - which also makes pygeoapi error out with an uncaught `TypeError`.

**Steps to Reproduce**

1. Configure pygeoapi to use any of the managers listed above
2. Stand up the pygeoapi development web server
    ```shell
    pygeoapi serve
    ```
4. Make a request to get details about a non-existing job, for example:
    ```shell
    curl localhost:5000/jobs/fake
    ```

**Expected behavior**
I would expect pygeoapi to respond with `HTTP 404` with a well-formatted exception, as defined in [OGC API - Processes, section 7.12.3](https://docs.ogc.org/is/18-062r2/18-062r2.html#toc33)

ricardogsilva:
# Overview

**NOTE: This PR builds upon #1278 - please review it first**

This PR enhances some API methods to catch possible errors thrown by the process manager. It deals mostly with two kinds of errors:

- when the input `job_id` does not refer to a known job
- when a job's result cannot be retrieved

Implementation consists in adding some custom exception classes in the `process.exceptions` module and then:

- have the process manager raise these errors
- document the `BaseManager` class that there is a requirement of raising these errors specifically
- have the outer `API` methods try to call a manager method and guard against the possible errors

As an example:

- when `API.get_jobs()` calls into `manager.get_job(job_id)`, the manager will raise `exceptions.JobNotFound`
- `API` thus guards this call to `manager.get_job(job_id)` with a `try..except` clause and handles the error appropriately - in this case it would send back a well-formatted exception response to the client

As a side-effect of this error catching method, the `API.delete_job` is made similar to the other job-related methods as in it now includes a `request` parameter too.

Moreover, this PR includes fixes for core managers:

- `DummyManager`
- `TinyDBManager`
- `MongoDBManager`

# Related Issue / Discussion
- fixes #1279 

# Contributions and Licensing


pygeoapi/process/manager/__init__.py
```diff
@@ -1,8 +1,10 @@
 # =================================================================
```
francbartoli, line 1: Maybe this is missing a rebase from a previous merge @ricardogsilva, can you have a look?

francbartoli (approved):

pygeoapi/process/exceptions.py
tomkralidis, line 1: Remove this file and place exceptions in `pygeoapi/process/base.py`.

tomkralidis (changes requested):

francbartoli (approved):

Issue #1286, opened 2023-06-25
Vector Tiles Issues


doublebyte1:
**Description**
I am going to summarize here some issues related to the publication of vector tiles, using OGC API - Tiles.
1 - ElasticSearch Tiles does not provide tile metadata. As a workaround we are using TilesMetadataFormat.NONE, but this is causing some issues because the tileset links are not resolving to anything.
2 - Even, when we have tilemetadata, the tileset links are missing links with self and alternate representation of the datasets, like we have [here](http://localhost:5000/collections/lakes/tiles/WorldCRS84Quad?f=json).

**Steps to Reproduce**
1. Access [this](https://emotional.byteroad.net/collections/hex350_grid_cardio_1920/tiles?f=json) page and try to resolve any of the tilesetlinks. An internal server error will be thrown.
2. Access [this](https://demo.pygeoapi.io/master/collections/lakes/tiles?f=json) page. From there, navigate to [one](https://demo.pygeoapi.io/master/collections/lakes/tiles/WorldCRS84Quad?f=json) of the tilesetlinks.

**Expected behavior**
1. We should see the tileset representation.
2. We should see the links for the different representations of this resource.

**Environment**
 - OS: Ubuntu 22.04.2 LTS
 - Python version: 3.7.5
 - pygeoapi version: 0.16.dev0 (master build on 24/07/23

**Additional context**
It is possible to test this issues using `gdal/ogr`. 
1. `./ogrinfo --config API TILES --config CPL_DEBUG ON "OGCAPI:https://demo.pygeoapi.io/master/collections/lakes"`
2. `ogrinfo --config API TILES --config CPL_DEBUG ON "OGCAPI:https://emotional.byteroad.net/collections/hex350_grid_cardio_1920"
`
Here we get the right response:`
./ogrinfo --config API TILES --config CPL_DEBUG ON "OGCAPI:https://maps.gnosis.earth/ogcapi/collections/Daraa:AgricultureSrf"`

`GDAL version: GDAL 3.8.0dev-360a9aea02, released 2023/06/24 (debug build)

Also, see this issue: https://github.com/OSGeo/gdal/issues/7970#issuecomment-1605687527

doublebyte1:
I think one option to solve issue #1, could be to attempt to always attempt to generate tilesetmetadata with default values, even when we don't have a tilejson file (or other) from the provider.

doublebyte1:
[...]

PR #1288, opened 2023-06-26
Fixing error when retrieving the TMS in vector tiles that do not provide tilejson metadata

doublebyte1:
# Overview
When generating the tiles metadata, pygeoapi is looking for a tilejson file. This happens, even when we don't use the `tilejson` metadata format.
Other tile providers like `elasticsearch` and `pg_tiles` do not provide a metadata.json file, and that is ok, because it is a tippecanoe specific thing. For that reason, we support a `default` metadata format.

# Related Issue / Discussion
This problem is summarized in this issue: https://github.com/geopython/pygeoapi/issues/1286 (#1)

# Additional Information
Now that the `default` metadata format is supported for any provider, we could get rid of the TilesMetadataFormat.NONE format.

# Contributions and Licensing


doublebyte1:
I think that the failing test is unrelated to this PR

francbartoli (approved):

PR #1306, opened 2023-07-03
Install python3-filelock via deb package

totycro:
# Overview

Improvement as suggested in previous PR here: https://github.com/geopython/pygeoapi/pull/1273#pullrequestreview-1505315366

Now filelock will be installed as debian package as a debian requirement and also in the Dockerfile. It's still in the requirements.txt, but pip detects that the requirement is already satisfied.

Is this the desired behavior? It's reasonable to me at least :)

# Contributions and Licensing


totycro:
Closing since tinydb is not in scope for debian.

PR #1332, opened 2023-08-09
Echo process

xcomagent95:
# Overview

In our attempts to develop a pygeoapi instance which can be successfully tested against the test suit for the OGC API - Processes - Part 1: Core standard, we also developed a testable echo process for the pygeoapi. As the TEAM engine suggests, it supports the pause parameter, which enables a user to influence the processing time. 

# Related Issue / Discussion

Enhanced conformance of the pygeoapi towards the OGC API - Processes - Part 1: Core standard.

# Additional Information

The process is closely related to the exemplary echo process described in the OGC API - Processes - Part 1: Core standard document. It accepts String input and echoes that value back. A float value may be passed optionally by the user to influence the processing time. If a pause value is passed the process execution will be halted for the specified number of seconds.

# Contributions and Licensing

- [ ] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.

### DIFF
--- a/pygeoapi/process/echo.py
+++ b/pygeoapi/process/echo.py
@@ -83,14 +83,13 @@ PROCESS_METADATA = {
       }
     }
   },
-  "links": [
-    {
-      "href": "https://processing.example.org/oapi-p/processes/echo/execution",
-      "rel": "http://www.opengis.net/def/rel/ogc/1.0/execute",
-      "title": "Execute endpoint",
-      "type": "endpoint"
-    }
-  ],
+  "links": [{
+        'type': 'text/html',
+        'rel': 'about',
+        'title': 'information',
+        'href': 'https://example.org/process',
+        'hreflang': 'en-US'
+    }],
   'example': {
     'inputs': {
       'echo': 'echoValue',
--- a/pygeoapi/process/echo.py
+++ b/pygeoapi/process/echo.py
@@ -1,0 +1,140 @@
+# =================================================================
+#
+# Authors: Alexander Pilz <a.pilz@52north.org>
+#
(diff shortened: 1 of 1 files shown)


tomkralidis:
Thanks @xcomagent95.  Out of interest, what are the CITE results as a result of this PR?  We would like to update our demo server in pursuit of compliance and this PR will help greatly, then.

xcomagent95:
> Thanks @xcomagent95. Out of interest, what are the CITE results as a result of this PR? We would like to update our demo server in pursuit of compliance and this PR will help greatly, then.

We are happy to hear that our work might be useful to you. We just ran a conformance test for our test instance of the pygeoapi which features the echo process in question. Here only two test cases are marked as failed, namely the _Job Creation Input Validation_ and the _Job Creation Input Ref_ test. I have been in contact with Benjamin Pross who is involved in the implementation of the test suite. We came to the conclusion that the cause of the failures is likely to be a bug in the test suite. The team is already working towards a fix (https://github.com/opengeospatial/ets-ogcapi-processes10/issues/54). Fell free to have a look at the test results attached. 

[Conformance test results.zip](https://github.com/geopython/pygeoapi/files/12344897/Conformance.test.results.zip)

tomkralidis:
Thanks @xcomagent95!  For reference, note that there is a branch in https://github.com/geopython/pygeoapi/tree/oaproc-cite that I'm targeting for official compliance.  This branch (based on current master) should land in master branch in October 2023 at the upcoming OGC API sprint.

Issue #1335, opened 2023-08-17
"Install in 5 minutes" Results in TypeError


mlinds:
**Description**
I am getting a Python TypeError with the message "Field() missing 1 required positional argument: 'default'"

**Steps to Reproduce**
I followed the "Install in 5 minutes" instructions found [here](https://pygeoapi.io/).

```shell
python3 -m venv pygeoapi
cd pygeoapi
 . bin/activate
git clone https://github.com/geopython/pygeoapi.git
cd pygeoapi
pip3 install -r requirements.txt
python3 setup.py install
cp pygeoapi-config.yml example-config.yml
export PYGEOAPI_CONFIG=example-config.yml
export PYGEOAPI_OPENAPI=example-openapi.yml
pygeoapi openapi generate $PYGEOAPI_CONFIG --output-file $PYGEOAPI_OPENAPI
```

**Expected behavior**
The openAPI file should be generated

**Screenshots/Tracebacks**

The error occurs on the step `pygeoapi openapi generate $PYGEOAPI_CONFIG --output-file $PYGEOAPI_OPENAPI`
```
Traceback (most recent call last):
  File "/home/mlinds/miniconda3/bin/pygeoapi", line 5, in <module>
    from pygeoapi import cli
  File "/home/mlinds/miniconda3/lib/python3.10/site-packages/pygeoapi/__init__.py", line 40, in <module>
    from pygeoapi.config import config
  File "/home/mlinds/miniconda3/lib/python3.10/site-packages/pygeoapi/config.py", line 36, in <module>
    from pygeoapi.util import to_json, yaml_load
  File "/home/mlinds/miniconda3/lib/python3.10/site-packages/pygeoapi/util.py", line 73, in <module>
    from pygeoapi.models import config as config_models
  File "/home/mlinds/miniconda3/lib/python3.10/site-packages/pygeoapi/models/config.py", line 35, in <module>
    class APIRules(BaseModel):
  File "/home/mlinds/miniconda3/lib/python3.10/site-packages/pygeoapi/models/config.py", line 37, in APIRules
    api_version: str = Field(regex=r'^\d+\.\d+\..+$',
TypeError: Field() missing 1 required positional argument: 'default'
```

**Environment**
 - OS: Ubuntu 22.04.1 LTS (On WSL2
 - Python version: Python 3.10.8
 - pygeoapi version: latest from master branch

hiteshbedre:
[...]

PR #1336, opened 2023-08-17
Added default parameter for Field function

hiteshbedre:
# Overview

This pull request resolves the Python TypeError issue that occurs during the  pygeoapi openapi generate  step. The error is caused by a missing required positional argument  default  in the  Field()  function call. 

# Related Issue
https://github.com/geopython/pygeoapi/issues/1335

**Changes Made:** 
 
- Added the missing  default  argument to the  Field()  function call in the  pygeoapi/models/config.py  file. 
- Tested the solution on Ubuntu 22.04.1 LTS with Python 3.10.8. 

# Additional Information
 
- Followed the provided steps to reproduce the issue. 
- After applying the changes from this pull request, re-ran the  pygeoapi openapi generate  step. 
- Verified that the openAPI file was successfully generated without any errors. 

# Contributions and Licensing

- [x] I'd like to contribute bugfix to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.

### DIFF
--- a/pygeoapi/models/config.py
+++ b/pygeoapi/models/config.py
@@ -34,8 +34,11 @@ from pydantic import BaseModel, Field
 
 class APIRules(BaseModel):
     """ Pydantic model for API design rules that must be adhered to. """
-    api_version: str = Field(regex=r'^\d+\.\d+\..+$',
-                             description="Semantic API version number.")
+    api_version: str = Field(
+        "",
+        regex=r'^\d+\.\d+\..+$',
+        description="Semantic API version number."
+    )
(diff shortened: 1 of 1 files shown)


doublebyte1:
@hiteshbedre thank you for looking into this, but as mentioned [here](https://github.com/geopython/pygeoapi/issues/1335), it does not seem code related. I don't experience any issues on Ubuntu 22.04. I am closing this PR, as the issue that it is addressing is also closed.

Issue #1302, opened 2023-07-02
Allow more flexible handling of external dependencies in templates


xamanu:
**Is your feature request related to a problem? Please describe.**

The software currently creates a potential privacy breach by fetching data from an external website at runtime. This happens usually in the templates pulling in javascript, css and font files from `https://unpkg.com` and Google.

**Describe the solution you'd like**

It would be great to be able to set the path to external libraries with a variable, in order to use provide the libraries with a relative path and being served from the same web-server.

**Additional context**

Output from lintian, while packaging it for Debian and derivatives:

```
E: python3-pygeoapi: privacy-breach-uses-embedded-file You may use the libjs-bootstrap package. (https://unpkg.com/bootstrap@5.1.3/dist/css/bootstrap.min.css) [usr/lib/python3/dist-packages/pygeoapi/templates/_base.html]
E: python3-pygeoapi: privacy-breach-uses-embedded-file You may use the node-html5shiv package (virtual package). (https://cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7.3/html5shiv.js) [usr/lib/python3/dist-packages/pygeoapi/templates/_base.html]
W: python3-pygeoapi: privacy-breach-generic [<link href="https://fonts.googleapis.com/css?family=montserrat:300,400,700|roboto:300,400,700" rel="stylesheet">] (https://fonts.googleapis.com/css?family=montserrat:300,400,700|roboto:300,400,700) [usr/lib/python3/dist-packages/pygeoapi/templates/openapi/redoc.html]
W: python3-pygeoapi: privacy-breach-generic [<link rel="icon" type="image/png" href="https://unpkg.com/swagger-ui-dist/favicon-16x16.png" sizes="16x16" />] (https://unpkg.com/swagger-ui-dist/favicon-16x16.png) [usr/lib/python3/dist-packages/pygeoapi/templates/openapi/swagger.html]
W: python3-pygeoapi: privacy-breach-generic [<link rel="icon" type="image/png" href="https://unpkg.com/swagger-ui-dist/favicon-32x32.png" sizes="32x32" />] (https://unpkg.com/swagger-ui-dist/favicon-32x32.png) [usr/lib/python3/dist-packages/pygeoapi/templates/openapi/swagger.html]
[...]

PR #1339, opened 2023-08-25
Get bbox query parameter interactively

MTachon:
# Overview

This PR implements a way for end-users to get interactively the `bbox` query parameter from a collection's HTML page. In order to get `bbox` coordinates, end-users can click and drag the mouse over a rectangular area of interest, while pressing the the `Ctrl` key. A popup appears with the `bbox` query parameter corresponding to the rectangular area drawn on the map. The popup's text can then be copied and pasted directly in the query URL.

# Additional Information

Several provider types may support an additional `bbox-crs` query parameter that specify which coordinate reference system the bbox's coordinates are in. In the [part 2 of the OGC API Features standard](https://docs.ogc.org/is/18-058r1/18-058r1.html#_parameter_bbox_crs), it says that "*If the bbox-crs parameter is not specified then the coordinate values of the bbox parameter SHALL be assumed to be in the default CRS specified in [OGC API - Feature - Part 1: Core](https://docs.ogc.org/is/18-058r1/18-058r1.html#OAFeat-1); that is http://www.opengis.net/def/crs/OGC/1.3/CRS84...*". It is likely that other OGC API standards will use the same assumption. This implementation allows to get interactively the `bbox` query parameter directly in the default CRS. The `bbox` can therefore be used without specifying any `bbox-crs`.
  
# Contributions and Licensing

### DIFF
--- a/pygeoapi/templates/collections/collection.html
+++ b/pygeoapi/templates/collections/collection.html
@@ -8,8 +8,8 @@
 {% endblock %}
 
 {% block extrahead %}
-    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.3.1/dist/leaflet.css"/>
-    <script src="https://unpkg.com/leaflet@1.3.1/dist/leaflet.js"></script>
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css"/>
+    <script src="https://unpkg.com/leaflet@1.7.1/dist/leaflet.js"></script>
 {% endblock %}
 
 {% block body %}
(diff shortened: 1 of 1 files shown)


pygeoapi/templates/collections/collection.html
```diff
@@ -8,8 +8,8 @@
 {% endblock %}
 
 {% block extrahead %}
-    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.3.1/dist/leaflet.css"/>
-    <script src="https://unpkg.com/leaflet@1.3.1/dist/leaflet.js"></script>
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css"/>
```
tomkralidis, line 11: Let's keep to 1.3.1 for consistency, until #1302 is addressed.

tomkralidis (changes requested):

MTachon:
CI issues are not related to PR.

Issue #1363, opened 2023-10-03
Close the elastic ports on the elastic docker example


doublebyte1:
**Is your feature request related to a problem? Please describe.**
The elastic example is [publishing the elastic container ports on the host](https://github.com/geopython/pygeoapi/blob/master/docker/examples/elastic/docker-compose.yml#L62). If run in production, this could open a vulnerability as the elastic server is unprotected. 

**Describe the solution you'd like**
In reality, we do not need to publish these ports, as the [pygeoapi container is accessing the elastic container directly](https://github.com/geopython/pygeoapi/blob/master/docker/examples/elastic/pygeoapi/docker.config.yml#L151). Therefore, we can just not publish these ports in the docker-compose file and everything will work as expected. For debugging purposes, maybe we can just leave that code there, but commented out, and perhaps add a warning that if someone uses it in production they will need to protect elastic.

krishnaglodha:
I'll work on it

melencholicmice:
Hello , is the issue still open ? if it is  I would look into it.

doublebyte1:
@melencholicmice yes, it is! Thank you for your interest 👍🏽

doublebyte1:
> Hello , is the issue still open ? if it is I would look into it.

I think that a PR already solved this issue. 
https://github.com/geopython/pygeoapi/pull/1382

@melencholicmice Sorry about that!

melencholicmice:
@doublebyte1  I understand, I'm Intrested  in contributing  to the project  so is there any other issues I can look into ?

Issue #1362, opened 2023-10-03
Remove datasets which are not served with elastic, from the elastic docker example


doublebyte1:
**Description**
In the elastic docker example, [pygeoapi is configured to serve 4 datasets](https://github.com/geopython/pygeoapi/blob/master/docker/examples/elastic/pygeoapi/docker.config.yml#L88), but only 1 of them is actually [using elastic as a backend provider](https://github.com/geopython/pygeoapi/blob/master/docker/examples/elastic/pygeoapi/docker.config.yml#L124).

**Steps to Reproduce**
Start the docker-compose file and navigate to the collections page. You will see 4 datasets there.

<img width="1440" alt="Screenshot 2023-10-03 at 15 37 50" src="https://github.com/geopython/pygeoapi/assets/1038897/5dfdf86b-2b7a-47b4-9d9f-402822c85723">

These datasets use different providers: CSV, GeoJSON, SQLiteGPKG. Only one of them uses elastic.

**Expected behavior**

In this example, we should publish only the dataset that is served using the elastic provider:

<img width="1440" alt="Screenshot 2023-10-03 at 15 36 43" src="https://github.com/geopython/pygeoapi/assets/1038897/98889862-26a3-4421-9760-9d7c3a6cd263">

In order to achieve this, you can edit the [pygeoapi configuration file](https://github.com/geopython/pygeoapi/blob/master/docker/examples/elastic/pygeoapi/docker.config.yml) to remove all the datasets that are not published using the elastic provider.

krishnaglodha:
On it

krishnaglodha:
Fixed in PR #1364

krishnaglodha:
<img width="1496" alt="Screenshot 2023-10-04 at 10 06 36 AM" src="https://github.com/geopython/pygeoapi/assets/47075664/21a70df7-c7cb-4237-8315-b352c47476a5">

doublebyte1:
Thanks @krishnaglodha! Please see my comment here: https://github.com/geopython/pygeoapi/pull/1364#pullrequestreview-1656981283

webb-ben:
Closing with examples fixed and moved to separate repository.

KilimAnnejaro:
Note: I'm thinking this issue belongs on <https://github.com/geopython/pygeoapi-examples>, rather than here, after <https://github.com/geopython/pygeoapi-examples/pull/1>.

PR #1364, opened 2023-10-04
closed ports of elastic docker

krishnaglodha:
# Related Issue / Discussion
PR against issue #1363 
# Additional Information

Ports are closed now, although I see `undefined` in the popup , should I create issue and fix this ? 

# Contributions and Licensing

- [x] I'd like to contribute #1363 to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.

### DIFF
--- a/docker/examples/elastic/docker-compose.yml
+++ b/docker/examples/elastic/docker-compose.yml
@@ -60,8 +60,8 @@ services:
     container_name: elastic
 
     ports:
-      - 9300:9300
-      - 9200:9200
+      - 9300
+      - 9200
     volumes:
       - elastic_search_data:/usr/share/elasticsearch/data
 
--- a/docker/examples/elastic/pygeoapi/docker.config.yml
+++ b/docker/examples/elastic/pygeoapi/docker.config.yml
@@ -86,41 +86,6 @@ metadata:
         role: pointOfContact
 
 resources:
-    obs:
-        type: collection
-        title: Observations
-        description: My cool observations
-        keywords:
-            - observations
-            - monitoring
-        links:
-            - type: text/csv
-              rel: canonical
-              title: data
-              href: https://github.com/mapserver/mapserver/blob/branch-7-0/msautotest/wxs/data/obs.csv
-              hreflang: en-US
-            - type: text/csv
-              rel: alternate
-              title: data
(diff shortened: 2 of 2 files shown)


PascalLike:
Thank you Krishna! But for the month of October, it is probably better to leave these hacktoberfest issues for people who are approaching pygeoapi for the first time. These are very suitable activities to get closer to the code base as ice breakers :pray:

docker/examples/elastic/docker-compose.yml
```diff
@@ -60,8 +60,8 @@ services:
     container_name: elastic
 
     ports:
-      - 9300:9300
```
doublebyte1, line 63: @krishnaglodha : this is not closing the port. It is assigning the port to a random port in the host.

docker/examples/elastic/pygeoapi/docker.config.yml
```diff
@@ -86,41 +86,6 @@ metadata:
         role: pointOfContact
 
 resources:
-    obs:
```
doublebyte1, line 89: @krishnaglodha Thank you for this, but this is actually addressing a different issue https://github.com/geopython/pygeoapi/issues/1362 . Please don't address it in this PR.

doublebyte1 (commented):
Thank you for contributing @krishnaglodha! 👍🏽 

It want to address two issues, it would be better to issue two different PR, each one with the code that addresses specifically that issue :-) . You could achieve that by using branches.

doublebyte1:
> Thank you Krishna! But for the month of October, it is probably better to leave these hacktoberfest issues for people who are approaching pygeoapi for the first time. These are very suitable activities to get closer to the code base as ice breakers 🙏

@krishnaglodha I agree :-) If you want to help during the #hacktoberfest, you could create issues, similar to the ones I did: very simple and self contained, for instance fixing a bug on the documentation, or something in the configuration. That would be really helpful! 🥇

docker/examples/elastic/pygeoapi/docker.config.yml
doublebyte1: @krishnaglodha Please don't do anything else and read the other comments from @PascalLike and myself.

krishnaglodha: Yes will not work on it

Issue #1365, opened 2023-10-04
Pydantic v1 is not supported anymore


francbartoli:
**Description**
When the latest pygeoapi python package is used as a library in third-party application where there is already a Pydantic dependency pinned at v1 the overall dependencies tree is broken

**Steps to Reproduce**
`poetry update` with `pyproject.toml``
```toml
[tool.poetry.dependencies]
pygeoapi = {git = "https://github.com/geopython/pygeoapi", rev = "master"}
pydantic = "1.10.6"
```

**Expected behavior**
Update is smooth and clean

**Screenshots/Tracebacks**
NA

**Environment**
 - OS: MacOS
 - Python version: 3.9
 - pygeoapi version: master

**Additional context**
NA

PR #1366, opened 2023-10-04
Pin pydantic dependency

francbartoli:
Fix https://github.com/geopython/pygeoapi/issues/1365

### DIFF
--- a/pygeoapi/provider/xarray_.py
+++ b/pygeoapi/provider/xarray_.py
@@ -439,10 +439,10 @@ class XarrayProvider(BaseProvider):
             if coord.lower() == 'time':
                 time_var = coord
                 continue
-            if self._data.coords[coord].attrs['units'] == 'degrees_north':
+            if self._data.coords[coord].attrs.get('units') == 'degrees_north':
                 y_var = coord
                 continue
-            if self._data.coords[coord].attrs['units'] == 'degrees_east':
+            if self._data.coords[coord].attrs.get('units') == 'degrees_east':
                 x_var = coord
                 continue
 


kalxas:
pydantic 2.0 is not yet available in Debian, Ubuntu, Redhat, SUSE distros.

francbartoli:
It is in preparation for Debian https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1052619. It is instead available for Fedora https://pkgs.org/download/python-pydantic-core

kalxas:
Can we revert to pydantic 1.x until 2.x is generally available in the distros?

kalxas:
Perhaps we could offer a pydantic 2.x package in UbuntuGIS once it is included in Debian.

ricardogsilva:
@francbartoli 

We would actually need to pin pydantic with a version >=2.0.2 in  because this is the version that renamed `regex` to `pattern`, which pygeoapi is using in

https://github.com/geopython/pygeoapi/blob/2f33ff5d15367fcc14e706fb07136984a3df228a/pygeoapi/models/config.py#L37

@kalxas @tomkralidis 

I think this situation is a symptom of the overall issue that pygeoapi does not really employ a strategy to ensure reproducible installs when used standalone - we should discuss this elsewhere, but IMO a solution would be to:

1. Pin versions that are known to work in the `requirements.txt` files - these files would be used for when pygeoapi is installed in standalone mode (i.e. not as a library);
2. Do not populate the `setup.py:setup`'s `packages=` section automatically from the contents of the requirements files, but rather ensure the dependencies fed to setup.py are not pinned - this allows more freedom with dependencies when using pygeoapi as a library, which is @francbartoli 's use case here
3. Be more conservative with modifying the pygeoapi codebase to comply with newer versions of our dependencies, in order to avoid this awkward situation where pydantic v2 is now required by pygeoapi but is not available on the debian repos, which is one of our target installation methods

tomkralidis:
Can we be conservative by making the pydantic code work with 1.x?

Having said this, we had a dependency policy meeting this past week, and there will be an RFC soon for review/comment.

huard:
Hi, just got tripped by this. Installed pygeoapi with pip and got the `regex` is removed error. Fixed by downgrading to pydantic 1.10.

tomkralidis:
Thanks @huard.  We will be discussing next steps on this issue next week, at which point we will propagate a release with the fix which is affecting numerous downstream applications.

pvgenuchten:
this relates to the problem: 

```
raise TypeError("To define root models, use `pydantic.RootModel` rather than a field called '__root__'")
TypeError: To define root models, use `pydantic.RootModel` rather than a field called '__root__'
```

or is a different one?

Issue #1367, opened 2023-10-04
removal of dependency datamodel-code-generator


ricardogsilva:
Pygeoapi currently has a dev dependency on [datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator). This third-party dependency is never used by pygeoapi at runtime. Moreover, it was only used when autogenerating the models in [pygeoapi/models/cql.py](https://github.com/geopython/pygeoapi/blob/96cf72236a771d33cd61b8d14e0a7c0e62714db8/pygeoapi/models/cql.py#L3).
Since its creation, this file that contains the autogenerated models has even had some manual modifications, meaning it has diverged from the automated generation process - thus making it harder to ever regenerate it, if needed.

Keeping the dependency on `datamodel-code-generator` also has the downside that it actually prevents pygeoapi from moving to a newer version of the `jsonschema` package[1] - due to this dependency, pygeoapi's jsonschema is stuck at v4.17.3, and this is the last version to support Python 3.7, which pygeoapi still aims to support (however, check #1360 for more info).

As such, I am proposing that `datamodel-code-generator` be removed from pygeoapi's `requirements-dev.txt` file.

[1] - `datamodel-code-generator` depends on `openapi-schema-validator` and on `openapi-spec-validator`, both of which are not used by pygeoapi, and both of which mandate `jsonschema < 4.18.0. The current version of `jsonschema` is v4.19.1

ricardogsilva:
Completed by merging #1370

Issue #1360, opened 2023-10-02
Upgrade minimum Python version to 3.8


ricardogsilva:
This is related to the discussion in #1176

Now that we are [three months past Python 3.7's EOL](https://endoflife.date/python), is probably a good time to drop it from pygeoapi's CI and move our minimum Python version up to Python 3.8. which seems to be the most conservative version to keep support for nowadays.

My main motivation for this is I'd like to do some work related to validation of pygeoapi input/output with the corresponding OGC schemas, using jsonschema validation - including adding support for linked schema files (_i.e._ those `$ref` properties that show up all over the OGC schemas). 
There seems to be a nice [referencing](https://github.com/python-jsonschema/referencing) package made by the folks of python-jsonschema and I'm keen to study how to integrate this into pygeoapi. This third-party package unfortunately only supports Python 3.8+.

francbartoli:
+1, also for security reason it's useful to move

doublebyte1:
@ricardogsilva  I think we may need to adjust the code with the upgrade.  I noticed that for instance the unit tests don't work with python 3.8.

ricardogsilva:
@doublebyte1 I'm looking into failed unit tests to try and get them in shape for bumping Python version to 3.8

ricardogsilva:
With the recent merge of #1399 this is completed

PR #1370, opened 2023-10-06
Removed datamodel-code-generator from dev requirements

ricardogsilva:
This PR removes `datamodel-code-generator` from dev requirements, following the discussion from #1367 

# Related Issue / Discussion
- #1367

# Contributions and Licensing

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -23,39 +23,6 @@ To run all tests, simply run ``pytest`` in the repository.  To run a specific te
 run ``pytest tests/test_api.py``, for example.
 
 
-CQL extension lifecycle
------------------------
-
-Limitations
-^^^^^^^^^^^
-
-This workflow is valid only for the `CQL-JSON` format.
-
-Schema
-^^^^^^
-
-The Common Query Language (CQL) is the part 3 of the standard OGC API - Features. This extension has its specification available at 
-`OGC API - Features - Part 3: Filtering and the Common Query Language (CQL) <https://portal.ogc.org/files/96288>`_ and the schema exists in development at
-`cql.json <https://portal.ogc.org/files/96288#cql-json-schema>`_.
-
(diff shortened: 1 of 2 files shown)


ricardogsilva:
@francbartoli

docs/source/development.rst
```diff
@@ -23,39 +23,6 @@ To run all tests, simply run ``pytest`` in the repository.  To run a specific te
 run ``pytest tests/test_api.py``, for example.
 
 
-CQL extension lifecycle
```
francbartoli, line 26: Why are we removing this entire section? Maybe it's better for DX to remove only `Model generation` and `How to merge` and add a paragraph `How to hack the models`, what do you think?

francbartoli (changes requested):

docs/source/development.rst
ricardogsilva: What do you envision as being the content of this alternative `How to hack the models` section? 

In my opinion it would likely make sense if you'd still like to recommend a workflow like the autogeneration of models using `datamodel-code-generator` and then some manual tweaking. But I don't see much value in it, as the models already exist and have modifications, so it seems more likely that further changes will not restart from scratch - in which case, it seems to me that _How to hack the models_ is not so different from _how to develop pygeoapi as a whole_.

But maybe I'm missing the point :sweat_smile: ?

ricardogsilva:
@francbartoli can I ask for more clarification on what would be the way to move this forward? I'm not really sure what I would replace the deleted docs content with.

On the other hand I'd like to move on with the work needed for #1360 and this PR is a blocker for that

francbartoli:
@ricardogsilva I would consider relevant to keep where this model applies and it is limited to only `CQL-JSON` schema for the ES provider.

francbartoli (approved):

PR #1394, opened 2023-11-08
remove ajax call to fetch jsonld representation

pvgenuchten:
# Overview

the jsonld reference is duplicated, this is the non functional one

# Related Issue / Discussion

see discussion in to https://github.com/geopython/pygeoapi/pull/916

# Contributions and Licensing

### DIFF
--- a/pygeoapi/templates/_base.html
+++ b/pygeoapi/templates/_base.html
@@ -102,22 +102,5 @@
         style="height:24px;vertical-align: middle;" /></a> {{ version }}</footer>
     {% block extrafoot %}
     {% endblock %}
-    <script>
-      // Requests and embeds JSON-LD representation of current page
-      var xhr = new XMLHttpRequest();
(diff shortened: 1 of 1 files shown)


webb-ben:
I might be missing something here, but as it stands this PR removes all linked-data from the HTML pages. 

Directly inserting the JSON-LD halves the server processing for each request, but I don't think it is necessarily a bug. An AJAX request improves the SEO of default templates by optimizing the human-readable version of the content and delegating the machine-readable content to a separate request. 

The upside is that the way linked-data fits into pygeoapi HTML pages is customizable based on the jinja2 templates used. Perhaps figuring out a better way to offer direct insertion OR AJAX request linked-data sections of the <head> is worth being flushed out?

pvgenuchten:
Thanx for review benn, I’ll give it a look, removal is certainly not intended

pvgenuchten:
you are correct @webb-ben, this removes the relevant script include, this is not intended, closing

PR #1399, opened 2023-11-11
Prepare code and CI for Python 3.8

ricardogsilva:
# Overview
This PR refactors some of the tests (and code) in order to make them pass when running under Python 3.8 and also modifies the main CI workflow to use both Python v3.7 and v3.8

Ideally, it should verify that the pygeoapi is ready to move on to Python 3.8 and drop Python 3.7. This PR does not drop Python 3.7 from CI though - it merely prepares the code for it

# Related Issue / Discussion
- #1360 

# Additional Information

Some notes about the PR changes:

- In order to better identify failing tests I have modified some test functions to make use of [pytest's parametrize feature](https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions). This allows having more granular test cases and helps identify exactly which part of a large test is failing. However, note that **I did not change any test logic** (if you find some logic that has changed, then it is a bug in this PR)
- Specifically with regard to the test data file in `tests/data/CMC_hrdps_continental_TMP_TGL_80_ps2.5km_2020102700_P005-00.grib2`, which is for the `gdps-temperature` collection in the test configuration, I did change pygeoapi config in order to remove the custom GDAL creation option `DATA_ENCODING: COMPLEX_PACKING` - the previous value caused a rasterio error under Python 3.8. Additionally, by consulting the [data encoding section of the GDAL documentation on the GRIB driver](https://gdal.org/drivers/raster/grib.html#data-encoding), this option seemed to not be relevant, as the specific file being used as a test sample does not have a NO_DATA value

# Contributions and Licensing

### DIFF
--- a/pygeoapi/provider/xarray_edr.py
+++ b/pygeoapi/provider/xarray_edr.py
@@ -140,7 +140,8 @@ class XarrayEDRProvider(BaseEDRProvider, XarrayProvider):
                                 query_params[self.time_field]
                         )
                     }
-                data = data.sel(time_query).sel(remaining_query,method='nearest')
+                data = data.sel(
+                    time_query).sel(remaining_query, method='nearest')
             else:
                 data = data.sel(query_params, method='nearest')
         except KeyError:
--- a/pygeoapi/provider/xarray_edr.py
+++ b/pygeoapi/provider/xarray_edr.py
@@ -30,7 +30,6 @@
 import logging
 
 import numpy as np
-import pandas as pd
 
 from pygeoapi.provider.base import ProviderNoDataError, ProviderQueryError
 from pygeoapi.provider.base_edr import BaseEDRProvider
@@ -268,7 +267,7 @@ class XarrayEDRProvider(BaseEDRProvider, XarrayProvider):
                 LOGGER.debug('Reversing slicing from high to low')
                 return slice(end, begin)
         else:
-            return pd.Timestamp(datetime_)
+            return np.datetime64(datetime_)
 
     def _get_time_range(self, data):
         """
--- a/pygeoapi/provider/xarray_edr.py
+++ b/pygeoapi/provider/xarray_edr.py
@@ -30,6 +30,7 @@
 import logging
 
 import numpy as np
+import pandas as pd
 
 from pygeoapi.provider.base import ProviderNoDataError, ProviderQueryError
 from pygeoapi.provider.base_edr import BaseEDRProvider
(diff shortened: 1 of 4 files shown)


francbartoli (approved):

pygeoapi/provider/xarray_edr.py
```diff
@@ -252,7 +261,7 @@ def _make_datetime(self, datetime_):
                 LOGGER.debug('Reversing slicing from high to low')
                 return slice(end, begin)
         else:
-            return datetime_
+            return pd.Timestamp(datetime_)
```
tomkralidis, line 264: Are we able to do this without pandas here?

tomkralidis (changes requested):
Thanks @ricardogsilva -- see minor comment about pandas.  In addition:

- can we put the pytest changes as part of another PR?
- +1 to remove 3.7 support as part of this PR

pygeoapi/provider/xarray_edr.py
ricardogsilva: I'll investigate a bit further. The reason I did it this way is because [xarray's docs mention how its datetime indexing borrwos a lot from pandas](https://docs.xarray.dev/en/stable/user-guide/time-series.html#datetime-indexing) and also because pandas was already listed as a requirement in `requirements-provider.txt`. Maybe I can just use some analogous numpy datatype to get the same result, I'll check it out

ricardogsilva:
>     * can we put the pytest changes as part of another PR?

Do you mean the usage of oytest's `parametrize`? Sure I can do that - I included it here because they made my life easier, as I could isolate the specific test case(s) that were failing and rerun - I can revert to a non-parametrized version if you prefer

tomkralidis:
> > ```
> > * can we put the pytest changes as part of another PR?
> > ```
> 
> Do you mean the usage of oytest's `parametrize`? Sure I can do that - I included it here because they made my life easier, as I could isolate the specific test case(s) that were failing and rerun - I can revert to a non-parametrized version if you prefer

Yes that would work, thanks.

pygeoapi/provider/xarray_edr.py
ricardogsilva: tests pass by replacing `pd.Timestamp` with `np.datetime64`, so I switched to numpy

ricardogsilva:
@tomkralidis 

- [x] Reverted tests back to their previous form
-  [x] Dropped Python 3.7 from CI file
- [x] Switched `pdTimestamp` usage to more humble `np.datetime64`

tomkralidis (approved):

PR #1411, opened 2023-11-28
Logfile rotation and configurable log format

### LINKED ISSUE #1388: Configurable log format
Hello,

for my customer, who are using Pygeoapi, we need the log output with milliseconds in the datetime.

It would be nice when the log format would be configurable.

My suggestions is, to have this in the config:
```
logging:
    level: ERROR
    #logfile: /tmp/pygeoapi.log
    #logformat: [%(asctime)s] {%(pathname)s:%(lineno)d} %(levelname)s - %(message)s
    #dateformat: %Y-%m-%dT%H:%M:%SZ
```

If this is ok for you, you can assign the issue to me and i will implement it.

Regards,
Andreas

### LINKED ISSUE #1389: Logfile rotation
Hello,

it would be nice to have a logfile rotation, because the logs could grow fast.

In the Python logger it is possible to rotate logs size or time based.

http://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler
http://docs.python.org/3/library/logging.handlers.html#timedrotatingfilehandler

My suggestion would be, to make it configurable in the config file.
```
logging:
    level: ERROR
    #logfile: /tmp/pygeoapi.log
    #rotate_mode: off # [off|time|size]
    #rotate_when: midnight # [s|m|h|d|w0-w6|midnight]
    #rotate_interval: 0
    #rotate_max_bytes: 20
    #rotate_backup_count
```

If this is ok for you, you can assign this issue to me and i will implement it.

Regards,
Andreas

xkosubek:
# Overview
- Added configurable log and date format
- Added support for logfile rotation
- Updated documentation

# Related Issue / Discussion

fixes #1389
fixes #1388 

# Contributions and Licensing

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -99,13 +99,13 @@ The ``logfile rotation`` supports rotation of disk log files. The ``logfile`` fi
   logging:
       logfile: /path/to/pygeoapi.log  # the full file path to the logfile
       rotation:
-          mode: # [None|time|size]
+          mode: # [time|size]
           when: # [s|m|h|d|w0-w6|midnight]
           interval: 
           max_bytes: 
           backup_count: 
 .. note::
-  The ``mode`` can be defined by size or time.
+  Rotation block is not mandatory and defined only when needed. The ``mode`` can be defined by size or time.
   For RotatingFileHandler_ set mode size and parameters max_bytes and backup_count.
 
   .. _RotatingFileHandler: http://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler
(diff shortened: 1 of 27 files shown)


docs/source/configuration.rst
```diff
@@ -82,11 +82,39 @@ The ``logging`` section provides directives for logging messages which are usefu
...
+^^^^^^^^^^^
+
+The ``logfile rotation`` supports rotation of disk log files. The ``logfile`` file is opened and used as the stream for logging.
+
+.. code-block:: yaml
+
+  logging:
+      logfile: /path/to/pygeoapi.log  # the full file path to the logfile
+      rotation:
+          mode: # [None|time|size]
```
totycro, line 102: Currently if the `mode` is set to `null` in yaml, which would correspond to python `None,` this would lead to an exception because it would be detected as an unknown mode.

I think the code is fine though, and we could require here in the documentation that mode must only be `time|size`. If someone doesn't want log rotation, they can just not have the whole `rotation` block.

totycro (approved):

Issue #1418, opened 2023-12-07
Error displaying Tilejson metadata


doublebyte1:
**Description**

The [tileset metadata endpoint](https://demo.pygeoapi.io/master/collections/lakes/tiles/WorldCRS84Quad/metadata) throws a 500 error. This is a regression behaviour, as it was working at [this](https://github.com/geopython/pygeoapi/commit/ec564847170b6bc4ce1018d6272c05a6053b7eab) point in time.

**Steps to Reproduce**

Access the tileset metadata for an OGC API collection, that uses [tilejson as metadata format](https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_master/local.config.yml#L158):
https://demo.pygeoapi.io/master/collections/lakes/tiles/WorldCRS84Quad/metadata

![500](https://github.com/geopython/pygeoapi/assets/1038897/b7c33e5e-cb3b-4cf4-9ab6-c32cdc72a07a)

**Expected behavior**

The request should return the tileset metadata in tilejson format.

![200](https://github.com/geopython/pygeoapi/assets/1038897/8d7bad2b-2344-46aa-875a-f799662fb5bc)

**Environment**
 - OS: macOS Sonoma
 - Python version:  3.8.6
 - pygeoapi version: 0.16.dev0

**Additional context**

The error is triggered by the pydantic validation of the MVTTilesJSON model:

`pydantic_core._pydantic_core.ValidationError: 3 validation errors for MVTTilesJson
tiles
  Field required [type=missing, input_value={'name': '/data/tiles/', ...,"51","52","53"]}]}]}}'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.4/v/missing
attribution
  Field required [type=missing, input_value={'name': '/data/tiles/', ...,"51","52","53"]}]}]}}'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.4/v/missing
vector_layers
  Field required [type=missing, input_value={'name': '/data/tiles/', ...,"51","52","53"]}]}]}}'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.4/v/missing`

Pydantic is failing the validation, because of missing (optional) fields.

PR #1353, opened 2023-09-20
Refactor code base to make it work with pydantic v2

MTachon:
# Overview
This PR refactor the code using pydantic to make it work with pydantic V2. Originally implemented in PR https://github.com/geopython/pygeoapi/pull/1349, but moved the pydantic related changes to this new PR.

# Related Issue / Discussion
https://github.com/geopython/pygeoapi/pull/1349

# Additional Information
This PR add `typing-extensions` to `requirements.txt` to allow for the use of the `Literal` type. If the python version is update in CI to version 3.8 or above, the `Literal` can be directly imported from the `typing` standard library and `typing-extensions` will not be needed.

# Contributions and Licensing

(as per https://github.com/geopython/pygeoapi/blob/master/CONTRIBUTING.md#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

PR #1421, opened 2023-12-07
Fixing pydantic validation issue of tilejson metadata

doublebyte1:
# Overview

The tileset metadata endpoint for tilejson metadata is no longer working.

# Related Issue / Discussion

https://github.com/geopython/pygeoapi/issues/1418

# Additional Information

This issue is related to the upgrade of the pydantic version: https://github.com/geopython/pygeoapi/pull/1353

According to [this](https://docs.pydantic.dev/latest/migration/), pydantic2 requires all the optional fields to be present, unless they have a default value of None. This PR adds defaults values to all the optional fields, so that we[ continue to instantiate the model, without having all the information](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/mvt.py#L283).

# Contributions and Licensing

### DIFF
--- a/pygeoapi/models/provider/base.py
+++ b/pygeoapi/models/provider/base.py
@@ -3,8 +3,10 @@
 # =================================================================
 #
 # Authors: Antonio Cerciello <anto.nio.cerciello@gmail.com>
+#          Francesco Bartoli <xbartolone@gmail.com>
 #
 # Copyright (c) 2022 Antonio Cerciello
+# Copyright (c) 2023 Francesco Bartoli
 #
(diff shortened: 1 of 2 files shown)


ricardogsilva:
+1

This is a consequence of migrating to pydantiv V2, where the `Optional` type annotation no longer makes pydantic provide an automatic intialization value of `None` - more info here (see the _note_ box):

https://docs.pydantic.dev/latest/concepts/models/#required-fields

PascalLike:
The same potential issue also here https://github.com/geopython/pygeoapi/blob/master/pygeoapi/models/provider/base.py

Issue #1420, opened 2023-12-07
Parameter 'outputs' from request object is not propagated to the manager execute_process and then to processor execute


francescoingv:
In **API.execute_process()**,
the (optional) parameter '**outputs**', if present in the request, is not passed on to **BaseManager.execute_process()** (or derived).
Even more, the parameter '**outputs**' is not collected: only the parameter '**inputs**' is collected.

To properly process a request the processor (i.e. "**BaseProcessor**" or derived) should get the requested outputs to know which outputs to produce.
I expect the manager (e.g. "**BaseManager**" or derived) to use this information, and also the property "**transmitionMode**" of each output, to prepare:
- body;
- headers (i.e. 'Location');
- content pointed by 'Location' in the headers.

In case of **"transmissionMode": "reference"**, I am not sure if either the **processor** or the **manager** should prepare the reference (possibly writing to a file the single output).
As far as I understand, the reference is not expected to be served by the API, therefore only the processor can prepare it (possibly on a different endpoint).
If the reference could be server by the API (but I do not understand the corresponding entrypoint), then the manager could be in charge of it.

**I propose** to add an **additonal parameter** (a dictionary with the content of outputs) **to BaseManager.execute_process()**
and **the same additional parameter to BaseProcessor.execute()**

francbartoli:
Hi @FrancescoIngv, few comments on the formal aspects of the ticket:

- It's still not clear to me if we should consider the ticket as a `feature request` or a `bug report` even if you have flagged it as enhancement.
- Can you kindly assess with the OGC API - Processes specification if the current pygeoapi interface of the OpenAPI document doesn't comply with the standard? (i.e. `outputs` is not taken into account) If it doesn't then it's clearly a bug otherwise it would be a feature to be added
- Regardless of the outcome I would strongly encourage you to edit your explanation above and adapt its content under one of the relevant [template](https://github.com/geopython/pygeoapi/issues/new/choose) that the team has prepared for the reporters. This would help us to understand the problem and review properly any eventual pull request that would fix the issue.

Thanks,
Francesco

FrancescoIngv:
**Description**
The processor cannot return outputs based on the `outputs` parameter of a `process-execute`

**Steps to Reproduce**
Create a processor (say `TestProcessor`) derived from `BaseProcessor` with the following `outputs` section in `PROCESS_METADATA`
```
  "outputs": {
    "echo1": {
      "schema": {
        "type": "string"
      },
    "echo2": {
[...]

Issue #1285, opened 2023-06-23
Process execution request parsing not aligned with OGC API - Processes


ricardogsilva:
**Description**
When asking for the execution of a process, pygeoapi parses the incoming request and only keeps the `inputs` property of the request:

https://github.com/geopython/pygeoapi/blob/f436f0553f36213fb7be43f19a37d30c143c1c8a/pygeoapi/api.py#L3558

[According to OGC API - Processes, an execution request has other relevant properties](https://docs.ogc.org/is/18-062r2/18-062r2.html#toc32), such as:

- `outputs` - allows the client to request different transmission modes (by value or reference) and also to select between available media-types for each output - as such, only the default transmission mode (by value) and for a single media type per output are supported

- `response` - allows the client to either ask for a raw response or a document response. As such, only the default value, wich is raw response, is supported

- `subscriber` - allows the client to ask to be notified when processing is over. 

In order to be inline with the spec, pygeoapi ought to pass all properties of the execution request down to the process manager and onto processes, and implement these alternative process execution features

totycro:
We actually need the `subscriber` property in our company, so I'm going to look into this.

StefanBrand:
@tomkralidis @ricardogsilva Please, comment on the status of this issue and the related merge request(s). We would like to send Slack messages for finished jobs.

ricardogsilva:
@StefanBrand 

To the best of my knowledge there hasn't been much progress on this.

I've reviewed the proposed PR back in July and I'm +1 for it to be merged (with some smaller tweaks, which have mostly already been addressed). 

Personally I'm more interested in having support for `outputs`, as that would unlock a lot of customization in how processes are run - I'm also waiting on the pending PR to be merged so that I can work on implementation of `outputs` after it.

There could however be some deeper plan for this, as I am not fully in the loop of all the latest OGC-related developments. So I guess the status of this is that we are waiting on @tomkralidis input.

StefanBrand:
Related PRs #1313 and #1448 were updated 5 days and 34 days ago, respectively.

StefanBrand:
Current status if I'm not mistaken:
- `subscriber`: #1313 was merged on March 11th
- `outputs`: #1602 was merged on May 10th
- `response`: currently not implemented

tomkralidis:
> 
>     * `response`: currently not implemented

[...]

PR #1435, opened 2023-12-18
ogc api processes - outputs #1420

FrancescoIngv:
# Overview
Add support for OGC API Processes Outputs

The parameter `outputs` is passed to the manager, which in turn pass it to the Processor.

By default the processor returns all the outputs defined into the section `outputs` of `PROCESS_METADATA`.
If the processor receives the parameter `outputs` (a dictionary), then only the outputs with the keys present into the parameter should be returned.

# Related Issue / Discussion
Solve #1420,
Subset of #1285
Similar issue of the same subset #1313

# Additional Information
The change would be a breaking change,
but it would be fairly simple for maintainers to accept the parameters and ignore them for now if not interested,
as shown in the modified `HelloWorldProcessor`.

# Contributions and Licensing


FrancescoIngv:
Made an error and mixed two PR.
Going to close and will open two different PR.

Issue #1413, opened 2023-11-30
Elasticsearch provider with basic security


emiliopardo:
**Is your feature request related to a problem? Please describe.**
I have a elasticsearch cluster with basic security, when I use elasticsearch provider I need to pass user and password. When I configuere without user and password 

```
pygeoapi    |   File "/usr/local/lib/python3.10/dist-packages/requests/models.py", line 1021, in raise_for_status
pygeoapi    |     raise HTTPError(http_error_msg, response=self)
pygeoapi    | requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url:
```

**Describe the solution you'd like**
configure in the elasticsearch provider username an password  for https elastic conections

tomkralidis:
The ES provider `data` value can be set as an RFC1738 URL, i.e.:

```yaml
        providers:
            - type: feature
              name: Elasticsearch
              data: http://USERNAME:PASSWORD@localhost:9200/canada-surface-weather-obs
              id_field: some_id
```

To avoid encoding the credentials directly, you also use environment variables, i.e.:

```yaml
        providers:
            - type: feature
              name: Elasticsearch
              data: http://${USERNAME}:${PASSWORD}@localhost:9200/canada-surface-weather-obs
              id_field: some_id
```

Testing this works locally.  If this works for you, I'll update the [ES provider documentation](https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-features.html#elasticsearch).

emiliopardo:
Thanks Tom

If I load the example data It's work. 

"https://raw.githubusercontent.com/geopython/pygeoapi/master/tests/data/ne_110m_populated_places_simple.geojson"

when data is imported with "https://raw.githubusercontent.com/geopython/pygeoapi/master/tests/load_es_data.py" the mapping create two levels of ['properties']

```json
{
    "properties": {
        "geometry": {
            "type": "geo_shape"
        },
        "properties": {
            "properties": {
                "nameascii": {
                    "type": "text",
                    "fields": {
                        "raw": {
                            "type": "keyword"
                        }
                    }
                }
            }
        }
    }
}
```

the method def [get_fields()](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/elasticsearch_.py#L94C5-L94C19) of elastic provider inspect de fields of the index

If the mappings is not acording with the example with two levels of properties not work

when you create an index with ogr2ogr  the default mapping os properties is 
```json
{
  "prueba": {
[...]

PR #1441, opened 2023-12-19
add notes on using RFC1738 URLs for ES connections (#1413)

tomkralidis:
# Overview
Adds a note about using RFC1738 URLs for ES instances that have access control applied.
# Related Issue / Discussion
#1413 
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/docs/source/data-publishing/ogcapi-features.rst
+++ b/docs/source/data-publishing/ogcapi-features.rst
@@ -109,7 +109,17 @@ To publish an Elasticsearch index, the following are required in your index:
          id_field: geonameid
          time_field: datetimefield
 
-This provider has the support for the CQL queries as indicated in the table above.
+.. note::
+
+   For Elasticseach indexes that are password protect, a RFC1738 URL can be used as follows:
+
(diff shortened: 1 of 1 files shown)


docs/source/data-publishing/ogcapi-features.rst
```diff
@@ -109,7 +109,17 @@ To publish an Elasticsearch index, the following are required in your index:
          id_field: geonameid
          time_field: datetimefield
 
-This provider has the support for the CQL queries as indicated in the table above.
+.. note::
```
doublebyte1, line 112: @tomkralidis good catch! Should this be added to the records provider, as well? https://github.com/geopython/pygeoapi/blob/master/docs/source/data-publishing/ogcapi-records.rst?plain=1#L50

tomkralidis: Good idea!  PR updated.  I also noticed that the ES provider in records also does transactions, so this has been updated as well in the docs/this PR.

tomkralidis:
FYI CI failures are unrelated.

doublebyte1 (approved):

Issue #1417, opened 2023-12-06
Dynamic configuration update (for dynamic service registration)


geotom:
# Dynamic configuration (without server restart)

I would love to work with pyGeoAPI as a dynamic service registry. This means a hot reloading of the service configuration file to allow registration and unregistration of services. This could also benefit other services of pyGeoAPI and allow for dynamic setups where pyGeoAPI works as outbound interface with a custom internal orchestrator tool which supports to register / unregister services.

I do not think we need an internal API for this. It would be enough to be able to alter the configuration by custom code and call a configuration reload on pyGeoAPI. Right now, I have to manipulate the configuration and restart the full pyGeoAPI.

Thanks for considering this (if it is not already possible)

geotom:
That sounds really great @tomkralidis! UI part for me is less important, but the ability to update parts of the configuration dynamically and the server dynamically responding to the updates would be a killer feature. Thanks for hinting the issue tickets

tomkralidis:
That is an option however it is gunicorn specific (note: you can don this now by updating the gunicorn startup). Perhaps a watcher in pygeoapi proper would be the most generic approach.

tomkralidis:
Yes, see #1443.

PR #1137, opened 2023-02-10
Add Admin API

webb-ben:
# Overview
This PR introduces an OGC API - Features - Part 4 inspired admin API around configuration of pygeoapi resource configurations. It has been implemented and tested for a couple of months as a part of the WMO wis2box OSS software developments efforts in [wis2box-api](https://github.com/wmo-im/wis2box-api). 

@ksonda @tomkralidis Interested to hear what you think about the next steps

Tasks that I am aware still need to be completed:
- [x] Better unit & integration testing for admin API. Might make sense to move the admin api tests to their own CI job.
- [x] conservation of environment variables. 
- [x] deploy the admin CI with a mechanism better than the building the docker example.
- [x] Implementation of corresponding routes with Starlette and Django
- [x] Remove vue from HTML templates
- [x] Create Admin API documentation

# Related Issue / Discussion
https://github.com/geopython/pygeoapi/pull/756, https://github.com/geopython/pygeoapi/issues/728

# Additional Information
Demo deployment exists in the docker examples folder. To Hotload the changes from the API to the configuration, requires changes like that made in the examples entrypoint script.

# Contributions and Licensing

(as per https://github.com/geopython/pygeoapi/blob/master/CONTRIBUTING.md#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

PR #1443, opened 2023-12-19
add gunicorn reload trigger to Docker setup

tomkralidis:
# Overview
This PR adds gunicorn's reload functionality based on changes to a given file.  As a result, when pygeoapi configuration changes, gunicorn will restart its workers accordingly.
# Related Issue / Discussion
Partial solution for #1417 (we will introduce 
# Additional Information

# Contributions and Licensing

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -100,6 +100,8 @@ case ${entry_cmd} in
 				--timeout ${WSGI_WORKER_TIMEOUT} \
 				--name=${CONTAINER_NAME} \
 				--bind ${CONTAINER_HOST}:${CONTAINER_PORT} \
+                                --reload \
+				--reload-extra-file ${PYGEOAPI_CONFIG} \
 				pygeoapi.flask_app:APP
 	  ;;
 	*)


webb-ben (approved):
This was left as user driven for the docker entrypoint in https://github.com/geopython/pygeoapi/pull/1137 because in the [gunicorn reload docs](https://docs.gunicorn.org/en/stable/settings.html#reload), it says this is really intended for development. 

In lieu of implementing a hot-reload into pygeoapi itself, its too easy of a solution to not include. +1

ricardogsilva:
@tomkralidis 

How about just sending a signal to the gunicorn container instead? As per the gunicorn docs at:

https://docs.gunicorn.org/en/latest/signals.html#master-process

The `HUP` signal can be sent in order to ask gunicorn to reload itself. The pygeoapi docker image `exec`s gunicorn in the entrypoint script, which mean gunicorn is always `PID 1` inside the container.  This means that we are able to reload gunicorn (and subsequently pygeoapi) like this:

```sh
docker exec -ti <running-pygeoapi-container> kill -HUP 1
```

I've used this incantation in the past and found it to work nicely. Granted, it is a super privileged operation, which requires access to the environment where the docker is running, so it is not appropriate for some use cases. But it avoids using the `--reload` flag for gunicorn, [which is documented as being intended only for development purposes](https://docs.gunicorn.org/en/latest/settings.html#reload). I also like that it is not automatic, the admin must explicitly choose to reload.

I'm not sure what is the use case that you are wanting to support with this though, so it is really just a suggestion.

Perhaps in the future we can also consider a scenario where the configuration of pygeoapi resources would be split from the pygeoapi config file and kept elsewhere, like some DB - this could enable adding/removing collections at runtime without restarting the server - then even building a secure web API for to allow CRUD on it?

kalxas (approved):

tomkralidis:
Thanks for the comments.  The use case is automated reload for runtime workflow.  Thinking more, perhaps an application level hot reload is best (which would be an optional flag/parameter for those who desire this functionality).

Issue #1440, opened 2023-12-19
BaseManager.execute_process fails if connection is not defined


francescoingv:
**Description**
Using the `BaseManager` as manager,
while it is safe to not define the configuration parameter `output_dir`,
it causes an error if it is not defined the configuration parameter `connection`.

I think it should be possible to safely use BaseManager also without the `connection` configuration parameter.
All other methods that could benefit from this parameter (`get_jobs, add_job, update_job, get_job, get_job_result, delete_job`)
currently return the proper Error, as they will possibly be implement by the derived classes as required.

**Steps to Reproduce**
Modify the `manager` section of `example-config.yml` with the following lines:
```
    manager:
        name:  'pygeoapi.process.manager.base.BaseManager'

```

Call a EchoProcessor
`curl localhost:5000/processes/hello-world/execution -H 'Content-Type: application/json' -d '{"inputs": {"message": "my message,", "name": "My Name"}}'
`
**Expected behavior**
The call produce the following result:
```
{
    "id":"echo",
    "value":"Hello My Name! my message,"
}

```

**Environment**
 - OS: Ubuntu
 - Python version: 3.8.10
 - pygeoapi version: master

**Additional context**
N/A

PR #1449, opened 2023-12-21
BaseProcessor without configuration parameter 'connection'

francescoingv:
# Overview
BaseProcessor works also without configuration parameter `connection`.

Methods `add_job()` and `update_job()` are now called subject to the condition the configuration parameter `connection` is provided.

Additional minor changes:
-) in `BaseProcessor`: improved the error description returned in case of errors during the processing.
-) in `API`: improved returned message if an error is found during the parsing of the json input string
-) in `echoProcessor`: amended returned message in case of wrong input parameter
``
# Related Issue / Discussion
Solve #1440 

# Contributions and Licensing

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,36 @@
+name: Build documentation ⚙️
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'docs/**'
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'docs/**'
+  release:
+    types:
+      - released
+
+jobs:
+  main:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        include:
+          - python-version: '3.10'
+    steps:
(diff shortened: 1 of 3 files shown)


FrancescoIngv:
The test **failed** because of the **modification of Exception** handling in [pygeoapi/process/manager/base.py](https://github.com/geopython/pygeoapi/pull/1449/commits/481153968b4e010a4fc910f2839380cb7500c1c5#diff-433b2272baac7285bacfdfb472b9575d9c726cac4b6080673cb1dbc15061b8bb)
from (line 293):
```
            outputs = {
                'code': code,
                'description': 'Error updating job'
            }
```
to (line 296):
```
            outputs = {
                'code': code,
                'description': str(err)
            }
```

somehow implementing the note to the Exception: `TODO assess correct exception type and description to help users`.

I can change back the above line and resubmit the PR, but I would rather suggest to modify the test.
Unfortunately I have no experience with testunits, and I cannot make any suggestion/proposal.

PR #1452, opened 2023-12-21
Make sure openapi.yml gets generated in Kubernetes environments

ymoisan:
# Overview
In some K8S environments the execution of the entrypoint that generates the openapi.yml file hits a problem : 

START /[entrypoint.sh](http://entrypoint.sh/)
Generating /pygeoapi/local.openapi.yml
Error: Could not open file ‘/pygeoapi/local.openapi.yml’: Permission denied

`touch`-ing and `chmod`-ing ensures this does not happen.


ymoisan:
Bad commits

Issue #1460, opened 2023-12-27
Documentation inadequate for starlette deployment


webb-ben:
**Description**
Following the documentation to deploy pygeoapi with Gunicorn and Starlette as per (https://docs.pygeoapi.io/en/latest/running.html#gunicorn-and-starlette), yields errors. Furthermore there is no way to run Starlette out of the current pygeoapi docker image. 

**Steps to Reproduce**
Run
```bash
gunicorn pygeoapi.starlette_app:app -w 4 -k uvicorn.workers.UvicornWorker
```

**Expected behavior**
I was able to run starlette by installing the debian package `uvicorn` and pip installing `requirements-starlette.txt` in the Dockerfile, and modifying the entrypoint exec to:
```
exec uvicorn \
    --workers=${WSGI_WORKERS} \
    --host=${CONTAINER_HOST} \
    --port=${CONTAINER_PORT} \
    --loop=asyncio \
    pygeoapi.starlette_app:APP
```

**Screenshots/Tracebacks**
Verbatim docs:
```
START /entrypoint.sh
Trying to generate openapi.yml
Generating /pygeoapi/local.openapi.yml
Done
openapi.yml generated continue to pygeoapi
make SCRIPT_NAME empty from /
Start uvicorn 0.0.0.0:80 with 4 workers and SCRIPT_NAME=
[2023-12-27 16:28:21 +0000] [1] [INFO] Starting gunicorn 20.1.0
[2023-12-27 16:28:21 +0000] [1] [INFO] Listening at: http://127.0.0.1:8000 (1)
[2023-12-27 16:28:21 +0000] [1] [INFO] Using worker: uvicorn.workers.UvicornWorker
[2023-12-27 16:28:21 +0000] [26] [INFO] Booting worker with pid: 26
[2023-12-27 16:28:21 +0000] [37] [INFO] Booting worker with pid: 37
[2023-12-27 16:28:21 +0000] [48] [INFO] Booting worker with pid: 48
[2023-12-27 16:28:21 +0000] [49] [INFO] Booting worker with pid: 49
Failed to find attribute 'app' in 'pygeoapi.starlette_app'.
[2023-12-27 16:28:22 +0000] [26] [INFO] Worker exiting (pid: 26)
Failed to find attribute 'app' in 'pygeoapi.starlette_app'.
[2023-12-27 16:28:22 +0000] [37] [INFO] Worker exiting (pid: 37)
[...]

Issue #1400, opened 2023-11-14
Replace flask and django dev servers with gunicorn


ricardogsilva:
[gunicorn](https://gunicorn.org/) is a pretty robust Python server that can be used to run both wsgi and asgi (with the uvicorn worker) web apps. In addition to being production-ready, it also contains features that make it a nice tool for using in development mode, such as:

- Nice documentation
- Automatic reload of the code and any extra files
- Controlling the number of worker processes that are used

Using gunicorn as both the dev and production web server, and regardless of the web framework being flask, starlette or django would tighten up the pygeoapi codebase and provide a more uniform running interface to users. We'd be able to recommend something like this:

```sh
export PYGEOAPI_CONFIG=my-config.yml
export PYGEOAPI_OPENAPI=my-pygeoapi-openapi.yml

# for production
pygeoapi serve --flask

# for development
pygeoapi serve --flask --debug
```

Both commands would run pygeoapi with `gunicorn`, and `pygeoapi serve --debug` would take care of turning on the needed gunicorn configuration stuffs for when dev'ing.

**Is your feature request related to a problem? Please describe.**
The current pygeoapi codebase is a bit fragmented when it comes to running pygeoapi for development. The fact that we support three different web frameworks and a different web server for each causes duplication in the code and leads to subtle bugs.

Additionally, the initialization of each of the three web frameworks contains a lot of different places where pygeoapi tries to read values from the environment and also a lot of global variables that deal with keeping state of the pygeoapi configuration.

I've been working on an implementation for this and will be submitting a PR soon.

francbartoli:
I'm sorry @ricardogsilva, but I have to give my -1 for this addition since I can see value only in certain cases of downstream projects. For example, if I want to run pygeoapi in serverless environments (lambda, az functions, etc) I see the fragmentation beneficial with its high grade of flexibility based on the selected framework.

ricardogsilva:
@francbartoli 

Perhaps I am missing some use case that you care about? For the concrete serverless example, I don't understand how this would be a hindrance - could you please elaborate on the difficulties that you foresee? I'm thinking that any shortcomings can likely be worked out if we discuss it :wink:

I've also submitted #1401 (it is still in draft, but its close to being ready - I'm just sorting out some docker-related stuff), which implements the ideas mentioned in this issue - perhaps you can take a look and test drive it to see whether it is indeed something negative for your use cases?

[...]

Issue #227, opened 2019-09-06
Improve the Docker strategy for deployments


francbartoli:
**Is your feature request related to a problem? Please describe.**
Support the deployment with Starlette through Docker

**Describe the solution you'd like**
`docker build -t geopython/pygeoapi-starlette -f starlette.dockerfile .`

**Describe alternatives you've considered**
Having Starlette/uvicorn in the same default image is dirty and it cannot be considered for production

**Additional context**

justb4:
Thinking out loud: will require some thinking how to design/build/provide `pygeoapi` Docker images. Many of the well-known Docker Images have variants in their naming and tagging. From a user-perspective we need a tagging strategy first. Similar to e.g. Python OS variants (Alpine, Debian etc) we need to standardize tagging variants like:

* `geopython/pygeoapi:0.7.0-flask`, `geopython/pygeoapi:0.7.0-starlette` or
* `geopython/pygeoapi-flask:0.7.0`, `geopython/pygeoapi-starlette:0.7.0` or ?

(would prefer first).
Also multiple tags can be attached to the same image. As many users would not care, they must be able to use defaults like `geopython/pygeoapi` which implies `geopython/pygeoapi:latest` which e.g. implies `geopython/pygeoapi:latest-flask`.

Next is then a build strategy as to not duplicate Dockerfiles. 

(need to go, to be continued).

francbartoli:
@justb4 I agree with the need for a strategy. Probably the first is more helpful but for the build, I don't think we can escape from having two different build path for Dockerfiles. For Starlette as an example, I would extend [those images](https://github.com/tiangolo/uvicorn-gunicorn-starlette-docker) rather than build our own from scratch.

frafra:
I think we can have a single Dockerfile with a build argument, which is then passed to `pdm` to install one or more groups of additional packages: https://github.com/geopython/pygeoapi/issues/344#issuecomment-1377218503.
We could have 2~3 default images for covering most of the needs, and the users could build upon them.

I would suggest using `geopython/pygeoapi:$version-$flavour`, which seems rather common.

PR #1461, opened 2023-12-27
Update starlette documentation

webb-ben:
# Overview
This PR adds the necessary components to deploy the pygeoapi Docker image with the startlette framework. This can be done by updating two necessary environment variables. Not sure WSGI_APP_FRAMEWORK is the best env variable name seeing as Starlette is ASGI

# Related Issue / Discussion
https://github.com/geopython/pygeoapi/issues/1460, https://github.com/geopython/pygeoapi/issues/1400, https://github.com/geopython/pygeoapi/issues/1111, https://github.com/geopython/pygeoapi/issues/227

# Additional Information

# Contributions and Licensing


francbartoli (commented):
Thanks @webb-ben, to be honest with you, if the intent is to provide a ready to go image for production purposes then I really don't understand why we are collapsing two different frameworks into the same image.
Also, as you mentioned starlette is ASGI and so I still really don’t understand why we should scale the workers number within the container rather than doing that at infrastructure level.
I can only understand this addition only if the goal here is to provide a uniform way to make the developer experience easier but I understood in the [#1252](https://github.com/geopython/pygeoapi/pull/1252) this is not the only intent.
Certainly it wouldn’t be the recommended way to deploy other ASGI framework like FastAPI (which is based on Starlette) on containers, k8s or even in serverless functions. In these case I should need a proper Dockerfile to build the optimised image from scratch.
Sorry, I might be wrong, but my feeling is more or less the same as I’ve got with [#1400](https://github.com/geopython/pygeoapi/issues/1400).
I'd like to help here but I don't have any idea other than having different Dockerfiles based on purposes/frameworks. Maybe we could start thinking about a separate `pygeoapi-devops` repository under the Geopython organisation.

webb-ben:
@francbartoli I do not have much experience deploying ASGI. This PR comes from my testing of the Admin API and following the pygeoapi docs on deploying pygeoapi with [gunicorn and starlette](https://docs.pygeoapi.io/en/latest/running.html#gunicorn-and-starlette) which are out of date. 

More than happy to reduce this PR to just fix the documentation if that is the intended purpose. This PR exists in lieu of proper Django and Starlette images for pygeoapi.

francbartoli:
@webb-ben at the same time I don't want to block it, thanks

tomkralidis:
+1 to reduce PR to documentation.  Thanks.

webb-ben:
@tomkralidis ready to merge. Although I am not sure we publish to ghcr anymore.

Issue #1458, opened 2023-12-27
Refactor Tiles Metadata UI to support multiple formats


doublebyte1:
**Is your feature request related to a problem? Please describe.**
The tiles Metadata UI only supports one format at a time (the one that was selected on the configuration file).

On this example, TileJSON was selected. 

![Screenshot from 2023-12-27 12-13-49](https://github.com/geopython/pygeoapi/assets/1038897/5e1060e9-b60b-445a-832c-1d33824e4e07)

![Screenshot from 2023-12-27 12-14-01](https://github.com/geopython/pygeoapi/assets/1038897/9137f6a3-7e8e-4e39-9c11-77dd3df89c79)

The "default" metadata format (OGC TMS) is not consistently supported by the UI. The option to display metadata does not appear in the tiles resource page: http://localhost:5000/collections/lakes/tiles

![Screenshot from 2023-12-27 12-33-19](https://github.com/geopython/pygeoapi/assets/1038897/9f24f6ae-75e2-4d10-8045-d831af575ae3)

And by accessing the url directly (http://localhost:5000/collections/lakes/tiles/WorldCRS84Quad/metadata), we see a UI page that is not consistent with the OGC 2D TMS format (see "tileJSON" reference):

![Screenshot from 2023-12-27 12-34-33](https://github.com/geopython/pygeoapi/assets/1038897/cdb31b84-1ff6-4e73-b9b7-951867aafadb)

**Describe the solution you'd like**
With [this](https://github.com/geopython/pygeoapi/issues/1457) refactoring, we would be able to support multiple formats:

- https://demo.pygeoapi.io/master/collections/lakes/tiles/WorldCRS84Quad/metadata?f=json -> OGC 2D TMS (default for machines)
- https://demo.pygeoapi.io/master/collections/lakes/tiles/WorldCRS84Quad/metadata?f=tilejson -> TileJSON
- https://demo.pygeoapi.io/master/collections/lakes/tiles/WorldCRS84Quad/metadata?f=html -> HTML (default for browsers)

The HTML page should be agnostic of the format. Similar to this implementation:
https://maps.gnosis.earth/ogcapi/collections/blueMarble/map/tiles/WorldCRS84Quad?f=tilejson
https://maps.gnosis.earth/ogcapi/collections/blueMarble/map/tiles/WorldCRS84Quad?f=json
https://maps.gnosis.earth/ogcapi/collections/blueMarble/map/tiles/WorldCRS84Quad?f=html

[...]

Issue #1457, opened 2023-12-27
Abstract Tile Matrix Set Metadata format from MVT Provider


doublebyte1:
**Is your feature request related to a problem? Please describe.**

The TMS format is defined by the user in the [configuration options](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/api.py#L2789). However, t[he ability to provide certain formats of metadata is restricted to specific providers.](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/mvt.py#L281). For instance, pygeoapi only returns metadata in TileJSON, if it finds a metadata.json file, which is something specific to Tippecanoe. In other cases, it does not honor the choice of the user.

**Describe the solution you'd like**
pygeoapi should return metadata in all supported formats (currently: TileJSON and OGC TMS), regardless of the provider. The user should not need to choose one single format in the configuration options, because all would be available.

This [refactoring](https://github.com/geopython/pygeoapi/issues/1450) introduces the opportunity to implement the metadata retrieval within the provider sub classes, which know the details of how they provide (or not) metadata. Then, it is a matter of doing our best effort to generate a minimal subset of metadata, for each format.

![mvt-metadata-small](https://github.com/geopython/pygeoapi/assets/1038897/40d23559-5149-4a60-a813-5a2df1f72fb4)

**Describe alternatives you've considered**
Leaving it as it is.

**Additional context**
This refactoring would also require a refactoring of the UI for Tile Matrix Set Metadata.

PR #1454, opened 2023-12-22
MVT Provider refactoring

doublebyte1:
# Overview
This PR introduces a refactoring in the MVT provider class. The MVT Provider becomes a base class, which is implemented through different providers. The tippecanoe and elastic MVT providers are also offered in this PR.

Tests, documentation and docker examples are also updated accordingly.

# Related Issue / Discussion
https://github.com/geopython/pygeoapi/issues/1450

# Contributions and Licensing

(as per https://github.com/geopython/pygeoapi/blob/master/CONTRIBUTING.md#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

PR #1482, opened 2024-01-08
Tiles Metadata provider refactor

PascalLike:
# Overview
Following the [tiles providers refactor](https://github.com/geopython/pygeoapi/pull/1454) started by @doublebyte1 

- Adding to each provider available (mvt_tippecanoe and mvt_elastic for now) methods to return HTML, [OGC Tiles Metadata](https://docs.ogc.org/is/17-083r4/17-083r4.html#toc20) (it should always be available) e TileJSON Metadata (default not implemented, available in Tippecanoe only)
- The metadata format to return data can be indicated in the query string, making it possible to return metadata in more formats (another thing that depends on the provider)
- Removing CUSTOM metadata (it does not make sense anymore)
- Changing metadata HTML template to show basic info

# Related issue / discussion
https://github.com/geopython/pygeoapi/issues/1457
https://github.com/geopython/pygeoapi/issues/1458

# Additional information
There is a PR drafted for demo.pygeoapi.io when this will be merged (https://github.com/geopython/demo.pygeoapi.io/pull/49). Btw this is not a breaking change. As far as I can test, this should not break anything.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


pygeoapi/templates/collections/tiles/index.html
```diff
@@ -39,25 +39,23 @@ <h3>Tiles</h3>
...
       <div class="row">
-        {% if data['format']=='tilejson' %}
           <div class="col-md-2 col-sm-12">{% trans %}Metadata{% endtrans %}</div>
-          <div class="col-md-8"><a id="tilejson" href="" target="_blank">Tiles metadata in {{ data['format'] }} format</a></div>
-        {% endif %}      </div>
+          <div class="col-md-8"><a id="metadata_link" href="" target="_blank">Metadata</a></div>
+      </div>
       <script>
         var select = document.getElementById('tilingScheme');
-        var tileset = select.value;
```
doublebyte1, line 48: I think this is problematic, because when the page loads, we arrive [here](https://github.com/geopython/pygeoapi/pull/1482/files#diff-a33f7def52ad3a6af980fe603069ad9add9fc080d131b966551b592241dbf432R87) with an empty "tileset" variable. It could work, if the map refreshed itself, each time we choose a tileset on the dropdown.

doublebyte1 (commented):
Great work @PascalLike !! 👏🏽 

When you have a moment, please remove also these lines in the docs:
https://github.com/geopython/pygeoapi/blob/e9bb5dcc95b9c440bfe9c713ce4e14b5cdb7b834/docs/source/data-publishing/ogcapi-tiles.rst?plain=1#L49

https://github.com/geopython/pygeoapi/blob/e9bb5dcc95b9c440bfe9c713ce4e14b5cdb7b834/docs/source/data-publishing/ogcapi-tiles.rst?plain=1#L80

It is also worth to remove this line:
https://github.com/geopython/demo.pygeoapi.io/blob/aa257cb51ce52dcc7b66f39be15d30b98b9ffc5e/services/pygeoapi_master/local.config.yml#L158

pygeoapi/templates/collections/tiles/index.html
PascalLike: Fixed here https://github.com/geopython/pygeoapi/pull/1482/commits/8a7b77d514873df44ab794ce221731f8a2866710

PascalLike:
> Great work @PascalLike !! 👏🏽
> 
> When you have a moment, please remove also these lines in the docs:
> 
> https://github.com/geopython/pygeoapi/blob/e9bb5dcc95b9c440bfe9c713ce4e14b5cdb7b834/docs/source/data-publishing/ogcapi-tiles.rst?plain=1#L49
> 
> https://github.com/geopython/pygeoapi/blob/e9bb5dcc95b9c440bfe9c713ce4e14b5cdb7b834/docs/source/data-publishing/ogcapi-tiles.rst?plain=1#L80
> 
> It is also worth to remove this line: https://github.com/geopython/demo.pygeoapi.io/blob/aa257cb51ce52dcc7b66f39be15d30b98b9ffc5e/services/pygeoapi_master/local.config.yml#L158

Thanks @doublebyte1 ! I fixed the things you suggested.

doublebyte1 (approved):

pygeoapi/provider/base_mvt.py
```diff
@@ -125,100 +119,102 @@ def get_tiles(self, layer=None, tileset=None,
...
+                            enum TilesMetadataFormat
+        :param title: title name
+        :param description: description name
+        :param keywords: keywords list
+
+        :returns: `dict` of JSON metadata
+        """
+        raise NotImplementedError()
+
+    def get_tilejson_metadata(self, dataset, server_url, layer, tileset,
```
francbartoli, line 161: Can we make this method name generic? Like `get_vendor_metadata`

pygeoapi/provider/mvt_tippecanoe.py
```diff
@@ -169,22 +174,89 @@ def get_tiles(self, layer=None, tileset=None,
             except FileNotFoundError as err:
                 raise ProviderTileNotFoundError(err)
 
-    def get_metadata(self, dataset, server_url, layer=None,
-                     tileset=None, metadata_format=None, title=None,
-                     description=None, keywords=None, **kwargs):
+    def get_html_metadata(self, dataset, server_url, layer, tileset,
```
francbartoli, line 177: Can we continue to render in `html` metadata fields for tippecanoe?

francbartoli (changes requested):

pygeoapi/provider/mvt_tippecanoe.py
PascalLike: Done! Thanks for the review, Francesco.

francbartoli (approved):

PR #1484, opened 2024-01-09
Added support for websocket processing microservices

herzogrh:
# Overview
This PR adds functionality to connect the pygeoapi with processing microservices via Websockets connections. 
In that way, it is possible to execute processes in other environments than in Python, but still make them accessible through pygeoapi's OGC API processes and job management. As an example, one could connect a [node.js processing microservice](https://github.com/citysciencelab/processing-microservice-nodejs) with pygeoapi and run Insight Maker simulations. Another benefit is that one can dynamically add and remove processing microservices without having to recompile pygeoapi. 
Once a microservice establishes a websocket connection, it registers a process that is then added to the list of processes in pygeoapi. Whenever an execution request is made by a client, the proposed feature will forward the inputs through the websocket connection and the microservice will start the calculation. Once the results are calculated, the microservice transmits the results back to be stored in the job management. 
**Important**: A database manager must be configured for the new feature to work.

# Background
In the urban digital twin project [Connected Urban Twins](https://www.connectedurbantwins.de/en/) we identified the use case of multi-dimensional simulations running on various different modelling paradigms and software stacks. Therefore it is necessary to easily spin up and deploy various processing microservices via standardised APIs such as the OGC API Processes. As pygeoapi is a great implementation of OGC API processes and provides both an interface and an integrated job management, we added the feature to connect Websocket processing microservices.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute a websocket endpoint for processing microservices to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -3108,9 +3108,14 @@ class API:
                         'InvalidParameterValue', msg)
 
             for key in relevant_processes:
-                p = self.manager.get_processor(key)
-                p2 = l10n.translate_struct(deepcopy(p.metadata),
-                                           request.locale)
+                p, p2 = None, None
+
+                # Check if process is a "normal" process or one connected with websockets
+                if "processor" in self.manager.processes[key]:
+                    p = self.manager.get_processor(key)
+                    p2 = l10n.translate_struct(deepcopy(p.metadata), request.locale)
+                elif "type" in self.manager.processes[key] and self.manager.processes[key]["type"] == 'process-socket':
+                    p2 = l10n.translate_struct(self.manager.processes[key]["metadata"], request.locale)
                 p2['id'] = key
 
                 if process is None:
@@ -3340,6 +3345,15 @@ class API:
             return self.get_exception(
                 HTTPStatus.NOT_FOUND, headers,
                 request.format, 'NoSuchProcess', msg)
+        
+        process_type = "process"
+        sid = None
+        
+        # Check if process is a "normal" process or one connected with websockets
+        if self.manager.processes[process_id]["type"] == 'process-socket':
+            process_type = "process-socket"
+            sid = self.manager.processes[process_id]["sid"]
+        
 
         data = request.data
         if not data:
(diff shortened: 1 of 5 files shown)


herzogrh:
@tomkralidis In case you find the time, could you let us know if such a feature would be generally interesting for pygeoapi?

tomkralidis:
@herzogrh thanks for reaching out, and for this PR.  Given the refactoring of `pygeoapi.api` in recent weeks, as well as RFC4 (i.e. `non-trivial contributions in the form of Pull Requests are best put forth in the context of an Issue, so that discussion can take place to identify any possible concerns or impacts`), I would suggest to close this PR and open an issue.  We can discuss / determine the best approach and feasibility on whether this should be in pygeoapi proper or an an example/add-on/downstream application.

geotom:
Hi @herzogrh, I find the idea behind your PR great and that is actually exactly what I want from pygeoapi. I am one of the authors of the older [Geoserver WPS Remote ](https://docs.geoserver.org/main/en/user/community/remote-wps/index.html) community extension that did something similar, but using XMPP instead of websockets at that time.

I would absolutely love to see such a feature in pygeoapi (as extension maybe). I would probably not use XMPP anymore and also not sure about Websockets. I would implement such a feature today with an agnostic abstraction API, so people could chose their own means of message transport.

But the central idea of using pygeoapi as a dynamic service interface with a messaging infrastructure behind is actually very great. It would allow
- to offload requests to remote workers
- scale processing horizontally
- a dynamic service registration/discovery

Did you already create a new issue for this @herzogrh? Let's get in contact if you're interested to maybe draft a proper concept together. Is there in general some work in this direction already @tomkralidis?

herzogrh:
@geotom @tomkralidis thanks for your comments! I created an issue (#1671) with the basic content of the PR. I have no experience with agnostic abstraction APIs, but sounds beneficial to not stick to a single technology. Happy to get in touch and look into a proper concept!

PR #1453, opened 2023-12-21
Translate between generic `geometry` queryable name and the actual geom field name in postgres db

ricardogsilva:
This PR adds a method to `pygeoapi.provider.postgresql.PostgreSQLProvider` that modifies the previously parsed filter expression, replacing occurrences of the generic `geometry` attribute with the actual name of the geometry column for the collection.

The proposed implementation operates on the existing pygeofilter ast node tree, traversing the tree and replacing the name of nodes of type `Attribute` - it only operates on nodes with `name = 'geometry'` and replaces this name with whatever is the value of the provider's `geom` property.

It also adds some targeted unit tests to verify the correctness of the transformation.

#### Note

The proposed implementation does not modify the previously existing logic that a queryable named `geometry` is presented as the default geometry field name. However, upon reading the relevant section of [OGC API Features - Part 3: Filtering (section 6, requirements 3.D and 3.E](https://docs.ogc.org/DRAFTS/19-079r1.html#queryables) it seems the current behavior is not compliant with the spec, which mandates a name like `geometry-any` - the spec is still a draft though. This PR does not try to modify the current implementation in this regard.

# Related Issue / Discussion
- fixes #1350 

# Contributions and Licensing

(as per https://github.com/geopython/pygeoapi/blob/master/CONTRIBUTING.md#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution.
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

PR #1489, opened 2024-01-11
Convert coordinates to storage crs when filtering via cql

### LINKED ISSUE #1488: Enable CQL filtering to accept filter geometries in any CRS
pygeoapi currently offers CQL filtering for some backends. However, when supplying a geometry as part of the CQL filter, one needs to ensure it uses the same CRS as the corresponding collection.

It would be great if pygeoapi would be able to accept geometries in whatever CRS and would then convert them to the appropriate CRS, as needed by the collection provider.

As an example, I have a feature collection which is stored in a PostGIS DB and has a storage CRS of `EPSG:3004`. I would like to
perform a request for retrieving those collection items which are located whithin 100 m of some reference point. This translates to the following request:

```sh
curl 'http://localhost:5000/collections/beni_confiscati/items?\
    filter=DWITHIN(geometry,SRID=3004;POINT(2313683.65%204641312.29),100,meters)'
```

As illustrated above, in order for my request to work, I currently need to pass in a point defined in EWKT with the storage CRS (_i.e._ `SRID=3004;POINT(2313683.65,4641312.29)` - this is not always convenient.

I would like to be able to perform a request with a filter where my `POINT` would simply be defined using the `OGC:CRS84` CRS implicitly and then pygeoapi would convert that to whatever CRS the feature collection is stored in - Something like this:

```sh
curl 'http://localhost:5000/collections/beni_confiscati/items?\
    filter=DWITHIN(geometry,POINT(12.512829%2041.896698),100,meters)'
``` 

I would also like to be able to keep specifying a custom CRS and being able to also have pygeoapi automatically convert input geometries (if needed) in order to be able to query the collection. For example, these would be nice to have:

```sh
# supplying the same CRS as the one being used internally - no conversion of coordinates needed
curl 'http://localhost:5000/collections/beni_confiscati/items?\
    filter=DWITHIN(geometry,SRID=3004;POINT(2313683.65%204641312.29),100,meters)'

# supplying another CRS
curl 'http://localhost:5000/collections/beni_confiscati/items?\
    filter=DWITHIN(geometry,SRID=3857;POINT(1392921%205145517),100,meters)' 
```

ricardogsilva:
# Overview

This PR implements conversion of geometries supplied in CQL filters to the storage CRS of a feature collection.

The main part of the implementation follows this logic:

1. Iterate through the nodes of the previously-parsed pygeofilter filter
2. If a node is of type `pygeofilter.values.Geometry`, retrieve its CRS information or default to the `OGC:CRS84` CRS if none is provided - CRS information is gotten from the `filter-crs` query parameter, if present - it can also be provided in the geometry definition, using EWKT notation
3. Depending on the CRS of the collection, as designated in the `storage_crs` configuration option of the relevant provider, determine whether it is necessary to transform the geometry to match the same CRS and the one used by the provider
4. If needed, convert the filter geometry's coordinates using a combination of pyproj and shapely, in similar fashion as the approach already being used elsewhere in the pygeoapi codebase
5. Update the pygeofilter node tree with the converted geometry, which will subsequently be used by the provider to query its datasource

This PR includes the following related modifications:

- `pygeoapi.util.get_crs_from_uri()` has been refactored in order to support both URN and URL types of URI. This was done in order to accomodate the fact that pygeofilter is using the URN form internally
- the logic to iterate through the pygeofilter node tree was previously only available inside the postgres provider. It has been moved out to the `pygeoapi.util.modify_pygeofilter()` function. This function now performs two checks:

    - Translates from the generic `geometry` name to whatever name the provider specifies in its `geom_field` configuration parameter - This is the same functionality introduced recently in #1453 
    - Converts coordinates to the appropriate storage CRS, _i.e._ the main purpose of this PR

This PR enables the use cases mentioned in #1488, which is basically to use whatever CRS when defining a geometry in a CQL filter, and default to assuming `OGC:CRS84`.

Example requests:

```sh

# no CRS specified - pygeoapi assumes OGC:CRS84 by default
curl http://localhost:5000/collections/beni_confiscati/items?\
    filter=DWITHIN(geometry,POINT(12.626000006069875 41.8108000011414),100,meters)

# CRS specified by means of the filter-crs query parameter
curl http://localhost:5000/collections/beni_confiscati/items?\
    filter=DWITHIN(geometry,POINT(2313683.65%204641312.29),100,meters)&\
    filter-crs=http://www.opengis.net/def/crs/EPSG/0/3004

# CRS specified by defining the filter geometry as EWKT
curl http://localhost:5000/collections/beni_confiscati/items?\
    filter=DWITHIN(geometry,SRID=3004;POINT(2313683.65%204641312.29),100,meters)
```

Note that regardless of the CRS being defined in the request, this PR makes pygeoapi ensure that it will always use the CRS of the underlying collection (as defined in its `storage_crs` configuration parameter), converting the filter geometry to the correct CRS if needed.

# Related issue / discussion
- fixes #1488

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/util.py
+++ b/pygeoapi/util.py
@@ -45,7 +45,7 @@ from decimal import Decimal
 from enum import Enum
 import pathlib
 from pathlib import Path
-from typing import Any, IO, Union, List, Callable
+from typing import Any, IO, Union, List, Optional, Callable
 from urllib.parse import urlparse
 from urllib.request import urlopen
 
@@ -883,25 +883,54 @@ def bbox2geojsongeometry(bbox: list) -> dict:
 
 def modify_pygeofilter(
         ast_tree: pygeofilter.ast.Node,
-        storage_crs,
+        *,
+        storage_crs_uri: Optional[str] = None,
+        geometry_column_name: Optional[str] = None
 ) -> pygeofilter.ast.Node:
     """
-    Prepare the input pygeofilter for querying the database.
+    Modifies the input pygeofilter with information from the provider.
 
-    Returns a new ``pygeofilter.ast.Node`` object that can be used for
-    querying the database.
+    :param ast_tree: `pygeofilter.ast.Node` representing the
+    already parsed pygeofilter expression
+    :param storage_crs_uri: An optional string containing the URI of
+    the provider's storage CRS
+    :param geometry_column_name: An optional string containing the
+    actual name of the provider's geometry field
+    :returns: A new pygeofilter.ast.Node, with the modified filter
+    expression
+
+    This function modifies the parsed pygeofilter that contains the raw
+    filter expression provided by an external client. It performs the
+    following modifications:
+
+    - if the filter includes any spatial coordinates and they are being
+      provided in a different CRS from the provider's storage CRS, the
+      corresponding geometries are transformed into the storage CRS
+
+    - if the filter includes the generic 'geometry' name as a reference to
+      the actual geometry of features, it is replaced by the actual name
+      of the geometry field, as specified by the provider
+
     """
     new_tree = deepcopy(ast_tree)
-    _inplace_transform_filter_geometries(new_tree, storage_crs)
+    if storage_crs_uri:
+        storage_crs = get_crs_from_uri(storage_crs_uri)
+        _inplace_transform_filter_geometries(new_tree, storage_crs)
+    if geometry_column_name:
+        _inplace_replace_geometry_filter_name(new_tree, geometry_column_name)
     return new_tree
 
 
 def _inplace_transform_filter_geometries(
         node: pygeofilter.ast.Node,
-        storage_crs
+        storage_crs: pyproj.CRS
 ):
     """
-    Recursively traverse node tree and convert any coordinates to the appropriate CRS.
+    Recursively traverse node tree and convert coordinates to the storage CRS.
+
+    This function modifies nodes in the already-parsed filter in order to find
+    any geometry literals that may be used in the filter and, if necessary,
+    proceeds to convert spatial coordinates to the CRS used by the provider.
     """
     try:
         sub_nodes = node.get_sub_nodes()
@@ -909,8 +938,11 @@ def _inplace_transform_filter_geometries(
         pass
     else:
         for sub_node in sub_nodes:
-            is_geometry_node = isinstance(sub_node, pygeofilter.values.Geometry)
+            is_geometry_node = isinstance(
+                sub_node, pygeofilter.values.Geometry)
             if is_geometry_node:
+                # NOTE: We specify a default CRS using a URI of type URN
+                # because this is what pygeofilter uses internally too
                 crs = get_crs_from_uri(
                     sub_node.geometry.get(
                         'crs', {}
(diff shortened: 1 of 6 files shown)


ricardogsilva:
The failing CI is unrelated to the changes contained in this PR, as mentioned in #1490

walkermatt:
@ricardogsilva Hi :wave:. I think the standards based approach to specifying the CRS of geometries in a CQL filter is via the `filter-crs` parameter as per https://portal.ogc.org/files/96288#filter-filter-crs.

ricardogsilva:
hi @walkermatt thanks for the pointer :+1: 

I'll take a look at the spec and implement accordingly

walkermatt:
@ricardogsilva fantastic. Thanks for implementing support for EWKT in [geopython/pygeofilter](https://github.com/geopython/pygeofilter), it's a great help.

ricardogsilva:
@walkermatt I've now updated this PR to use a `filter-crs` query parameter, as mentioned in the OAPIF - Part 3 spec you mentioned

walkermatt:
> @walkermatt I've now updated this PR to use a `filter-crs` query parameter, as mentioned in the OAPIF - Part 3 spec you mentioned

I've just updated my local copy of pygeoapi to use this PR branch and it's working well for me. I have a table in Postgres/PostGIS in EPSG:27700 with `storage_crs` set to `http://www.opengis.net/def/crs/EPSG/0/27700`. I'm now able to use the `WITHIN` CQL function like so:

```
http://localhost:5000/collections/licensed_area/items?org=example-com&f=json&filter=WITHIN(POINT(558760.2888089775%20139992.53573017253),geometry)&filter-crs=http://www.opengis.net/def/crs/EPSG/0/27700&crs=http://www.opengis.net/def/crs/EPSG/0/27700&limit=1
```

pygeoapi/util.py
```diff
@@ -886,3 +879,131 @@ def bbox2geojsongeometry(bbox: list) -> dict:
...
+                # NOTE2: We specify a default CRS using a URI of type URN
+                # because this is what pygeofilter uses internally too
+
+                crs_urn_provided_in_ewkt = sub_node.geometry.get(
+                    'crs', {}).get('properties', {}).get('name')
+                if crs_urn_provided_in_ewkt is not None:
+                    crs = get_crs_from_uri(crs_urn_provided_in_ewkt)
+                else:
+                    crs = filter_crs
+                print(f"{crs=}")
```
francbartoli, line 967: Can you use the logger instead of the print function?

francbartoli (changes requested):

pygeoapi/util.py
ricardogsilva: Those were not needed, so I removed them

tomkralidis (approved):

francbartoli:
@ricardogsilva found a small bug if the SRID is explicitly sent in the EWKT:

```shell
# no CRS specified - pygeoapi assumes OGC:CRS84 by default
curl http://localhost:5000/collections/beni_confiscati/items?\
    filter=DWITHIN(geometry,SRID=4326;POINT(12.626000006069875 41.8108000011414),100,meters)
```
 returns `0` results, even if the recommended query parameter `&filter-crs=https://www.opengis.net/def/crs/OGC/1.3/CRS84`  is added to the request

PR #1496, opened 2024-01-15
Log exception detail when debugging a process

walkermatt:
# Overview

Swap `LOGGER.error` for `LOGGER.exception` so that details of an exception that occurs during process execution are logged including the traceback to make debugging easier.

# Related issue / discussion

n/a

# Dependency policy (RFC2)

- [ x] I have ensured that this PR meets [RFC2](https://pygeoapi.io/development/rfc/2) requirements

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute this enhancment to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/pygeoapi/process/manager/dummy.py
+++ b/pygeoapi/process/manager/dummy.py
@@ -104,7 +104,7 @@ class DummyManager(BaseManager):
                 'description': 'Error updating job'
             }
             current_status = JobStatus.failed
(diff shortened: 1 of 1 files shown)


walkermatt:
@tomkralidis thanks, I think that's my smallest PR ever 😁

Issue #1462, opened 2023-12-28
pygeoapi not returning a proper TMS definition


doublebyte1:
**Description**
[Requirements class 2: "Tileset"](https://docs.ogc.org/is/20-057/20-057.html#toc30) states that the Tile Matrix Set Definition needs to be retrieved as application/json.
Unfortunately, GitHub hosted definitions won't satisfy this requirement, as GitHub does not provide content negotiation.

See here: https://github.com/geopython/pygeoapi/blob/master/pygeoapi/models/provider/base.py#L89

**Steps to Reproduce**

-  Publish a tile dataset, and [choose "default" as metadata format](https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_master/local.config.yml#L158)
- Access the tiles resource page and search for the tilesets "links" element. A GitHub repository is presented on the "href" element, with content type "application/json" (see sample bellow).

```
              {
                  "type":"application/json",
                  "rel":"http://www.opengis.net/def/rel/ogc/1.0/tiling-scheme",
                  "title":"WorldCRS84QuadTileMatrixSet definition (as JSON)",
                  "href":"https://raw.githubusercontent.com/opengeospatial/2D-Tile-Matrix-Set/master/registry/json/WorldCRS84Quad.json"
              },`

```
**Expected behavior**
Instead of pointing to a GitHub repository, the link should point to an endpoint capable of returning a JSON response, using content negotiation. This could be a TMS definition hosted by pygeoapi.

**Screenshots/Tracebacks**

![Screenshot from 2023-12-28 20-17-24](https://github.com/geopython/pygeoapi/assets/1038897/73d32407-a3c9-4e11-9624-a7a05d80c1ac)

**Additional context**
https://github.com/opengeospatial/ogcapi-tiles/issues/164
https://github.com/geopython/pygeoapi/issues/1262

PR #1497, opened 2024-01-16
Addressing Requirements class 2 (Requirements Class Tileset) of the OGC API Tiles Standard

doublebyte1:
# Overview
According to [section 8.3 ](https://docs.ogc.org/is/20-057/20-057.html#toc33) of the OGC API TIles - Standard: 

- The TileMatrixSets endpoint should be advertised in the landing page
- The TileMatrixSets endpoint retrieves a list of links to the descriptions of the tile matrix sets supported by the OGC Web API
- The TileMatrixSet endpoint retrieves the full description of a tile matrix set supported by Web API based servers and client applications following the schema described in the OGC Two Dimensional Tile Matrix Set and Tile Set Metadata 2.0 Standard

This PR adds the `TileMatrixSets` endpoint, on the root path, as well as our, now hosted, definitions for the `WorldCRS84Quad` and `WebMercatorQuad`. For instance:

http://localhost:5000/TileMatrixSets
http://localhost:5000/TileMatrixSets/WorldCRS84Quad
http://localhost:5000/TileMatrixSets/WebMercatorQuad

(the format of the resources can be retrieved using content negotiation, or the `f` parameter).

In addition, a link with "rel" type "http://www.opengis.net/def/rel/ogc/1.0/tiling-scheme", which points to the hosted tile matrix set definition was added at the tiles collection endpoint and tile metadata endpoint. For instance:

http://localhost:5000/collections/lakes/tiles?f=json
http://localhost:5000//collections/ne_110m_populated_places_simple/tiles/WorldCRS84Quad/metadata?f=json

# Related issue / discussion
https://github.com/geopython/pygeoapi/issues/1462

# Additional information
OGC API - Tiles  (core): https://docs.ogc.org/is/20-057/20-057.html
OGC Two Dimensional Tile Matrix Set and Tile Set Metadata: https://docs.ogc.org/is/17-083r4/17-083r4.html#:~:text=The%20OGC%20Two%20Dimensional%20Tile,in%20OGC%2018%2D005r5%20Abstract

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


pygeoapi/models/provider/base.py
```diff
@@ -74,34 +74,692 @@ class TileMatrixSetEnumType(BaseModel):
...
         crs="http://www.opengis.net/def/crs/OGC/1.3/CRS84",
-        tileMatrixSetDefinition=
-            {
-                'type': 'application/json',
-                'rel': 'http://www.opengis.net/def/rel/ogc/1.0/tiling-scheme',
-                'title': 'WorldCRS84QuadTileMatrixSet definition (as JSON)',
-                'href': 'https://raw.githubusercontent.com/opengeospatial/2D-Tile-Matrix-Set/master/registry/json/WorldCRS84Quad.json'  # authoritative TMS definition
-            }
-        )
+        title="World Mercator WGS84 (ellipsoid)",
```
tomkralidis, line 87: Is this entire definition available on schemas.opengis.net? Should we be baking artifacts (with values) in the code?  Is this better off in a resource file of some sort?

tomkralidis (changes requested):
Thanks @doublebyte1! See comments.  As well, we need the associated endpoint updates (`pygeoapi/flask_app.py`) in the starlette and django equivalents).

pygeoapi/models/provider/base.py
francbartoli: @tomkralidis if the goal is to get rid of a `pydantic` definition I would be inclined to keep this since it is aligned with the rest of the tiles implementation and procrastinate the removal of the whole pydantic code once we have the consensus of the community.

pygeoapi/flask_app.py
```diff
@@ -155,6 +155,36 @@ def conformance():
     return get_response(api_.conformance(request))
 
 
+@BLUEPRINT.route('/TileMatrixSets/WorldCRS84Quad')
```
francbartoli, line 158: Can we manage the route parameter dynamically like `@BLUEPRINT.route('/TileMatrixSets/<TileMatrixSetId>')`?

pygeoapi/flask_app.py
```diff
@@ -155,6 +155,36 @@ def conformance():
...
+    """
+    OGC API WebMercatorQuad endpoint
+
+    :returns: HTTP response
+    """
+    return get_response(api_.tilematrixset(request))
+
+
+@BLUEPRINT.route('/TileMatrixSets')
+def TileMatrixSets():
```
francbartoli, line 179: I would rename to a lowercase `def get_tilematrix_sets():` function name

pygeoapi/flask_app.py
```diff
@@ -155,6 +155,36 @@ def conformance():
     return get_response(api_.conformance(request))
 
 
+@BLUEPRINT.route('/TileMatrixSets/WorldCRS84Quad')
+def WorldCRS84Quad():
```
francbartoli, line 159: with a dynamic approach the name could be `def get_tilematrix_set(tileMatrixSetId=None):`

francbartoli (changes requested):

pygeoapi/models/provider/base.py
doublebyte1: @tomkralidis the only normative definition is on the OGC definition server, but unfortunately [that is not available](https://github.com/opengeospatial/NamingAuthority/issues/198). It seems everyone else is hosting definitions on their server, as recommendation of the spec.

doublebyte1:
> Thanks @doublebyte1! See comments. As well, we need the associated endpoint updates (`pygeoapi/flask_app.py`) in the starlette and django equivalents).

@tomkralidis I added the routes for starlette, but unfortunately I am unable to run the django server. I opened an issue here: https://github.com/geopython/pygeoapi/issues/1509

pygeoapi/models/provider/base.py
doublebyte1: https://test.cubewerx.com/cubewerx/cubeserv/demo/ogcapi/Daraa/tileMatrixSets
https://maps.gnosis.earth/ogcapi/tileMatrixSets
https://demo.ldproxy.net/daraa/tileMatrixSets/

doublebyte1: @tomkralidis routes for django were added here: https://github.com/geopython/pygeoapi/pull/1497/commits/5931e1a432d7a2843d8d4584cac5447ab98349a3

tomkralidis (approved):
Merging, but I would prefer as a next step that we move the tileMatrixes in `pygeoapi/models/provider/base.py` into a JSON configuration and load at runtime, since the configuration looks like "data" more than something to be "baked" in code.

In the meantime, moving ahead for OATiles compliance.  Great work!

Issue #1502, opened 2024-01-17
Address requirement 6A of OGC API - Tiles Core Requirements for the MVT-elastic provider


doublebyte1:
**Description**
[Requirement 6 A](https://docs.ogc.org/is/20-057/20-057.html#toc27) of the Core Requirements of OGC API - Tiles, states that:

`If the path parameter values tileMatrix, tileRow, tileCol for a tile request are out-of-range (outside the tile matrix set or tile matrix set limits of the resource), the HTTP response SHALL use a status code 404 or a 400. The response can also contain an exception report.`

Although this works for the MVT-tippecanoe provider, in the case of the MVT-elastic provider, pygeoapi is returning 500.
This difference in behaviour is related to how the backend providers handler the out of range request.

**Steps to Reproduce**

`curl -I https://emotional.byteroad.net/collections/hex350_grid_cardio_1920/tiles/WorldCRS84Quad/10/2147483647/2147483647\?f\=mvt

HTTP/2 500
`
**Expected behavior**
The server should return a 404, or 400 status code in the response.

PR #1503, opened 2024-01-17
Return a 400 status code for an "out-of-range" get tile request

doublebyte1:
# Overview
The MVT-elastic backend provider returns a 400 status code in response to a request for a tile out of range (e.g.: outside the limits of the TMS). This would satisfy req 6 A of the OGC API - Tiles standard (requirements class core). However, pygeoapi is transforming this into a 500 error.
This PR introduces a try/ except block on the MVT-elastic class, and makes sure that the Exception that arrives to the API class is appropriated. For that purpose, I reused the existing "ProviderInvalidQueryError", which is close enough in terms of semantics, but I could also have created a new Provider exception.

# Related issue / discussion
https://github.com/geopython/pygeoapi/issues/1502

# Additional information
OGC API - Tiles standard: https://docs.ogc.org/is/20-057/20-057.html

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -73,7 +73,7 @@ from pygeoapi.process.manager.base import get_manager
 from pygeoapi.plugin import load_plugin, PLUGINS
 from pygeoapi.provider.base import (
     ProviderGenericError, ProviderConnectionError, ProviderNotFoundError,
-    ProviderTypeError, ProviderInvalidQueryError)
+    ProviderTypeError)
 from pygeoapi.models.provider.base import TilesMetadataFormat
 
 from pygeoapi.models.cql import CQLModel
@@ -2759,11 +2759,6 @@ class API:
             return self.get_exception(
                 HTTPStatus.BAD_REQUEST, headers, format_,
                 'InvalidParameterValue', msg)
-        except ProviderInvalidQueryError:
-            msg = 'Invalid tile request'
-            return self.get_exception(
-                HTTPStatus.BAD_REQUEST, headers, format_,
-                'InvalidRequest', msg)
         except ProviderGenericError as err:
             LOGGER.error(err)
             return self.get_exception(
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -73,7 +73,7 @@ from pygeoapi.process.manager.base import get_manager
 from pygeoapi.plugin import load_plugin, PLUGINS
 from pygeoapi.provider.base import (
     ProviderGenericError, ProviderConnectionError, ProviderNotFoundError,
-    ProviderTypeError)
+    ProviderTypeError, ProviderInvalidQueryError)
 from pygeoapi.models.provider.base import TilesMetadataFormat
 
 from pygeoapi.models.cql import CQLModel
@@ -2759,6 +2759,11 @@ class API:
             return self.get_exception(
                 HTTPStatus.BAD_REQUEST, headers, format_,
                 'InvalidParameterValue', msg)
+        except ProviderInvalidQueryError:
+            msg = 'Invalid tile request'
+            return self.get_exception(
+                HTTPStatus.BAD_REQUEST, headers, format_,
+                'InvalidRequest', msg)
         except ProviderGenericError as err:
             LOGGER.error(err)
             return self.get_exception(
--- a/pygeoapi/provider/mvt_elastic.py
+++ b/pygeoapi/provider/mvt_elastic.py
@@ -170,14 +170,11 @@ class MVTElasticProvider(BaseMVTProvider):
                     resp = session.get(f'{base_url}/{layer}/{z}/{y}/{x}{url_query}')  # noqa
                     resp.raise_for_status()
                     return resp.content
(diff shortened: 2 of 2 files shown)


francbartoli (commented):
@doublebyte1 does the problem apply to the rest of the MVT provider implementations like for example `mvt_tippecanoe`?

pygeoapi/api.py
```diff
@@ -2759,6 +2759,11 @@ def get_collection_tiles_data(
             return self.get_exception(
                 HTTPStatus.BAD_REQUEST, headers, format_,
                 'InvalidParameterValue', msg)
+        except ProviderInvalidQueryError:
```
totycro, line 2762: I think this would even work without this code change. `ProviderInvalidQueryError` is a subclass of `ProviderGenericError`, so it's also caught in the `except` block below, and since `ProviderInvalidQueryError` has set `http_status_code` to 400, this will effectively produce a 400 response

pygeoapi/provider/mvt_elastic.py
```diff
@@ -162,11 +164,20 @@ def get_tiles(self, layer=None, tileset=None,
...
-                resp.raise_for_status()
-                return resp.content
+            try:
+                with requests.Session() as session:
+                    session.get(base_url)
+                    resp = session.get(f'{base_url}/{layer}/{z}/{y}/{x}{url_query}')  # noqa
+                    resp.raise_for_status()
+                    return resp.content
+            # Client is sending an invalid request
+            except requests.exceptions.HTTPError as e:
```
totycro, line 174: By the way this could also have other causes such as an internal server error at the server side. To be more specific about client side errrors, we could check if `400 <= e.response.status_code < 500`

totycro (approved):

doublebyte1:
> @doublebyte1 does the problem apply to the rest of the MVT provider implementations like for example `mvt_tippecanoe`?

MVT-tippecanoe looks for the tile on disk, instead of issuing a request to a server. When it does not find it, it throws a FileNotFound/404 error, which complies with the spec.

https://github.com/geopython/pygeoapi/blame/e1fe05c1c16e40ab8bf6d07540e86d6f78b14172/pygeoapi/provider/mvt_tippecanoe.py#L174-L175

pygeoapi/provider/mvt_elastic.py
```diff
@@ -162,11 +164,17 @@ def get_tiles(self, layer=None, tileset=None,
...
-                return resp.content
+            try:
+                with requests.Session() as session:
+                    session.get(base_url)
+                    resp = session.get(f'{base_url}/{layer}/{z}/{y}/{x}{url_query}')  # noqa
+                    resp.raise_for_status()
+                    return resp.content
+            except requests.exceptions.RequestException as e:
+                LOGGER.debug(e)
+                if resp.status_code <= 500:
```
totycro, line 175: I think `< 500` is meant here instead of `<= 500`, otherwise internal server errors would be reported as invalid client errors

doublebyte1: Indeed. Well spotted!

doublebyte1: @totycro fixed on my latest commit

totycro (approved):

francbartoli:
> > @doublebyte1 does the problem apply to the rest of the MVT provider implementations like for example `mvt_tippecanoe`?
> 
> @francbartoli MVT-tippecanoe looks for the tile on disk, instead of issuing a request to a server. When it does not find it, it throws a FileNotFound/404 error, which complies with the spec.
> 
> https://github.com/geopython/pygeoapi/blame/e1fe05c1c16e40ab8bf6d07540e86d6f78b14172/pygeoapi/provider/mvt_tippecanoe.py#L174-L175
> 
> `curl -I https://demo.pygeoapi.io/master/collections/lakes/tiles/WorldCRS84Quad/10/2147483647/2147483647?f=mvt`

@doublebyte1 if I'm not wrong pygeoapi supported tippecanoe built tiles which might be hosted remotely on s3-like (i.e. Minio, etc) buckets

doublebyte1:
> > > @doublebyte1 does the problem apply to the rest of the MVT provider implementations like for example `mvt_tippecanoe`?
> > 
> > 
> > @francbartoli MVT-tippecanoe looks for the tile on disk, instead of issuing a request to a server. When it does not find it, it throws a FileNotFound/404 error, which complies with the spec.
> > https://github.com/geopython/pygeoapi/blame/e1fe05c1c16e40ab8bf6d07540e86d6f78b14172/pygeoapi/provider/mvt_tippecanoe.py#L174-L175
> > `curl -I https://demo.pygeoapi.io/master/collections/lakes/tiles/WorldCRS84Quad/10/2147483647/2147483647?f=mvt`
> 
> @doublebyte1 if I'm not wrong pygeoapi supported tippecanoe built tiles which might be hosted remotely on s3-like (i.e. Minio, etc) buckets

@francbartoli good point. It turns out that is not working now. I'll have a look later

Issue #1506, opened 2024-01-18
Serving tiles from a url on mvt-tippecanoe is not working


doublebyte1:
**Description**
The [docs](https://docs.pygeoapi.io/en/stable/data-publishing/ogcapi-tiles.html?highlight=tippecanoe#mvt) state that the mvt-tippecanoe provider can serve tiles from a data path on disk, or an url. However, when serving tippecanoe tiles from a url, it is not possible to publish tiles or generate tile metadata in the tilejson specification, 

**Steps to Reproduce**
- Render tippecanoe tiles and place them on a web server or a bucket
- Configure pygeoapi to publish these tiles
```
            - type: tile
              name: MVT-tippecanoe
              data: http://localhost:3000/data/lakes/{z}/{x}/{y}.pbf
              options:
                bounds: [[-124.953634,-16.536406],[109.929807,66.969298]]
                zoom:
                    min: 0
                    max: 11
                schemes:
                    - WorldCRS84Quad
              format: 
                    name: pbf
                    mimetype: application/vnd.mapbox-vector-tile

```
- Retrieve a tile: http://localhost:5000/collections/lakes/tiles/WorldCRS84Quad/0/0/0?f=mvt
- Retrieve metadata in the tilejson format: http://localhost:5000/collections/lakes/tiles/WorldCRS84Quad/metadata?f=tilejson

**Expected behavior**
Both requests should return status code: 200.

doublebyte1:
https://github.com/geopython/pygeoapi/pull/1507 addresses serving tiles from an url, but the metadata part is still not working. Maybe @PascalLike can have a look? (:

PR #1507, opened 2024-01-18
Give support to serving mvt-tippecanoe tiles from an url

doublebyte1:
# Overview
This PR fixes a regression bug, introduced by the MVT class refactoring.

# Related issue / discussion
https://github.com/geopython/pygeoapi/issues/1506

# Additional information
The response status codes were fixed, to address the 5 and 6A core requirements from the OGC API - TIles standard.
https://docs.ogc.org/is/20-057/20-057.html

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/mvt_tippecanoe.py
+++ b/pygeoapi/provider/mvt_tippecanoe.py
@@ -29,12 +29,16 @@
 
 import json
 import logging
+import requests
 from pathlib import Path
 from urllib.parse import urlparse
+import os.path
 
 from pygeoapi.provider.tile import (
     ProviderTileNotFoundError)
-from pygeoapi.provider.base import ProviderConnectionError
+from pygeoapi.provider.base import (ProviderConnectionError,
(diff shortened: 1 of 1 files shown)


pygeoapi/provider/mvt_tippecanoe.py
```diff
@@ -69,20 +73,29 @@ def __init__(self, BaseMVTProvider):
         if is_url(self.data):
             url = urlparse(self.data)
             baseurl = f'{url.scheme}://{url.netloc}'
-            param_type = '?f=mvt'
+            extension = os.path.splitext(url.path)[-1]  # e.g. ".pbf"
```
francbartoli, line 76: Can we use `Path` instead of `os.path` since it was already loaded?

francbartoli (changes requested):

francbartoli (approved):

Issue #593, opened 2020-12-21
Dev style adoption


francbartoli:
**Is your feature request related to a problem? Please describe.**
It has been requested several times in different places to adopt a clear and concise set of best-practices for python developers. Let's open the discussion with some proposals about new tools.
For reference, there is one previous issue #288

**Describe the solution you'd like**
We are already using [Flake8](https://flake8.pycqa.org/en/latest/), [pytest](https://docs.pytest.org/en/latest/) and [sphinx](https://www.sphinx-doc.org/en/master/).

We'd like to evaluate the use of the following eventually:
- [Black](https://github.com/psf/black) - _Uncompromisingly_ eliminating all ambiguity about where to put whitespace in python code. Black has moved under the Python Software Foundation umbrella in 2019, which is a strong endorsement.
- [Pylint](https://www.pylint.org/) - It gives specific feedback to improve the code and how to better comply with conventions.
- [pre-commit](https://pre-commit.com/) - Checking for linter errors even one more time before committing to your local git clone and pushing to a PR.

More importantly [Poetry](https://github.com/python-poetry/poetry) and its use of [pyproject.toml PEP 518](https://www.python.org/dev/peps/pep-0518/) which could be, in the most optimistic future, the Python's npm to evolve the tooling for dependencies management.

**Describe alternatives you've considered**
Leave the tooling as it is

**Additional context**

jorgejesus:
workflow guidelines and code structure recommendations: https://github.com/geopython/pygeoapi/wiki/Code-development-workflow

solismaam:
@francbartoli has there been any progress on this matter?

Also, adding a simple [devcontainer](https://code.visualstudio.com/docs/remote/containers) configuration could be useful for this project, so that the contributors participating in the project would have an identical development environment with all the necessary binaries, libraries and utilities.

frafra:
Might I suggest [ruff](https://github.com/charliermarsh/ruff) instead of pylint?
[isort](https://pycqa.github.io/isort/) is also nice.

I am using black+isort+ruff+pre-commit for my projects. Here is the template: https://github.com/NINAnor/nina-template-python
Please notice that pyproject.toml includes a couple of lines to make them work together nicely.

ricardogsilva:
After many frustrating commits that I had to immediately amend with a message like 'fixed flake8 issue' after seeing CI turn red, I decided to add a `.pre-commit-config.yaml` file with a standard `flake8` hook. Proposed as #1508.

[...]

PR #1508, opened 2024-01-19
Added pre-commit config file

ricardogsilva:
# Overview

This PR adds a [pre-commit](https://pre-commit.com/) config file, thus enabling devs to opt in to use it for performing automated checks whenever they try to `git commit` some change.

The main motivations for this PR are the existing discussion on #593 and the fact that I always forget to run `flake8` before I push some change and I only find out about it when the CI check becomes red :angry: 

This PR thus makes the following changes:

- Adds a `.pre-commit-config.yaml` file to the root of the repo. The contents of this file only contain the default pre-commit checks (_i.e._ check for trailing whitespace, large files, etc) and an additional hook for `flake8`, [configured as mentioned in their docs](https://flake8.pycqa.org/en/latest/user/using-hooks.html#usage-with-the-pre-commit-git-hooks-framework)
- Updates the pygeoapi development docs to mention the existence of this pre-commit integration and how to use it
- Unignores the pre-commit config file from `.gitignore` as this file is useful to have on the repo - provided that there is a favorable review to this PR
- Does a first pass of `pre-commit run --all-files` that basically updated a bunch of files that had slight inconsistencies, like trailing whitespaces, no blank line at the end, etc - This PR does should not modify any logic in the pygeoapi codebase, and if it somehow does, it would be by accident.

Including a pre-commit config file in the pygeoapi repo does not make it mandatory that other devs start using pre-commit - this is an opt-in feature.

# Related issue / discussion
- #593

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


ricardogsilva:
@tomkralidis added a regex to only check `.py`, `.sh` and `.sql` files

PR #1514, opened 2024-01-24
Parametrize api-related tests

ricardogsilva:
# Overview
This PR uses pytest's [parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize) feature in order to break apart some of the test cases in `test_api`. It aims to be neutral in terms of functionality:

- Does not remove any existing test
- Does not add any new test

It merely refactors some of the larger test cases into smaller units in such a way as to not have more than a single request per run - This makes it easier to fix failing tests, as it becomes clearer to spot errors. It also becomes possible to run these individually, like this:

```sh
# pytest tests/test_api.py::function_name[paramspec-or-id]
tests/test_api.py::test_get_collection_items[closed-date-interval]
```

As a side effect, the reported number of tests in `test_api.py` count grew from 43 to 121 - this only means that we can now run the same tests as before, but with finer granularity.

# Related issue / discussion
I had attempted to refactor some tests previously, in #1399 and was advised to create a separate PR with such changes - check: https://github.com/geopython/pygeoapi/pull/1399#pullrequestreview-1733180330

Additionally, if this is seen as a valuable change, I would be willing to gradually convert other large test cases into smaller units in a similar fashion.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -802,110 +802,122 @@ def test_describe_collections_json_ld(config, api_):
     assert rsp_headers['Content-Language'] == 'en-US'
 
 
-def test_get_collection_items(config, api_):
-    req = mock_request()
-    rsp_headers, code, response = api_.get_collection_items(req, 'foo')
+@pytest.mark.parametrize(
+    [
+        'dataset_name',
+        'request_params',
+        'expected_code',
+        'expected_headers',
+        'expected_num_features',
+    ],
+    [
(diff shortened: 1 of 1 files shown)


ricardogsilva:
I'm withdrawing this PR, leaving pygeoapi test layout as it is

PR #1516, opened 2024-01-25
Replaced geoalchemy with geoalchemy2 and added numpy as explicit dependency

### LINKED ISSUE #1426: Make numpy a first-level dependency
Pygeoapi currently imports and uses numpy in xarray providers but it does not state it as a direct dependency. Instead it relies on numpy being installed implicitly as it is a requirement of some of the other pygeoapi dependencies.

Since numpy is directly imported in the pygeoapi codebase, we ought to make it a first-level dependency and thus include in the relevant requirements file, and also in the dockerfile.

### LINKED ISSUE #1515: replace geoalchemy with geoalchemy2 as a dependency
**Description**
pygeoapi is currently depending on [GeoAlchemy2](https://geoalchemy-2.readthedocs.io/en/latest/), as can be seen in:

https://github.com/geopython/pygeoapi/blob/e1fe05c1c16e40ab8bf6d07540e86d6f78b14172/pygeoapi/provider/postgresql.py#L54-L56

However, `geoalchemy2` is not listed as a pygeoapi dependency in any of the existing requirements files. It is being pulled in as a secondary dependency, as it is required by `pygeofilter[backend-sqlalchemy]`.

Moreover, pygeoapi declares to depend on [GeoAlchemy](https://github.com/geoalchemy/geoalchemy/tree/master#geoalchemy) in both:

- https://github.com/geopython/pygeoapi/blob/e1fe05c1c16e40ab8bf6d07540e86d6f78b14172/requirements-docker.txt#L2
- https://github.com/geopython/pygeoapi/blob/e1fe05c1c16e40ab8bf6d07540e86d6f78b14172/requirements-provider.txt#L7

But `geoalchemy` is not used anywhere in pygeoapi code and should thus be removed from the requirements files.

ricardogsilva:
# Overview
This PR replaces the dependency on `geoalchemy` with `geoalchemy2`, as discussed in #1515 and it also adds `numpy` as an explicit top-level dependency, as discussed in #1426

# Related issue / discussion
- Fixes #1515 
- Fixes #1426 

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/requirements-docker.txt
+++ b/requirements-docker.txt
@@ -1,5 +1,5 @@
 elasticsearch-dsl<8
-geoalchemy
+geoalchemy2
 pygeofilter[backend-sqlalchemy]
 pygeoif
 pygeometa
--- a/requirements-provider.txt
+++ b/requirements-provider.txt
@@ -4,8 +4,9 @@ elasticsearch
 elasticsearch-dsl
 fiona
 GDAL<=3.8.4
-geoalchemy
+geoalchemy2
 netCDF4
+numpy
 oracledb
 pandas
 psycopg2


ricardogsilva:
Failing CI is unrelated to the changes in this PR

ricardogsilva:
the CI failures are being caused by the issue described in #1517

tomkralidis:
Thanks @ricardogsilva; comments:

1. can you rebase against current master?
2. given numpy is used as part of the xarray provider, and that xarray is listed in `requirements-provider.txt`, suggest moving the numpy dependency to `requirements-provvider.txt`

ricardogsilva:
@tomkralidis rebased

Issue #1517, opened 2024-01-25
wfs provider tests fail occasionally


ricardogsilva:
**Description**
Tests for the WFS provider sometimes fail due to issues when communicating with third-party servers. These failures are unrelated to the pygeoapi codebase but still cause CI to fail.

**Steps to Reproduce**
This can be tested in isolation by replicating the behavior of the test in `tests/test_ogr_wfs_provider.py::test_get_ms` and then execute it in a sequential loop a bunch of times, like this:

```python
from pygeoapi.provider import ogr

provider = ogr.OGRProvider(
    {
        'name': 'OGR', 
        'type': 'feature', 
        'data': {
            'source_type': 'WFS', 
            'source': 'WFS:https://demo.mapserver.org/cgi-bin/wfs', 
            'source_capabilities': { 'paging': True }, 
            'source_options': { 
                'OGR_WFS_VERSION': '2.0.0', 
                'OGR_WFS_LOAD_MULTIPLE_LAYER_DEFN': 'NO'
            }, 
            'gdal_ogr_options': { 
                'GDAL_CACHEMAX': '64', 
                'GDAL_HTTP_VERSION': '1.1',
                'GDAL_HTTP_UNSAFESSL': 'YES', 
                'CPL_DEBUG': 'NO'
            },
        }, 
        'id_field': 'gml_id', 
        'layer': 'cities' 
    }
)

for i in range(20):
    r = provider1.get("cities.8338")
    print(r["id"])
    print(f'----- {i} -----')
```

Most of the time this will work OK, but occasionally you will get an error like this:

```python

cities.8338   
----- 0 -----                                                                                             
cities.8338   
----- 1 -----                                                                                             
cities.8338                                                                                               
----- 2 ----- 
cities.8338
----- 3 -----
Error Number: 1, Type: Failure, Msg: HTTP error code : 500
Error Number: 1, Type: Failure, Msg: Error returned by server : HTTP error code : 500 (0)
<built-in function Layer_SetAttributeFilter> returned a result with an error set
---------------------------------------------------------------------------
ProviderConnectionError                   Traceback (most recent call last)
File /datadisk/dev/pygeoapi/pygeoapi/provider/ogr.py:894, in GdalErrorHandler.handler(self, err_level, err_num, err_msg)
    893         if str(i) in last_error:
--> 894             raise ProviderConnectionError(last_error)
    895 else:

ProviderConnectionError: Error returned by server : HTTP error code : 500 (0)

[...]

PR #1524, opened 2024-01-29
Removed encoding parameter from pygeoapi config

### LINKED ISSUE #1521: Do not provide an option to choose charset for compressed GZIP responses
pygeoapi currently provides an `encoding` configuration key, which is briefly mentioned in the docs as being the _"default server encoding"_. Looking through the code, it seems this parameter is only used when sending back gzipped responses. Responses which are not gzipped are not being encoded in order to comply with the value of this `encoding` config parameter and thus use `utf-8` by default. So it seems the proper name for this parameter would be `gzip_encoding`, with a description of _"server encoding for gzipped responses"_.

In my opinion:

- It seems inconsistent to only allow choosing a charset when responses are being gzipped;
- Letting clients ask for encoded responses by specifying the charset on the client side, with the `Accept-Charset` HTTP header [is not done by any browser, and is not recommended](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset). This means that clients are happy to receive responses in `utf-8`

I am proposing that the `encoding` parameter be removed from pygeoapi's configuration and that gzipped responses would also use the default encoding of `utf-8`.

ricardogsilva:
# Overview
This PR removes the `encoding` configuration parameter. As discussed in #1521, this parameter only affected the charset of gzipped responses and was thus not being used very consistently. This proposed implementation simply removes this parameter from pygeoapi config and therefore makes gzipped responses use the default charset of `utf-8`, which is what the normal, non-gzipped responses were already doing.

# Related issue / discussion
- Fixes #1521

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -43,7 +43,6 @@ For more information related to API design rules (the ``api_rules`` property in 
         port: 5000  # listening port for incoming connections
     url: http://localhost:5000/  # url of server
     mimetype: application/json; charset=UTF-8  # default MIME type
-    encoding: utf-8  # default server encoding
     language: en-US  # default server language
     locale_dir: /path/to/translations
     gzip: false # default server config to gzip/compress responses to requests with gzip in the Accept-Encoding header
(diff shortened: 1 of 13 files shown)


ricardogsilva:
CI error is not related to the changes in this PR - it is a manifestation of #1517

PR #1525, opened 2024-01-29
Removed host and port from pygeoapi configuration

### LINKED ISSUE #1523: Remove server host and port from pygeoapi configuration
pygeoapi configuration currently has the parameters:

- `server.bind.host`
- `server.bind.port` 

These are used to specify how to run the pygeoapi web application server. However, given that these are just server initialization parameters, in my opinion, they should not be part of the configuration object, but rather specified at runtime, when initializing the web application server.

These parameters are not used anywhere except in the code that initializes the web application (flask, starlette, django).

Moreover, these parameters are simply not used if for example I use `gunicorn` to launch pygeoapi instead of `pygeoapi serve`, which is already the case on pygeoapi's docker image:

https://github.com/geopython/pygeoapi/blob/8ffc354adfdd7f872bc053cbf9a64986c85128be/docker/entrypoint.sh#L44-L45

I would like to have these parameters removed from the pygeoapi configuration and passed as either env variables or CLI parameters when calling `pygeoapi serve`.

ricardogsilva:
This PR removes web application server initialization parameters from the pygeoapi configuration, specifically the `server.bind` section. This includes:

- `server.bind.host`
- `server.bind.port`

These now become initialization parameters, passed to `pygeoapi serve`. Included in this PR is a working implementation for all of:

- flask
- starlette
- django

`pygeoapi serve` now accepts two optional parameters:

- `host` - defaults to `"127.0.0.1"` - this is intentionally not set to `"0.0.0.0"` in order to prevent accidental exposure of private services.
- `port` - defaults to `5000`

# Related issue / discussion
- Fixes #1523 

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -38,10 +38,7 @@ For more information related to API design rules (the ``api_rules`` property in 
 .. code-block:: yaml
 
   server:
-    bind:
-        host: 0.0.0.0  # listening address for incoming connections
-        port: 5000  # listening port for incoming connections
-    url: http://localhost:5000/  # url of server
+    url: http://localhost:5000/  # public url of server
     mimetype: application/json; charset=UTF-8  # default MIME type
     encoding: utf-8  # default server encoding
     language: en-US  # default server language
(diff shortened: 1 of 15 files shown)


ricardogsilva:
Withdrawing this PR, as there seems to be appreciation for having these parameters in pygeoapi config

PR #1526, opened 2024-01-31
Get tiles metadata from remote URL

PascalLike:
# Overview
This fixes a regression on the Tippecanoe MVT provider when metadata JSON is exposed from a non-local URL. This is similar to what was done in this [PR](https://github.com/geopython/pygeoapi/pull/1507) by @doublebyte1 for tiles.

# Related issue / discussion
https://github.com/geopython/pygeoapi/issues/1506

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/mvt_tippecanoe.py
+++ b/pygeoapi/provider/mvt_tippecanoe.py
@@ -237,6 +237,34 @@ class MVTTippecanoeProvider(BaseMVTProvider):
             LOGGER.error(msg)
             raise ProviderConnectionError(msg)
 
+    def get_metadata_from_URL(self, service_metadata_url=None):
+        """
+        Gets the metadata JSON from disk or URL
+
+        :param service_metadata_url: local path or URL
+
+        :returns: the metadata JSON
+        """
+        if (not service_metadata_url):
+            service_metadata_url = self.service_metadata_url
+
+        metadata_json_content = ''
+        if isinstance(service_metadata_url, Path):
+            if service_metadata_url.exists():
+                with open(service_metadata_url, 'r') as md_file:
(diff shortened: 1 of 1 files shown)


doublebyte1 (approved):
Thank You @PascalLike ! Excellent job 👏🏽

pygeoapi/provider/mvt_tippecanoe.py
```diff
@@ -253,21 +281,20 @@ def get_html_metadata(self, dataset, server_url, layer, tileset,
...
+            content = MVTTilesJson(**metadata_json_content)
+            content.tiles = service_url
+            content.vector_layers = json.loads(
+                    metadata_json_content["json"])["vector_layers"]
+            metadata['metadata'] = content.model_dump()
+            # Some providers may not implement tilejson metadata
+            metadata['tilejson_url'] = f'{metadata_url}?f=tilejson'
+        except ProviderConnectionError:
+            # No vendor metadata JSON
+            pass
```
francbartoli, line 297: @PascalLike what about to log and raise the `ProviderConnectionError` exception rather than pass? Something like

```python
except ProviderConnectionError as e:
    logger.debug(e)
    raise
```

PascalLike: It raises the exception if tilejson is required. To raise the exception in this point too will block the functioning of the html page, which otherwise has the link to the metadata in OGC Tile Set Metadata format and the other primary information, which is the only very mandatory information. But I was undecided too. Do you think it's better to raise the exception here too? Thanks

francbartoli (approved):

Issue #1528, opened 2024-01-31
Django EDR implementation broken on non-instance endpoints


petergarnaes:
**Description**
When using the Django server and requesting EDR query endpoints without an instance id in the path (f.ex. `http://localhost:5000/collections/icoads-sst/position?coords=POINT(12.779895%2055.783523)&f=json`), the django server fails with a "get_collection_edr_query() missing 1 required positional argument: 'instance_id'"

**Steps to Reproduce**
 - Fresh clone and setup of pygeoapi
 - Copy the `icoads-sst` part under `resources` from [this test config](https://github.com/geopython/pygeoapi/blob/master/tests/pygeoapi-test-config.yml) into the config provided to pygeoapi
 - Start pygeoapi with django by running `pygeoapi serve --django`
 - Request a valid EDR query for the configured EDR provider, such as: `http://localhost:5000/collections/icoads-sst/position?coords=POINT(12.779895%2055.783523)&f=json`

**Expected behavior**
pygeoapi running on the django server should allow non-instance EDR queries and forward them to the given EDR provider, having complete parity with the other server providers like flask, starlette etc. 

**Screenshots/Tracebacks**
```
Watching for file changes with StatReloader
Internal Server Error: /collections/icoads-sst/position
Traceback (most recent call last):
  File "/.../pygeoapi/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/.../pygeoapi/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
TypeError: get_collection_edr_query() missing 1 required positional argument: 'instance_id'
[31/Jan/2024 13:45:01] [35;1m"GET /collections/icoads-sst/position?coords=POINT(12.779895%2055.783523)&f=json HTTP/1.1" 500 67919
```

**Environment**
 - OS: Ubuntu 20.04
 - Python version: 3.8.10
 - pygeoapi version: on commit 2d0fc5d
 - django version: 4.2.9

**Additional information**
The stack trace refers to [this function](https://github.com/geopython/pygeoapi/blob/2d0fc5df3e5a2cd9959d331023bfe9f9aa9071bd/pygeoapi/django_/views.py#L452-L474) which do in fact require an instance ID, even though it optional as per the EDR spec.

PR #1530, opened 2024-02-02
Django server: EDR sanity test

petergarnaes:
Regression test added for fix in #1549 

# Overview

# Related issue / discussion

#1528 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute bugfix to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/tests/test_django.py
+++ b/tests/test_django.py
@@ -6,8 +6,14 @@ from unittest import mock
 import django
 from django.test import Client
 
+from .util import get_test_file_path
 
-@mock.patch.dict(os.environ, {"DJANGO_SETTINGS_MODULE": "django_.settings"})
+
+@mock.patch.dict(os.environ, {
+    "DJANGO_SETTINGS_MODULE": "django_.settings",
+    "PYGEOAPI_CONFIG": get_test_file_path('pygeoapi-test-config.yml'),
+    "PYGEOAPI_OPENAPI": get_test_file_path('pygeoapi-test-openapi.yml')
+})
 @mock.patch.object(sys, "path", sys.path + ["./pygeoapi"])
 def test_django_landing_page_loads():
     django.setup()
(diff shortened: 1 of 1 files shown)


tests/test_django.py
```diff
@@ -16,3 +22,17 @@ def test_django_landing_page_loads():
 
     assert response.status_code == HTTPStatus.OK
     assert response.json()["title"] == "pygeoapi default instance"
+
+
+def test_django_edr_without_instance_id():
```
totycro, line 27: I've recently added the other django test above and since it was just one test, I didn't add any nice test setup. However for it run, it was necessary to set DJANGO_SETTINGS_MODULE for `django.setup()`.
Here this is now only happening for the test above and not the new one, so how does this work when only the test below is run?

If we need the patching for all tests in this file, we could also do it in some kind of fixture, or maybe have a fixture which returns the django test client (with the correct `SERVER_NAME`) and also does the setup.

petergarnaes: Good point, I was not quite sure how this was supposed to work. I've made a version with fixtures, please take a look

petergarnaes:
The pipeline dies during requirements installation, I can not re-trigger the build, but the error does not stem from my changes

tests/test_django.py
totycro: It looks reasonable for our use case. Btw there is a django-pytest module, but i think that's too much for a handful tests, so i'm in favor of this implementation.

totycro (approved):

tomkralidis:
@petergarnaes sorry for missing this.  Given #1549 has been pushed/merged, are you able to rebase?  The added test is of value here as well.  Thanks.

petergarnaes:
> @petergarnaes sorry for missing this. Given #1549 has been pushed/merged, are you able to rebase? The added test is of value here as well. Thanks.

@tomkralidis No worries, it's done

PR #1533, opened 2024-02-02
run api in another thread when using starlette

### LINKED ISSUE #1527: Starlette get request data does not work after dropping dependency on nest-asyncio
**Description**
After #1493 the request body is retrieved via `asyncio.run_coroutine_threadsafe()`. There are two issues with this:

1. This method returns a [concurrent.futures.Future](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future) and to get the actual data one needs to call `future.result()`
2. Upon calling the `future.result()` the loop is blocked and nothing will ever run in it. As per https://docs.python.org/3/library/asyncio-dev.html#asyncio-multithreading `asyncio.run_coroutine_threadsafe().result()` must be called from a thread different from the one where the loop is running.

**Steps to Reproduce**
Reading the documentation at https://docs.python.org/3/library/asyncio-dev.html#asyncio-multithreading and inspecting the `APIRequest` class. The Starlette integration tests are not catching this since `request.data` is not accessed in the tested procedures.

**Expected behavior**
Starlette integration should work.

**Additional context**
To solve this the best approach would be to have the option of an fully async api. Providers that are not async can still run via `loop.run_in_executor`.

The quick and dirty approach would be to use `loop.run_in_executor` when calling the api in the `starlette_app`. This will open a new thread for each call. We can then run `asyncio.run_coroutine_threadsafe().result()` when getting the request body. As a bonus we can use async providers (like the async MongoDB driver). A issue with this solution might be that the API class is potentially not threadsafe.

ricardogsilva:
# Overview
This PR modifies the starlette execution code in order to run calls to pygeoapi `API` in a new thread, thus avoiding blocking the main event loop.

This fix is taken mostly verbatim from the changes committed by @constantinneagu's in 

[Geries-Ingenieure-GmbH/pygeoapi](https://github.com/geopython/pygeoapi/compare/master...Geries-Ingenieure-GmbH:pygeoapi:master)

# Related issue / discussion
- Fixes #1527 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -370,7 +370,7 @@ class APIRequest:
                 # has been implemented, with_data() can become async too
                 loop = asyncio.get_event_loop()
                 api_req._data = asyncio.run_coroutine_threadsafe(
-                    request.body(), loop)
+                    request.body(), loop).result(1)
         return api_req
 
     @staticmethod
--- a/pygeoapi/starlette_app.py
+++ b/pygeoapi/starlette_app.py
@@ -32,8 +32,9 @@
 # =================================================================
 """ Starlette module providing the route paths to the api"""
 
+import asyncio
 import os
-from typing import Union
+from typing import Callable, Union
 from pathlib import Path
 
 import click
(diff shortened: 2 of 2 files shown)


ricardogsilva:
@constantinneagu, can you please review?

francbartoli (approved):

constantinneagu (approved):
Looks good.

constantinneagu (approved):

PR #1534, opened 2024-02-06
Mock server in WFS provider tests

totycro:
# Overview

This commit mocks the layer in WFS tests and prepares return values required to satisfy the tests. This reduces the test fidelity, but it makes them stable and increases execution speed (literally 100x on my machine, 66 sec to 0.6 sec).

Since the return value is mocked, it doesn't make sense to send different parameters to the server, so some tests became obsolete.

The original tests are still there in another file, but are not supposed to be run in CI.

# Related issue / discussion

#1517 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


totycro:
Btw this PR would fix issues that surfaced here: https://github.com/geopython/pygeoapi/pull/1567

PR #1549, opened 2024-02-27
fix Django EDR implementation on non-instance endpoints (#1528)

Haj-Res:
# Overview
The issue was created by using the same view for a instance and non-instance endpoint, while having the instance_id argument be mandatory. By making the argument optional, the view works with and without the instance argument.

# Related issue / discussion

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/django_/views.py
+++ b/pygeoapi/django_/views.py
@@ -348,7 +348,7 @@ def collection_tiles_metadata(request: HttpRequest, collection_id: str,
 
 def collection_item_tiles(request: HttpRequest, collection_id: str,
                           tileMatrixSetId: str, tileMatrix: str,
-                          tileRow: str, tileCol: str, ) -> HttpResponse:
+                          tileRow: str, tileCol: str) -> HttpResponse:
     """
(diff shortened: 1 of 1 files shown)


pygeoapi/django_/views.py
```diff
@@ -347,7 +348,7 @@ def collection_tiles_metadata(request: HttpRequest, collection_id: str,
 
 def collection_item_tiles(request: HttpRequest, collection_id: str,
                           tileMatrixSetId: str, tileMatrix: str,
-                          tileRow: str, tileCol: str,) -> HttpResponse:
+                          tileRow: str, tileCol: str, ) -> HttpResponse:
```
tomkralidis, line 351: Can we update to:

```python
tileRow: str, tileCol: str) -> HttpResponse:
```

tomkralidis (changes requested):

pygeoapi/django_/views.py
Haj-Res: Removed the trailing comma

Issue #1551, opened 2024-02-27
Restore PostGIS provider for Tiles


PascalLike:
**Description**
Before the refactor on tiles providers, it was possible to use pg_tileserv and Martin as providers for tiles.

**Steps to Reproduce**
Add a pg_tileserver as tiles provider, for example:

```
   providers:
       - type: tile
         name: MVT 
         data: http://0.0.0.0:7800/public.ne_50m_admin_0_countries/{z}/{x}/{y}.mvt
            options:
              zoom:
                min: 0
                max: 15
             schemes:
                 - WebMercatorQuad
         format:
             name: pbf 
             mimetype: application/vnd.mapbox-vector-tile

``` 

**Expected behavior**
Is no longer possible to define MVT providers, so I would expect the possibility to define something like:

```
   providers:
       - type: tile
         name: MVT-postgis 
         data: http://0.0.0.0:7800/public.ne_50m_admin_0_countries/{z}/{x}/{y}.mvt
            options:
              zoom:
                min: 0
                max: 15
             schemes:
                 - WebMercatorQuad
         format:
             name: pbf 
             mimetype: application/vnd.mapbox-vector-tile

```
And have tiles available in pygeoapi without errors.

**Environment**
 - OS: Ubuntu
 - Python version: 3.10.12
 - pygeoapi version: latest

doublebyte1:
Well spotted, but I think it is a regression behaviour, more than a bug (:

PascalLike:
Fixed here https://github.com/geopython/pygeoapi/pull/1552

PR #1552, opened 2024-02-27
PostGIS MVT provider and docs

PascalLike:
# Overview
This restores the possibility of using PostGIS+pg_tileserve as a tiles provider. This same provider works also for PostGIS+Martin without big changes, so there is no need to create a separate one. The provider is named after PostGIS, since it's the source for data.

# Related issue / discussion
https://github.com/geopython/pygeoapi/issues/1551

# Additional information
I set up a docker compose to run the tests on both pg_tileserver and Martin, if it helps the testing I can provide it.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/data-publishing/ogcapi-tiles.rst
+++ b/docs/source/data-publishing/ogcapi-tiles.rst
@@ -20,6 +20,7 @@ pygeoapi core tile providers are listed below, along with supported features.
 
    `MVT-tippecanoe`_,❌,✅
    `MVT-elastic`_,✅,❌
+   `MVT-postgis`_,✅,✅
 
 Below are specific connection examples based on supported providers.
 
@@ -50,7 +51,7 @@ This code block shows how to configure pygeoapi to read Mapbox vector tiles gene
                  min: 0
                  max: 5
              schemes:
-                 - WorldCRS84Quad
+                 - WebMercatorQuad
          format:
              name: pbf 
              mimetype: application/vnd.mapbox-vector-tile
@@ -80,7 +81,7 @@ This code block shows how to configure pygeoapi to read Mapbox vector tiles from
                  min: 0
                  max: 5
              schemes:
-                 - WorldCRS84Quad
+                 - WebMercatorQuad
          format:
              name: pbf 
              mimetype: application/vnd.mapbox-vector-tile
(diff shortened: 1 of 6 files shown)


pygeoapi/plugin.py
```diff
@@ -52,6 +52,7 @@
         'MongoDB': 'pygeoapi.provider.mongo.MongoProvider',
         'MVT-tippecanoe': 'pygeoapi.provider.mvt_tippecanoe.MVTTippecanoeProvider',  # noqa: E501
         'MVT-elastic': 'pygeoapi.provider.mvt_elastic.MVTElasticProvider',  # noqa: E501
+        'MVT-postgis': 'pygeoapi.provider.mvt_postgis.MVTPostGISProvider',  # noqa: E501
```
francbartoli, line 55: Can we change this name to a more generic `'MVT-proxy': 'pygeoapi.provider.mvt_proxy.MVTProxyProvider'` and ensure that in the docs the supported targets like `martin` and `pg_tileserv` are listed?

francbartoli (changes requested):

pygeoapi/plugin.py
PascalLike: Thanks Francesco, I updated the code and the documentation following your suggestion.

francbartoli (approved):

docs/source/data-publishing/ogcapi-tiles.rst
```diff
@@ -50,7 +51,7 @@ This code block shows how to configure pygeoapi to read Mapbox vector tiles gene
                  min: 0
                  max: 5
              schemes:
-                 - WorldCRS84Quad
```
doublebyte1, line 53: @PascalLike you can leave it like this, for now, but this parameter will disappear with my PR.

tests/pygeoapi-test-config-apirules.yml
```diff
@@ -212,7 +212,7 @@ resources:
               data: tests/data/ne_110m_lakes.geojson
               id_field: id
             - type: tile
-              name: MVT
+              name: MVT-tippecanoe
```
doublebyte1, line 215: well spotted!

doublebyte1:
@PascalLike great work! Please add an example on [pygeoapi-examples](https://github.com/geopython/pygeoapi-examples), when you have the chance 👍🏽

doublebyte1 (approved):
Great job! Please add an example on [pygeoapi examples](https://github.com/geopython/pygeoapi-examples), when you have the chance

Issue #1439, opened 2023-12-18
expose shapely functions as sample process


tomkralidis:
As suggested by Sumit Sen at GISE OGC Stack Winter School 2023, given [Shapely](https://shapely.readthedocs.io) is a core pygeoapi dependency, it would be value to have `pygeoapi.process.shapely_functions.py` to expose various geometric processing functionality.  Candidates could be:

- measurement (length, area, etc.)
- predicates
- set operations
- constructive ops

(there could be others).

All functions should be able to accept either WKT or a GeoJSON geometry object and output same.

krishnaglodha:
Can I work on it ?

tomkralidis:
Sure!

tomkralidis:
Update: it looks like we have 2 open PRs in support of this issue:
- #1481 
- #1565 

Can you let me know if you are actively working on these PRs?  We should also consider coordinating this work further into a single PR.

jeafreezy:
Hi @tomkralidis , yes I am working on mine. I will be pushing the final docs this weekend.

jeafreezy:
I've pushed the docs. The PR is npw open for review @tomkralidis.🙏🏿

tomkralidis:
Initial implementation added in #1565.  Thanks!

PR #1565, opened 2024-02-28
Shapely functions as sample process

jeafreezy:
# Overview

This PR exposes some selected shapely functions as sample process. The selection cut across different operations in shapely. It uses the name of the category of that function as the namespace. E.g union operation under the set module is described as `set:union`. This is to avoid clashes with other functions.

As mentioned in the issue, it accepts list of geometry inputs (WKT and/or GeoJSON geometry), performs the operation and returns the result in the specified **output_format** (optional).

The approach is describe in the image below:

image

# Related issue / discussion

https://github.com/geopython/pygeoapi/issues/1439

# Additional information

**Todo/Idea/Challenge**

- Tests(units) - I didn't see any test for process in the codebase. I tried writing one, but I need to spend some more time to understand the test structure.

- I am yet to figure out how to pass optional function parameters in the inputs. I am still confused on the data type to use in the metadata, especially since different functions require different parameters. e.g **buffer** operation requires **distance**. @tomkralidis and I discussed an alternative approach earlier today, which is using different classes for different functions. That could potentially solve this issue. But I didn't have much time to explore that. Although this may also lead to repetitive code. Since the idea is to have it as a sample, hopefully this current implementation is sufficient.

- Idea: Perhaps in the future, it will be great to chain functions. E.g passing multiple **operations** like ['buffer', 'join'] can be interpreted as perfoming the **buffer** operation on the geometries, followed by the **join** operation before returning the result.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


francbartoli:
@jeafreezy thanks for your great contributions. I have just one general comment about security:
- Exposing these functions publicly would leave a big surface for DoS attacks so we should make sure that they are not enabled as default processes sample
- My suggestion is to pick up them from the configuration if additionally configured

tomkralidis:
Agree; it makes sense to have the plugin in the codebase without having in default configuration.

jeafreezy:
Thank you both for the review and feedback. Please, how do you think I should go about this? I have a few ideas;

- Should there be a new entry in the **example-config.yml** for functions, where the user can list the functions they want to enable? Something like below;

image

- Or the shapely-functions entry should be commented out in the config, so they user can enable it, if they want to ? Something like below;

image

- Or is there a better approach to implement the suggestion ?

Kindly advise. Thank you.

webb-ben:
Hi @jeafreezy, I believe the correct approach would be to omit the configuration entirely from the default pygeoapi configuration file, and instead, documenting the capacities of the Shapely Process Provider [here](https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-processes.html) / [location in source code](https://github.com/geopython/pygeoapi/blob/master/docs/source/data-publishing/ogcapi-processes.rst)

jeafreezy:
Thank you @webb-ben . I will work on that

.pre-commit-config.yaml
```diff
@@ -4,16 +4,15 @@
 files: '\.(py|sh|sql)$'
```
webb-ben, line 4: Are changes to this file necessary?

docs/source/data-publishing/ogcapi-processes.rst
```diff
@@ -9,17 +9,24 @@ fashion (inputs, outputs).
...
+    
+    # enabled by default
        hello-world:
            processor:
                name: HelloWorld
+        
+        #Add this block to the configuration file to enable shapely processes.
+        #shapely-functions:
+           #processor:
+               #name: ShapelyFunctions
```
webb-ben, line 29: Could you move this to the start of the Shapely section, with the process uncommented?
```md
Shapely Functions (Optional)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^

.. code-block:: yaml

        shapely-functions:
           processor:
               name: ShapelyFunctions
```

pygeoapi-config.yml
```diff
@@ -302,3 +302,4 @@ resources:
         type: process
```
webb-ben, line 302: Are changes to this file necessary?

pygeoapi/templates/tilematrixsets/tilematrixset.html
```diff
@@ -1,18 +1,19 @@
-{% extends "_base.html" %}
-{% block title %}{{ super() }} {% trans %}TileMatrixSet{% endtrans %} {% endblock %}
-{% block crumbs %}{{ super() }}
-/ <a href="../TileMatrixSets/">{% trans %}TileMatrixSets{% endtrans %}</a>
-/ <a href="../TileMatrixSets/{{ data['id'] }}?f=html">{{ data['id'] }}</a>
-{% endblock %}
-{% block body %}
+{% extends "_base.html" %} {% block title %}{{ super() }} {% trans
```
webb-ben, line 1: Are changes to this file necessary?

webb-ben (changes requested):
Not sure if we need need pytest to ensure function of the process capacities. 

Apart from that, small changes to docs and a couple of files I am not sure need to be included in this PR, got my +1.

jeafreezy:
> Not sure if we need need pytest to ensure function of the process capacities.
> 
> Apart from that, small changes to docs and a couple of files I am not sure need to be included in this PR, got my +1.

Thank you for reviewing my PR. I will address the issues.

.pre-commit-config.yaml
jeafreezy: image

I believe this happened when I synced with the main branch recently. I was not my edit.

docs/source/data-publishing/ogcapi-processes.rst
jeafreezy: Alright. Will do.

pygeoapi-config.yml
jeafreezy: It's as a result of syncing with the master branch.

pygeoapi/templates/tilematrixsets/tilematrixset.html
jeafreezy: I didn't changed the files. But I think it was just reformatted by my editor.

.gitignore
tomkralidis, line 1: Can we remove this addition?

.pre-commit-config.yaml
tomkralidis, line 1: Can we remove this addition?

docs/source/data-publishing/ogcapi-processes.rst
```diff
@@ -9,14 +9,16 @@ fashion (inputs, outputs).
 pygeoapi implements OGC API - Processes functionality by providing a plugin architecture, thereby
 allowing developers to implement custom processing workflows in Python.
 
-A `sample`_ ``hello-world`` process is provided with the pygeoapi default configuration.
+The pygeoapi offers two processes: a default ``hello-world`` process which allows you to quickly explore the capabilities of processes, and an optional ``shapely-functions`` process with more advanced features that leverages Shapely_ to expose various geometric processing functionality.
```
tomkralidis, line 12: `Shapely_` -> `` `Shapely`_``

docs/source/data-publishing/ogcapi-processes.rst
```diff
@@ -107,14 +112,87 @@ Processing examples
...
    # execute a job for the ``hello-world`` process with a success subscriber
+    curl -X POST http://localhost:5000/processes/hello-world/execution \
+        -H "Content-Type: application/json" \
+        -d "{\"inputs\":{\"name\": \"hi there2\"}, \
+            \"subscriber\": {\"successUri\": \"https://www.example.com/success\"}}"
+
+Shapely Functions (Optional)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The `shapely-functions` process exposes some selected Shapely_ functions as sample process. The selection cut across different operations in shapely. To avoid function collision, it uses the name of the function category as the namespace. E.g *union* operation under the *set* module is described as *set:union*.
```
tomkralidis, line 124: - `Shapely_` -> `` `Shapely`_``
- shapely -> Shapely

pygeoapi-config.yml
tomkralidis: Can we remove this addition?

pygeoapi/process/shapely_functions.py
```diff
@@ -0,0 +1,409 @@
...
+# OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+# HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+# WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# =================================================================
+
+import enum
```
tomkralidis, line 32: Imports should be spaced and ordered as follows:

```python
import enum
import logging

from typing import Any, Dict, List, Tuple, Union

from shapely.geometry import mapping, shape
from shapely.geometry.base import BaseGeometry
from shapely.prepared import prep
from shapely.wkt import loads

from pygeoapi.process.base import BaseProcessor, ProcessorExecuteError
```

pygeoapi/process/shapely_functions.py
```diff
@@ -0,0 +1,409 @@
...
+        "predicates",
+        "set operations",
+        "constructive ops",
+    ],
+    "links": [
+        {
+            "type": "text/html",
+            "rel": "about",
+            "title": "information",
+            "href": "https://example.org/process",
```
tomkralidis, line 125: Update link to Shapely, i.e. `https://shapely.readthedocs.io`

pygeoapi/templates/tilematrixsets/tilematrixset.html
```diff
@@ -15,4 +15,4 @@ <h1>{{ data['id'] }} tile matrix set</h1>
   </ul>
 
 </section>
-{% endblock %}
+{% endblock %}
```
tomkralidis, line 18: Can we remove this addition?

pygeoapi/process/shapely_functions.py
tomkralidis, line 1: Overall comment: any Python objects with trailing commas (dicts, lists, etc.) should have the trailing commas removed; example:

from

```python
dict_ = {
    'foo': 'bar',
}
```

to


```python
dict_ = {
    'foo': 'bar'
}
```

tomkralidis (changes requested):

jeafreezy:
Thank you for reviewing my PR @tomkralidis. I have addressed the issues to the best of my knowledge. Kindly let me know if there is anything else I missed. Thank you.

pygeoapi/process/shapely_functions.py
jeafreezy: Thank you for the feedback @tomkralidis

tomkralidis (approved):

PR #1571, opened 2024-03-02
Updated emoji to something more readable

doublebyte1:
# Overview
The question mark symbol on the [documentation of tiles](https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-tiles.html#publishing-tiles-to-ogc-api-tiles) is not very readable, due to the table background. 

Screenshot from 2024-03-02 11-09-54

This PR updates it to a more readable colour.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/data-publishing/ogcapi-tiles.rst
+++ b/docs/source/data-publishing/ogcapi-tiles.rst
@@ -20,7 +20,7 @@ pygeoapi core tile providers are listed below, along with supported features.
 
    `MVT-tippecanoe`_,❌,✅,✅,❌
    `MVT-elastic`_,✅,❌,✅,❌
-   `MVT-proxy`_,❔,❔,❔,❔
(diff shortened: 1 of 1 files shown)


tomkralidis:
Thanks @doublebyte1.  What does the question mark mean in this regard?  I'm wondering whether we should apply (in another PR) the same logic to the other support matrixes in the docs.

doublebyte1:
@tomkralidis the idea was to express uncertainty. As the MVT-proxy provider is very generic, in some cases it may support these properties, while in others it may not. Users will have to refer to the documentation of the specific provider being proxied. I am open to use another symbol, instead of the question mark, but it would be nice to have something with better contrast with the background (:

tomkralidis (approved):

PR #1575, opened 2024-03-04
fix links for vector tiles

simonseyock:
# Overview

This fixes the links for vector tiles that contained an error introduced by https://github.com/geopython/pygeoapi/pull/1473

# Related issue / discussion

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -1045,9 +1045,7 @@ class API:
             if 'format' in collection_data:
                 collection_data_format = collection_data['format']
 
-            is_vector_tile = (collection_data_type == 'tile' and
-                              collection_data_format['name'] not
-                              in [F_PNG, F_JPEG])
+            is_raster_tile = collection_data['name'] == 'WMTSFacade'
 
             collection = {
                 'id': k,
@@ -1161,7 +1159,8 @@ class API:
                     'href': f'{self.get_collections_url()}/{k}/schema?f={F_HTML}'  # noqa
                 })
 
-            if is_vector_tile or collection_data_type in ['feature', 'record']:
+            if (not is_raster_tile
+                    and collection_data_type in ['feature', 'record', 'tile']):
                 # TODO: translate
                 collection['itemType'] = collection_data_type
                 LOGGER.debug('Adding feature/record based links')


pygeoapi/api.py
```diff
@@ -1045,9 +1045,7 @@ def describe_collections(self, request: Union[APIRequest, Any],
             if 'format' in collection_data:
                 collection_data_format = collection_data['format']
 
-            is_vector_tile = (collection_data_type == 'tile' and
-                              collection_data_format['name'] not
-                              in [F_PNG, F_JPEG])
+            is_raster_tile = collection_data['name'] == 'WMTSFacade'
```
simonseyock, line 1048: I do not like the check this way, but at this point in the function we do not have the actual provider object, but only the configuration.

I see 2 ways to resolve this:
* We can include the tile type into the configuration. To be consistent this should be done for vector tiles as well.
* We can move the link generation inside the function to appear after the instantiation of the providers.

simonseyock: A third way is to test for the format, but the check might need to be updated if at any time another format is introduced.

doublebyte1, line 1048: thanks for bringing this up, @simonseyock . Relying on the `collection_data['name'] key` (or the `collection_data_type`, as you were doing before) is problematic when using multiple providers, as it will only take the values of the first provider. So, for instance, if you have a collection that has features, vector tiles, and raster tiles (by this order), it will assume the value for features and it will fail the condition on line 1048. You can test it, using this configuration:
```

        providers:
            - type: feature
              name: GeoJSON
              id_field: id
            - type: tile
              name: MVT-tippecanoe
              data: tests/data/tiles/ne_110m_lakes
              options:
                bounds: [[-124.953634,-16.536406],[109.929807,66.969298]]
                zoom:
                    min: 0
                    max: 5
              format:
                    name: pbf
                    mimetype: application/vnd.mapbox-vector-tile
            - type: tile
              name: WMTSFacade
              data: https://emotional.byteroad.net/geoserver/gwc/service/wmts
              format:
                  name: png  # png or jpeg
                  mimetype: image/png
              options:
                  wmts_layer: camb:hex350_grid_mental_1920 # the layer name of the wmts
                  wmts_tile_matrix_set: WebMercatorQuad  # the name of the tile matrix set of the wmts.
                  scheme: WebMercatorQuad  # the aligning scheme in pygeoapi.
                  zoom:
                      min: 0
                      max: 20

```

doublebyte1: > I do not like the check this way, but at this point in the function we do not have the actual provider object, but only the configuration.
> 
> I see 2 ways to resolve this:
> 
> * We can include the tile type into the configuration. To be consistent this should be done for vector tiles as well.
> * We can move the link generation inside the function to appear after the instantiation of the providers.

Including the tile type into the configuration would face the same problem that I described above. However, I think it is worth to explore your second option.

doublebyte1:
@simonseyock I created a separated issue to discuss this problem: https://github.com/geopython/pygeoapi/issues/1583

Issue #1573, opened 2024-03-02
downgrade to Pydantic 1


tomkralidis:
**Description**

As discussed [this week](https://github.com/geopython/pygeoapi/wiki/Code-Sprint-2024-02-29-to-2024-03-01#notes), it is decided to downgrade Pydantic support to v1.  While we are embarking on a broader effort to make pydantic optional, we still need to downgrade to Pydantic to be able to cut a release (last pygeoapi release was May 2023), as as priority 1 exercise.

cc @ricardogsilva

tomkralidis:
Pydantic v1/2 compatability added in #2014

PR #1584, opened 2024-03-09
Revert pydantic models to v1 version

francbartoli:
# Overview

This PR addresses the issue with pydantic v2 which doesn't allow to cut a release as per our dependency policy (RFC2).

Please be aware that this is a **breaking change** for all downstream projects which are using `pygeoapi` as a dependent library together with [pydantic-settings](https://github.com/pydantic/pydantic-settings) like [fastgeoapi](https://github.com/geobeyond/fastgeoapi). `pydantic-settings` is only compatible with `pydantic>2.0`.
The plan is to move back to pydantic v2 as soon as it will be released as debian package and then remove it as a dependency in favor of dataclasses and the approach with duck typing pattern as per [meeting notes](https://github.com/geopython/pygeoapi/wiki/Code-Sprint-2024-02-29-to-2024-03-01#notes). [Here](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1052028) the progress of the request.

# Related Issue / discussion

#1573 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/config.py
+++ b/pygeoapi/config.py
@@ -4,7 +4,7 @@
 #          Francesco Bartoli <xbartolone@gmail.com>
 #
 # Copyright (c) 2022 Tom Kralidis
-# Copyright (c) 2023 Francesco Bartoli
+# Copyright (c) 2024 Francesco Bartoli
 #
 # Permission is hereby granted, free of charge, to any person
 # obtaining a copy of this software and associated documentation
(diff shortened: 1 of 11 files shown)


francbartoli:
I would wait to merge @kalxas, I'd like to add more tests for the models so that it would be easier to do the next steps

kalxas:
ok, I will wait

tomkralidis (approved):

PR #1585, opened 2024-03-09
OpenAPI: skip failing collections (#1577)

### LINKED ISSUE #1577: Graceful failure for xarray provider
I'm primarily working with the xarray provider. When I work with a pygeoapi config file contains many resources, but just a subset of them fail to properly generate the openapi, the entire process fails.

 I think it would be useful to have the xarray provider (and possibly other providers as well) fail gracefully and issue/log a warning for a resource and then continue to the next one. This would make it easier to identify the data source(s) that need to be corrected, but also make it so that a single invalid dataset doesn't prevent the entire build process.

tomkralidis:
# Overview
Skips failing collections on OpenAPI document generation

# Related Issue / discussion

Fixes #1577 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/administration.rst
+++ b/docs/source/administration.rst
@@ -49,6 +49,10 @@ To generate the OpenAPI document as JSON, run:
    property names and their data types.  Whenever you make changes to your pygeoapi configuration,
    always refresh the accompanying OpenAPI document.
 
+.. note::
+   By default, pygeoapi OpenAPI document generation will fail on any collection whose provider is improperly
+   configured or when encountering a connection issue.  To skip failing collections, use the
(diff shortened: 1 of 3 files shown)


pvgenuchten:
Tom, how is the administrator made aware of a failing collection, if a failing collection is skipped from the index, it may be hard to identify a failing collection

tomkralidis:
Good idea as well @pvgenuchten. Added to PR (default behaviour is to fail).

webb-ben (approved):
+1

kalxas (approved):

PR #1599, opened 2024-03-18
Add support for parameter Outputs of OGC API Processes

francescoingv:
# Overview
The parameter `outputs` is passed to the Manager, which in turn pass it to the Processor,
subject to the Processor _being declared to support the parameter_.

The conditional parameter passing to the Processor _grant backward compatibility for already developed Processors_.

# Related Issue / discussion
Solve https://github.com/geopython/pygeoapi/issues/1420,
Subset https://github.com/geopython/pygeoapi/issues/1285

# Additional information
This PR has a breaking signature in `BaseManager.execute_process()`
If needed, an additional PR could be made to have backward compatibility:
adding a `supports_output` property to the Manager and handling it as required (currently not in place as I think it may set a maintenance issue).

The backward compatibility of the Processor is given by a `supports_outputs` property of the BaseProcessor,
by default set to `True`.
To check the compatibility, the example processor `HelloWorld` has `supports_outputs=True`,
while `EchoProcessor` follow the old signature.

I think an **error 501** should be raised on receiving the parameter outputs, and not having the Processor handling it.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api.py
+++ b/pygeoapi/api.py
@@ -3497,7 +3497,7 @@ class API:
 
         out_dict = data.get('outputs', None)
         LOGGER.debug(f'outputs: {out_dict}')
-        
+
         subscriber = None
         subscriber_dict = data.get('subscriber')
         if subscriber_dict:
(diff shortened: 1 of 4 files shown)


francescoingv:
Sorry, I made some errors in the pull request process.
Will close and reopen a new one.

PR #1626, opened 2024-04-12
use typing.Union for pre-3.10 compat

tomkralidis:
# Overview
Non-breaking change to use `typing.Union`, which work pre 3.10 installs.

# Related Issue / discussion

# Additional information
None
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


webb-ben (approved):

GeoSander (approved):
Looks good to me!

PR #1640, opened 2024-04-29
[WIP] File system/Data storage abstraction

MTachon:
# Overview

This PR intends to abstract away the local/remote file system or byte storage where the data used by the providers are stored. The implementation leverages a file-system interface provided through `fsspec`. `fsspec` provides already support for various file-systems and cloud storage services through [built-in implementations](https://filesystem-spec.readthedocs.io/en/latest/api.html#built-in-implementations). It also allows for using [other known implementations](https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations), as well as [implementing](https://filesystem-spec.readthedocs.io/en/latest/developer.html#implementing-a-backend) and [registering](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.registry.register_implementation) new backends.

The data storage abstraction happens mostly in the implementation of the `BaseProvider`, where a new instance attribute `fs` is introduced. This attribute, which is a `fsspec` file-system interface, is inherited by other providers. Other providers can use this file-system interface in their implementation to access the data.

The instantiation of `fsspec` file-system objects may use configuration variables set in the `providers` section (some backends  used by `fsspec` may also use environment variables).

This PR introduces a new optional `file_system` section, in the `providers` section of pygeoapi's runtime configuration. If the `file_system` section is omitted, the `fs` attribute of the `BaseProvider` will be an instance of the [`fsspec.implementations.local.LocalFileSystem`](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.local.LocalFileSystem) class. In that case, the `fs` attribute can be used by other providers to access and read files on the *"local"* file system. In the implementations of the providers, calls to the builtin `open` functions can then be replaced by the `open` method of the [`LocalFileSystem`](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.local.LocalFileSystem):

```python

# Calls of 'open' builtin function...
with open(self.data, mode='rt') as f:
    ...

# ... can be replaced with calls of 'open' method of LocalFileSystem instance
with self.fs.open(self.data, mode='rt') as f:
    ...
```

The `file_system` section, if given in pygeoapi's configuration, has one `protocol` mandatory field. The value passed to this field must be one of the protocols supported by `fsspec` (see [`fsspec.available_protocols()`](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.available_protocols)). For faster access, the data can be cached locally (e.g. when the data is on remote storage). This is of course not suitable for very large datasets, as the data needs to be downloaded on the first query, which is in that case both time consuming and takes much space. To cache locally the data, one can configure pygeoapi's runtime as follows:

``` yaml

providers
    - type: ...
      ...
      data: <my-bucket>/<key>
      file_system:  # optional
          protocol: gs  # mandatory, anything from the `fsspec.available_protocols()` list
          storage_options:  # optional
              # Credentials and other keywords parameters, specific to implementations supported by fsspec.
              # See https://filesystem-spec.readthedocs.io/en/latest/api.html#implementations and https://filesystem-spec.readthedocs.io/en/latest/api.html#external-implementations
              ...
          cache_storage: /path/to/cached/data  # optional, if not given, a temporary directory (cleaned up when process ends) will be used
          cache_options:  # optional
              # see https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.cached.WholeFileCacheFileSystem
              expiry_time: ...
              ...
```

# Related Issue / discussion

https://github.com/geopython/pygeoapi/issues/824

# Additional information

- This PR also adapted the implementation of several providers (GeoJSON, CSV, FileSystem, Rasterio, Xarray) to the data storage abstraction, which should give a good starting point for the implementation of other providers.
- As for now, the implementation of the data storage abstraction cannot be used by the `OGRProvider`, as the `Open` method of drivers from `osgeo.gdal`/`osgeo.ogr` accepts only a path (`str`) to the data to open, and not a *file-like* handle. However, this provider can already access data on remote storage with [GDAL's virtual file systems](https://gdal.org/user/virtual_file_systems.html). The environment variables (if any needed) need then to be set in pygeoapi's parent process to make it work.
- `fsspec` was added to `requirements.txt` as it is used in `pygeoapi/provider/base.py`. `fsspec` has had a OS package for Ubuntu for some time now, and does not seem to have known high or critical vulnerabilities. 
- `fsspec` has support for [builtin implementations](https://filesystem-spec.readthedocs.io/en/latest/api.html#built-in-implementations). Packages for [external implementations](https://filesystem-spec.readthedocs.io/en/latest/api.html#external-implementations) supported by `fsspec` can be added to the `requirements-provider.txt` file.
- `netCDF4` was replaced by `h5py` and `h5netcdf` to allow reading from remote storage with `xarray`.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

# Updates to public demo

# Contributions and licensing


tomkralidis:
@MTachon I see this PR is marked as WIP.  Is this still the case or is it ready for review?  Thanks.

tomkralidis:
Closing this PR for now.  Please feel free to re-open once it is ready for review.

Issue #1578, opened 2024-03-04
Modifications to CRS handling in xarray_.py provider


sjordan29:
**Description**
[These lines of code](https://github.com/LimnoTech/pygeoapi/blob/master/pygeoapi/provider/xarray_.py#L492-L509) in the xarray provider are causing issues for some of the EDR datasets I'm trying to serve. 

I'm working with [CF-compliant zarrs/netCDFs](https://cfconventions.org/cf-conventions/cf-conventions.html#appendix-grid-mappings), and I see a few issues as the current code base and its ability to work with a broad range of datasets. 
1. `epsg_code` is not a required attribute for CF compliance, so this section of the code can result in a failure to build if you're working with a dataset with a CRS variable that does not have this attribute.
2. The link we're filling in with EPSG codes ( [http://www.opengis.net/def/crs/OGC/1.3/](http://www.opengis.net/def/crs/OGC/1.3/)) doesn't actually reference any EPSG codes. Should this instead be referring to [https://www.opengis.net/def/crs/EPSG/0/](https://www.opengis.net/def/crs/EPSG/0/)?
3. Some datasets we would like to serve with pygeoapi are WGS84 and they still have a `crs` variable. These lines of code would change the `crs_type` attribute to 'ProjectedCRS', even though it should remain as GeographicCRS.
4. CF conventions also don't require any particular name for the data variable that holds the CRS information. I'd recommend (potentially after checking to see if there's a variable named `crs`), looking to see if any of the spatio-temporal variables have a `grid_mapping` attribute, and use the information instead. This would allow us to dynamically search for the variable that contains coordinate reference system information, in case it is not named `crs`.

**Screenshots/Tracebacks**
Here's the error I got:
```
2024-03-01 13:05:15   File "/mambaforge/lib/python3.11/site-packages/pygeoapi/provider/xarray_.py", line 94, in __init__
2024-03-01 13:05:15     raise ProviderConnectionError(err)
2024-03-01 13:05:15 pygeoapi.provider.base.ProviderConnectionError: 'DataArray' object has no attribute 'epsg_code'
2024-03-01 13:05:15 ERROR: openapi.yml could not be generated ERROR
```

sjordan29:
I've started drafting a solution to this [on our fork](https://github.com/LimnoTech/pygeoapi/blob/crs_handling/pygeoapi/provider/xarray_.py#L401-L425) within `_get_coverage_properties` with the following

```python
[...]

Issue #1104, opened 2023-01-15
CRS query parameter support for EDR


petergarnaes:
Section 8.2.1.4 of the [EDR specification v. 1.0.1](https://docs.ogc.org/is/19-086r5/19-086r5.html) (which is the newest version of writing this) states that a client can specify a `crs` query parameter.

Currently this parameter is not passed to the provider as [this code shows](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/api.py#L3612-L3623), so there is no way to support other CRS than WGS84, because it is the default when not specified

In fact pygeoapi seems to break requirement A.2.12 part C which states that HTTP status code 400 should be returned if CRS is unsupported. If CRS is not passed to the provider, there is no way it can fail if CRS is unsupported. In my testing the `crs` query parameter is completely ignored by pygeoapi for EDR.

**Describe the solution you'd like**
`crs` should be passed to the provider as a query argument and should default to either WGS84 or WGS84h depending on if the Z-axis parameter is specified in either in the `bbox` query argument or `z` query parameter.

The config should also allow specifying supported CRS for the collection, so requirement A.57 can be adhered to.

So config would look something like:
```yml
resources:
  my_collection:
    type: collection
    title: My Collection
    ...
    crs:
     - http://www.opengis.net/def/crs/OGC/1.3/CRS84
     - native
```

There should also be an exception that can be thrown from a provider that results in an HTTP status code 400. Don't know if this is already supported.

sjordan29:
I came across this issue while addressing #1578. This will be important to support datasets in included in your STAC, @dblodgett-usgs -- there are a number of datasets whose primary coordinates are in a projected coordinate reference system rather than WGS84. @aufdenkampe

@tomkralidis, what does the future of adding support for a crs query parameter in EDR (and subsequent CRS transformations) look like? It looks like there is support for transforming features (via the crs_transform decorator) , as well as in the `rasterio_.py` coverage provider (using `pyproj`'s `Transformer`.

PR #1641, opened 2024-04-29
CRS handling in xarray provider properties

sjordan29:
# Overview
As discussed in #1578, I was experiencing a build failure for several EDR datasets that had a `crs` variable that did not have very specific attributes. 

This solution adds a `storage_crs` variable/attribute to the `XarrayProvider` -- I named it that based on CRS handling in the `features` provider, but am open to other naming schemes if you think something else is more appropriate. I had considered using the `crs` under `extents` in the config, but could see an instance where the extents are in WGS84 while the data itself are projected to some other coordinate system. 

`storage_crs` is defined with the added `_parse_storage_crs` method, which does the following:
1. Looks to the config file for a `storage_crs`. As noted in the documentation change, this value in the config could be anything accepted by [`pyproj.CRS.from_user_input`](https://pyproj4.github.io/pyproj/stable/api/crs/crs.html#pyproj.crs.CRS.from_user_input). If no `storage_crs` is found in the config, we attempt to parse the crs information directly from the zarr/netCDF file with the following steps.
2. Start by checking to see if the xarray dataset is CF-compliant ([details on grid mapping here](https://cfconventions.org/cf-conventions/cf-conventions.html#grid-mappings-and-projections)). This code looks to see if there is a grid mapping attribute for spatiotemporal variables. If there is, then we will parse crs information from the variable with that name in the xarray using [`pyproj.CRS.from_cf`](https://pyproj4.github.io/pyproj/stable/api/crs/crs.html#pyproj.crs.CRS.from_cf).
3. If a grid mapping does not exist, but a variable named `crs` does, then we will attempt to parse the crs information from the attributes of that variable using [`pyproj.CRS.from_dict`](https://pyproj4.github.io/pyproj/stable/api/crs/crs.html#pyproj.crs.CRS.from_dict).
4. If all else fails, assume a default storage CRS of http://www.opengis.net/def/crs/OGC/1.3/CRS84

In this fix, we just use `storage_crs` to define properties for the dataset, including `bbox_crs`, `inverse_flattening`, and `crs_type`. I'd envision in the future, the `storage_crs` could be used much like it is for features, or as `crs_src` is handled in the `rasterio_.py` to support a CRS query parameter.

I appreciate any feedback on any additional widely-used cases we should consider for CRS handling in the xarray provider and/or chagnes to make this more helpful for addressing future EDR development (e.g., #1104).

# Related Issue / discussion
Addresses #1578 
Will help pave a way forward for #1104 

# Dependency policy (RFC2)

# Updates to public demo

The netcdf file in the config above does not have a CRS variable or any grid mapping information, so the changes would assume WGS84 lat/lon.

# Contributions and licensing

- [x] I'd like to contribute this feature to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/pygeoapi/provider/xarray_.py
+++ b/pygeoapi/provider/xarray_.py
@@ -557,59 +557,6 @@ class XarrayProvider(BaseProvider):
 
         return crs
 
-    def _parse_grid_mapping(self):
-        spatiotemporal_dims = (self.time_field, self.y_field, self.x_field)
-        grid_mapping_name = None
-        for var_name, var in self._data.variables.items():
-            if all(dim in var.dims for dim in spatiotemporal_dims):
-                try:
-                    grid_mapping_name = self._data[var_name].attrs['grid_mapping']
-                except KeyError:
-                    LOGGER.debug()
-        return grid_mapping_name
-    
-    def _parse_storage_crs(
-        self,
-        provider_def: dict
-    ) -> pyproj.CRS:
-        """
-        Parse the storage CRS from an xarray dataset.
-        :param provider_def: provider definition
-        :returns: `pyproj.CRS` instance parsed from dataset
-        """
-        
-        try:
-            storage_crs = provider_def['storage_crs']
-            crs_function = pyproj.CRS.from_user_input            
-        except KeyError:
-            LOGGER.debug('''
-                No storage crs provided in the provider configuration.
-                Attempting to parse the CRS.
-                '''
-            )
-
-        if storage_crs is None:
-            grid_mapping = self._parse_grid_mapping()
-            crs_function = pyproj.CRS.from_dict
-            if grid_mapping is not None:
-                storage_crs = self._data[grid_mapping].attrs
-            elif 'crs' in self._data.variables.keys():
-                storage_crs = self._data['crs'].attrs
-            else:
-                storage_crs = DEFAULT_STORAGE_CRS
-                crs_function = get_crs_from_uri
-                LOGGER.debug('Failed to parse dataset CRS. Assuming WGS84.')
-
-        LOGGER.debug(f'Parsing CRS {storage_crs} with {crs_function}')
-        try:
-            crs = crs_function(storage_crs)
-        except CRSError as e:
-            LOGGER.debug(f'Unable to parse projection with pyproj: {e}')
-            LOGGER.debug('Assuming default WGS84.')
-            crs = get_crs_from_uri(DEFAULT_STORAGE_CRS)
-
-        return crs
-
 
 def _to_datetime_string(datetime_obj):
     """
--- a/pygeoapi/provider/xarray_.py
+++ b/pygeoapi/provider/xarray_.py
@@ -409,10 +409,11 @@ class XarrayProvider(BaseProvider):
             if epsg_code == 4326:
                 pass
             elif epsg_code is None:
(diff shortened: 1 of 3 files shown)


webb-ben (approved):

pygeoapi/provider/xarray_.py
```diff
@@ -401,16 +411,22 @@ def _get_coverage_properties(self):
...
-                properties['bbox_crs'] = f'http://www.opengis.net/def/crs/OGC/1.3/{self._data.crs.epsg_code}'  # noqa
-
-                properties['inverse_flattening'] = self._data.crs.\
-                    inverse_flattening
-
+        # Update properties based on the xarray's CRS
+        self.storage_crs = self._parse_storage_crs(provider_def)
+        epsg_code = self.storage_crs.to_epsg()
+        LOGGER.debug(f'{epsg_code}')
+        if (epsg_code == 4326) | (self.storage_crs == 'OGC:CRS84'):
```
tomkralidis, line 418: `if epsg_code == 4326 or self.storage_crs == 'OGC:CRS84':`

pygeoapi/provider/xarray_.py
```diff
@@ -476,6 +492,71 @@ def get_time_coverage_duration(self):
...
+                crs_function = pyproj.CRS.from_dict
+            else:
+                storage_crs = DEFAULT_STORAGE_CRS
+                crs_function = get_crs_from_uri
+                LOGGER.debug('Failed to parse dataset CRS. Assuming WGS84.')
+
+        LOGGER.debug(f'Parsing CRS {storage_crs} with {crs_function}')
+        try:
+            crs = crs_function(storage_crs)
+        except CRSError as e:
```
tomkralidis, line 551: `except CRSError as err:` (we typically use `err`.  Sorry for the consistency nitpick).

pygeoapi/provider/xarray_.py
```diff
@@ -338,9 +347,10 @@ def gen_covjson(self, metadata, data, fields):
 
         return cj
 
-    def _get_coverage_properties(self):
+    def _get_coverage_properties(self, provider_def: dict):
```
tomkralidis, line 350: It looks like `provider_def` is now passed around in order to use `storage_crs` in `_parse_storage_crs`.

Can we set `self.storage_crs` instead in the init?

tomkralidis (changes requested):

pygeoapi/provider/xarray_.py
```diff
@@ -37,12 +37,19 @@
...
+from pyproj.exceptions import CRSError
+
+from pygeoapi.api import DEFAULT_STORAGE_CRS
 
 from pygeoapi.provider.base import (BaseProvider,
                                     ProviderConnectionError,
                                     ProviderNoDataError,
                                     ProviderQueryError)
-from pygeoapi.util import read_data
+from pygeoapi.util import (
```
tomkralidis, line 49: `from pygeoapi.util import get_crs_from_uri, read_data`

pygeoapi/provider/xarray_.py
```diff
@@ -37,12 +37,19 @@
...
+    read_data,
+    get_crs_from_uri
+)
```
webb-ben, line 52:
```suggestion
from pygeoapi.util import get_crs_from_uri, read_data
```

tomkralidis (approved):

PR #1643, opened 2024-05-03
Refactor postgresql provider to allow exposing cached engine and table models to other providers

ricardogsilva:
# Overview

I'm working on an OGC API EDR provider that is backed by a postgresql DB, which coincidentally is the same DB being used for exposing the same resource as an OGC API features collection.

Therefore I would like to reuse the same DB engine and table reflected models as the pygeoapi postgresql features provider. Unfortunately these are not currently exposed to code outside the postgresql feature provider. 
This PR thus presents a light refactor of the `pygeoapi.provider.postgresql.py` module, with the intention of letting other providers make use of the engine and table caching facilities.

With this proposed implementation, the private method `PostgreSQLProvider._get_engine_and_table_model()` is replaced by two public functions:

- `get_engine()`
- `get_table_model()`

Both of these functions use the standard lib's [functools.cache](https://docs.python.org/3/library/functools.html#functools.cache) decorator, ensuring that they are cached.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/postgresql.py
+++ b/pygeoapi/provider/postgresql.py
@@ -286,7 +286,7 @@ class PostgreSQLProvider(BaseProvider):
         self.db_port = parameters.get('port', 5432)
         self.db_name = parameters.get('dbname')
         # db_search_path gets converted to a tuple here in order to ensure it
-        # is hashable - which allows us to use functools.lru_cache() when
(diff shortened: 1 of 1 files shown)


tomkralidis:
cc @KoalaGeo for any initial comments.

Issue #1341, opened 2023-08-30
Pydantic version 2


MTachon:
This issue is related to issues https://github.com/geopython/pygeoapi/issues/1297 and https://github.com/geopython/pygeoapi/issues/1328, as well as PR https://github.com/geopython/pygeoapi/pull/1298.

I believe pydantic version 2 and above can still be used by changing the single occurence of the `regex` keyword argument to `pattern` in the `pydantic.Field` instantiation, in the `pygeoapi/models/config.py` file. See the [removed keywords arguments in pydantic V2](https://docs.pydantic.dev/2.0/usage/errors/#removed-kwargs) and the [change of regular expressions library](https://docs.pydantic.dev/2.0/migration/#patterns-regex-on-strings) used by pydantic starting from V2. No regular expression features unavailable with the new regex library used by pydantic >= 2 are used in the pattern from the `pygeoapi/models/config.py` file. The new way of constraining strings in `pydantic.Field` is by using the `pattern` keyword argument, as illustrated in [strings constraints](https://docs.pydantic.dev/latest/usage/fields/#string-constraints). The second link above shows a work around if someone comes up with some string constraining using complex regex features. It does not seem that much of pygeoapi's code base relies on pydantic, yet. I was thinking in using pydantic for schema generation/validation, as well as for data validation for transactions implementation. Not sure yet what pydantic features would be needed, some may be available only from version 2 and above. If pydantic will be used to a greater extent in pygeoapi, it may be a good idea to stick with version 2 and above so that we do not introduce more breaking changes (between version 1 and 2) that we may need to fix  at a later stage?

I am willing to go through the use of pydantic in pygeoapi's code base and *convert* pydantic1.x code to pydantic2.x code if there are any breaking changes, if that is of interest?

MTachon:
Closing this issue, migration to pydantic v2 was implemented in https://github.com/geopython/pygeoapi/pull/1353

PR #1663, opened 2024-05-17
Compatibility with Pydantic >= 2 through pydantic.v1

SarahG-579462:
# Overview

Recently, pygeoapi's pydantic was downgraded from 2.0-compatible to pydantic < 1. This removed compatibility with two of our libraries, [xscen](https://github.com/Ouranosinc/xscen) and [xclim](https://github.com/Ouranosinc/xclim), for climate data processing, due to [intake-esm's current requirement on pydantic > 2](https://github.com/intake/intake-esm/blob/60766304f86173d8cad970e8979693e281f80658/requirements.txt#L10)

Because we require these in the use of our pygeoapi process plugins, this change forced us to use either very outdated versions of xscen (and thus xarray, pandas, xclim, etc.), or a very outdated version of pygeoapi.

Thankfully, a simple fix is available by the use of pydantic 2's backwards-compatible pydantic.v1 module. 

It is not entirely clear to me why pydantic was downgraded, the best explanation I can see in the discussion was due to "versionitis". I was, however, able to run all tests with the latest version of pydantic installed in this fashion. The way this PR is written, if pydantic >= 2 is not available on the operating system in question, everything will continue as-is.

# Related Issue / discussion

Pydantic issues:
#1573 
#1341
#1584

# Additional information

# Dependency policy (RFC2)

(It follows RFC2 as far as I understand, as it allows backwards compatibility with pydantic < 2)

# Updates to public demo

No breaking changes shuold ensue as a result of this PR, since it is backwards compatible. I am not entirely sure how to test this demo server, help would be appreciated here.

# Contributions and licensing


.github/workflows/vulnerabilities.yml
```diff
@@ -21,6 +21,10 @@ jobs:
       run:
         working-directory: .
     steps:
+    - id: string
```
tomkralidis, line 24: What is the nature of this update?  Should this be in another PR?

requirements-django.txt
```diff
@@ -1,2 +1,2 @@
 Django
-pydantic<2.0
+pydantic
```
tomkralidis, line 2: Not sure how this landed in `requirements-django.txt`, but given that pydantic is in `requirements.txt` I would actually remove this requirement in `requirements-django.txt`.

requirements.txt
```diff
@@ -4,7 +4,7 @@ filelock
 Flask
 jinja2
 jsonschema
-pydantic<2.0
+pydantic
```
tomkralidis, line 7: Is there a timeline for pydantic 2 on Ubuntu 22.04?  We need to consider this for this PR (note that our next release is scheduled for next month).

tomkralidis (changes requested):

.github/workflows/vulnerabilities.yml
SarahG-579462: This is related to the comment above. because I was working on a fork which has a capital letter in the repo name, it was failing to build the docker image in this github action. I don't think it's a great idea for this particular case, because it relies on a third party github action for security vulnerability checks, so I'll yeet it. :slightly_smiling_face:

requirements.txt
SarahG-579462: Are you saying that without `pydantic<2.0` here, it will override the system pydantic version, if someone is building without a virtual env, and that is an issue?

tomkralidis:
To add, it would be valuable to have pygeoapi.models.pydantic that covers the try/except in one place and have the rest of code import from pygeoapi.models.pydantic

tomkralidis:
Closing given https://github.com/geopython/pygeoapi/issues/1664#issuecomment-2192417748

PR #1673, opened 2024-06-03
Extra params fixes #1667

### LINKED ISSUE #1667: Pass extra parameters to requests to allow for custom behaviour down the line
For some requests there might be a need custom behaviour. This can be implemented using extra parameters in the request which are evaluated later on in providers or plugins. If they're not passed to the request nothing happens. When for example &foo=bar is added to the request this should be available in a dictionary passed to the providers for additional functionality. 
  

Request can be in the form http://localhost:8080/collections/lakes/items?&foo=bar
and a dictionary containing 'foo':'bar' is avaialable in the providers. 

**Additional context**
@tomkralidis @totycro we talked about this at the vienna code sprint.

Moritz-Langer:
# Overview
This PR contains the necessary adjustments for passing extra_params to providers and allow for custom behaviour in plugins. 

# Related Issue / discussion
Fixes #1667 

# Additional information
In the itemtypes a new variable is introduced. It contains a dictionary with unknown request parameters, basically anything can be passed to the request without changing behavior except there are no errors facing the client. The variable `extra_params` is then passed to the providers and accessible there. 
Another oracle specific change is the provider definition is getting passed to the manipulation class for possible adjustments to queries and filtering using the geometry column. Because it could also be named diffenently the name must be accessible later on. 

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


webb-ben (approved):

tomkralidis:
Note: OGC API - Features supports additional query parameters that are not defined in OpenAPI via [`vendorSpecificParameters`](https://docs.ogc.org/is/17-069r4/17-069r4.html#query_parameters), which pygeoapi natively supports on OpenAPI generation for feature/record `.../items` definitions.

As a result, this PR was updated to support this approach on queries to itemtypes (which now applies to all feature/record providers).  This PR was also tested against OAFeat CITE with no impact to conformance.

pygeoapi/provider/oracle.py
```diff
@@ -804,6 +817,7 @@ def query(
                     q,
                     language,
                     filterq,
+                    extra_params=extra_params
```
doublebyte1, line 820: +1 for adding this functionality. I think we should add something about these extra parameters for Oracle in the [documentation of the provider](https://github.com/geopython/pygeoapi/blob/master/docs/source/data-publishing/ogcapi-features.rst#oracle). Maybe just give an example of how this could be useful?

Moritz-Langer: I have added the documentation here:
https://github.com/geopython/pygeoapi/compare/master...Moritz-Langer:pygeoapi:master

If that resonates with you I can create a PR.

tomkralidis: +1 for a PR, thanks.

Moritz-Langer: PR is here:
https://github.com/geopython/pygeoapi/pull/2075
Thank you :)

Issue #1676, opened 2024-06-11
Process: allow the processor to know the process_id


francescoingv:
**Is your feature request related to a problem? Please describe.**
There are use cases where the process may benefit from knowing the processed `job_id`.
E.g.: a specific `ProcessManager` may want to read the output files from the process: the process would then write job specific files in the `job_id` directory.

**Describe the solution you'd like**
The setting should happen before the call to `process.execute()`.

One solution would be the `BaseManager` try to set the `processor.job_id`, but do nothing if it fails.
```
BaseManager.execute_process()
        processor = self.get_processor(process_id)
+        try:
+           # if the processor can set the job id, do it.
+            processor.set_job_id(job_id)
+        except AttributeError:
+            pass

```
Even better: the `BaseProcessor` defines a method `set_job_id()` that does nothing.
The method is available for derived classes, where needed.
```
BaseManager.execute_process()
        processor = self.get_processor(process_id)
+          processor.set_job_id(job_id)

BaseProcessor
+    def set_job_id(self, job_id: str) -> None:
+        """
+        set the job_id within the processor
+        To be implemented by derived classes where required.
+        """
+        pass
```

**Describe alternatives you've considered**
A derived `BaseManager` class implementing the solution should modify the `BaseManager.execute_process()`.
Given the minimal change, also **not affecting any in place code**, I would prefer to have the solution included in the release.

**Additional context**
If approved I would like to submit a PR with the second solution.

tomkralidis:
2024-06-13: +1 for option 2 implementation @francescoingv.

PR #1680, opened 2024-06-13
1676 Manager pass job_id to processor

francescoingv:
# Overview
Solve issue  #1676 using option 2

# Related Issue / discussion
Issue #1676 

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/process/base.py
+++ b/pygeoapi/process/base.py
@@ -57,6 +57,12 @@ class BaseProcessor:
         """
         set the job_id within the processor
         To be implemented by derived classes where required.
+
+        :job_id: the job_id assigned to the request by the Manager.
+            The function shuold be called by the Manager upon assigning
+            the job_id. The job_id is intended to be used by derived
+            classes, e.g. to write temporary files where filenames
+            contains the string job_id.
         """
         pass
 
--- a/pygeoapi/process/base.py
+++ b/pygeoapi/process/base.py
@@ -58,11 +58,12 @@ class BaseProcessor:
         set the job_id within the processor
         To be implemented by derived classes where required.
 
-        :job_id: the job_id assigned to the request by the Manager.
+        :param job_id: the job_id assigned to the request by the Manager.
             The function shuold be called by the Manager upon assigning
             the job_id. The job_id is intended to be used by derived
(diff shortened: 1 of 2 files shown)


francescoingv:
@tomkralidis
The reason for the failed check (Build / main (3.10) (pull_request) )
is not related to this PR.

pygeoapi/process/base.py
```diff
@@ -53,6 +53,13 @@ def __init__(self, processor_def: dict, process_metadata: dict):
         self.metadata = process_metadata
         self.supports_outputs = False
 
+    def set_job_id(self, job_id: str) -> None:
+        """
+        set the job_id within the processor
+        To be implemented by derived classes where required.
+        """
```
tomkralidis, line 60: Can you add a description of `job_id` in the docstring? i.e.

```python
 def set_job_id(self, job_id: str) -> None:
        """
        Sets a job identifier explicitly

        To be implemented by derived classes as required
        """

        :param job_id: job identifier

        :returns: `None`

        pass
```

tomkralidis (changes requested):

pygeoapi/process/base.py
tomkralidis: @francescoingv thanks for the update.  The return type needs to be added to the docstring.  As well, the `job_id` description needs the `param prefix`, and the text of the `job_id` descripition needs to be aligned.  So:

```python
 def set_job_id(self, job_id: str) -> None:
        """
        Sets a job identifier explicitly

        To be implemented by derived classes as required
        """

        set the job_id within the processor

        To be implemented by derived classes where required.

        :param job_id: the job_id assigned to the request by the Manager.
                    The function shuold be called by the Manager upon assigning
                    the job_id. The job_id is intended to be used by derived
                    classes, e.g. to write temporary files where filenames
                    contains the string job_id.

        :returns: `None`

        pass
```

tomkralidis (changes requested):

pygeoapi/process/base.py
francescoingv: @tomkralidis Thank you

tomkralidis (approved):

Issue #1655, opened 2024-05-14
Implement Connection Pooling for Oracle Provider to Improve Performance (potentially other Database Providers as well)


Moritz-Langer:
**Is your feature request related to a problem? Please describe.**

Performance improvements might be gained when using a connection pool for database providers. 
I have seen this commit to the postgres provider: https://github.com/geopython/pygeoapi/commit/644281359a5c622a7808b67ddf7773723ac60b4d 

**Describe the solution you'd like**
Upon application start check for config whether there are features/collections using the Oracle provider. If the Oracle provider is in use, the application should initialize a connection pool. Subsequent connections required by the Oracle provider should then be acquired from this pool rather than creating new connections per request. 

Open Question: Initialize Pool globally for example within itemtypes.py? 

**Describe alternatives you've considered**
I thought about handling within the oracle provider alone, but this might lead to initializing pools for every request or am I mistaking here? 

**Additional context**
Connection pooling can drastically reduce the overhead associated with establishing connections to the database, particularly under load. By managing a pool of reusable connections, the system can gain improvements in response time and reduce the load on the database server.

***Discussion Points***

- I invite feedback on the proposed solution, especially regarding where and how to initialize the connection pool.
- Insights into how connection pooling is being handled in similar projects or different providers would be extremely valuable.
- Any concerns or additional benefits not already considered?

Moritz-Langer:
On the weekend I started with an initial implementation along the line of the following:
```
class DatabaseConnection:
    """Database connection class to be used as 'with' statement.
    The class returns a connection object.
    """
    pool = None  # Class-level connection pool
    @classmethod
    def initialize_pool(cls, conn_dict):
        """Initialize the connection pool if not already initialized."""
        LOGGER.debug(f"conn_dict contains {conn_dict}")
        if DatabaseConnection.pool is None:
```
This looks promising and I will create a pull-request after some more polishing

tomkralidis:
cc'ing @totycro and @8luewater for feedback/input.

Moritz-Langer:
@8luewater  I am currently finishing up on this.
We ran internal tests using locust which showed an around 10x improvement in speed using session pooling. 

In the current design you would have to set environment variables like this to your linux / server environment:

    export ORACLE_POOL_MIN=2
    export ORACLE_POOL_MAX=10

Then the DatabaseConnection class uses these env variables as input parameters for the creation of a session pool.
[...]

PR #1688, opened 2024-06-19
Oracle Connection Pooling implementation of issues/1655 

Moritz-Langer:
# Overview
This is an implementation of connection pooling in oracle. It leaves the decision up to the environment, if the variables 
`ORACLE_POOL_MIN` and `ORACLE_POOL_MAX` are found in the environment, a Pool is created with min and max connections according to the variable values. If they are not set, than single connections are used, just like before. 
This should lead to no breaks for current users and is easy to set up. 

Tests and Documentation are added as well.

# Related Issue / discussion
Related to https://github.com/geopython/pygeoapi/issues/1655  

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/data-publishing/ogcapi-features.rst
+++ b/docs/source/data-publishing/ogcapi-features.rst
@@ -388,15 +388,17 @@ Extra properties is a list of strings which are added as fields for data retriev
 can be used to return expressions computed by the database.
 
 Session Pooling
-""""""""""""""""""""
+""""""""""""""""
+
 Configured using environment variables.
 
+.. code-block:: bash
 
    export ORACLE_POOL_MIN=2
-   
    export ORACLE_POOL_MAX=10
 
-The ORACLE_POOL_MIN and ORACLE_POOL_MAX environment variables are used to trigger Session Pool creation in the Oracle Provider and the DatabaseConnection class. See https://python-oracledb.readthedocs.io/en/latest/api_manual/module.html#oracledb.create_pool for Documentation of the create_pool function.
+
+The ``ORACLE_POOL_MIN`` and ``ORACLE_POOL_MAX`` environment variables are used to trigger session pool creation in the Oracle Provider and the ``DatabaseConnection`` class. See https://python-oracledb.readthedocs.io/en/latest/api_manual/module.html#oracledb.create_pool for documentation of the ``create_pool`` function.
 
 If none or only one of the environment variables is set, session pooling will not be activated and standalone connections are established at every request.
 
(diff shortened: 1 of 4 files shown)


Moritz-Langer:
@8luewater fyi

totycro (approved):

pygeoapi/provider/oracle.py
```diff
@@ -1,6 +1,7 @@
 # =================================================================
 #
 # Authors: Andreas Kosubek <andreas.kosubek@ama.gv.at>
+# Authors: Moritz Langer <moritz.b.langer@gmail.com>
 #
 # Copyright (c) 2023 Andreas Kosubek
```
tomkralidis, line 6: Can you also add yourself to the copyright?

docs/source/data-publishing/ogcapi-features.rst
```diff
@@ -387,6 +387,19 @@ Extra properties
 Extra properties is a list of strings which are added as fields for data retrieval in the SELECT clauses. They
 can be used to return expressions computed by the database.
 
+Session Pooling
```
tomkralidis, line 390: Suggested formatting updates:

```
Session Pooling
""""""""""""""""

Configured using environment variables.

.. code-block:: bash

   export ORACLE_POOL_MIN=2
   export ORACLE_POOL_MAX=10


The ``ORACLE_POOL_MIN`` and ``ORACLE_POOL_MAX`` environment variables are used to trigger session pool creation in the Oracle Provider and the ``DatabaseConnection`` class. See https://python-oracledb.readthedocs.io/en/latest/api_manual/module.html#oracledb.create_pool for documentation of the ``create_pool`` function.

If none or only one of the environment variables is set, session pooling will not be activated and standalone connections are established at every request.
```

pygeoapi/provider/oracle.py
```diff
@@ -33,6 +34,8 @@
 import oracledb
```
tomkralidis, line 34: Can we update the import ordering as follows:

```python
import importlib
import json
import logging
import os
import threading
from typing import Optional

import oracledb
import pyproj

from pygeoapi.api import DEFAULT_STORAGE_CRS
```

tomkralidis (approved):

tomkralidis (changes requested):
Requesting a few minor changes -- good work!

Moritz-Langer:
Hi Tom,
thank you for the feedback and the review. I have adjusted according to your comments.

pygeoapi/provider/oracle.py
Moritz-Langer: Changed accordingly

docs/source/data-publishing/ogcapi-features.rst
Moritz-Langer: Changed accordingly

pygeoapi/provider/oracle.py
Moritz-Langer: Changed accordingly

tomkralidis (approved):

Issue #1608, opened 2024-04-02
Time querying in `xarray_edr.py`


sjordan29:
**Description**
When I make an EDR query on a dataset for a single datetime, pygeoapi is returning an empty dataset. I think the following section is causing the error,

```python
        try:
            if select_properties:
                self.fields = {k: v for k, v in self.fields.items() if k in select_properties}  # noqa
                data = self._data[[*select_properties]]
            else:
                data = self._data

            if self.time_field in query_params:
                remaining_query = {
                    key: val for key, val in query_params.items()
                    if key != self.time_field
                }
                if isinstance(query_params[self.time_field], slice):
                    time_query = {
                        self.time_field: query_params[self.time_field]
                    }
                else:
                    time_query = {
                        self.time_field: (
                                data[self.time_field].dt.date ==
                                query_params[self.time_field]
                        )
                    }
                data = data.sel(
                    time_query).sel(remaining_query, method='nearest')
            else:
                data = data.sel(query_params, method='nearest')
```

specifically 
```python
time_query = {
    self.time_field: (
            data[self.time_field].dt.date ==
            query_params[self.time_field]
            )
```

**Steps to Reproduce**

An example dataset
```yaml
alaska_et_2020_ccsm4_historical_simulation:
    description:
      Gridded 20km Daily Reference Evapotranspiration for the State of
      Alaska from 1979 to 2017
    extents:
      spatial:
        bbox:
          - -179.9994335861477
          - 37.24836569956209
          - 179.99945621556017
          - 88.17225180794964
        crs: http://www.opengis.net/def/crs/OGC/1.3/CRS84
      temporal:
        begin: 1979-04-01 00:00:00+00:00
        end: 2100-09-30 00:00:00+00:00
    keywords:
      - s3
      - usgs
      - alaska_et_2020_ccsm4_historical_simulation
    providers:
      - data: s3://mdmf/gdp/alaska_et_2020_ccsm4_historical_simulation.zarr
        format:
          mimetype: application/zip
          name: zarr
        name: xarray-edr
        options:
          s3:
            anon: true
            client_kwargs:
              endpoint_url: https://usgs.osn.mghpcc.org/
        type: edr
        x_field: x
        y_field: y
    title: alaska_et_2020_ccsm4_historical_simulation
    type: collection
```

Here's the query:

```
[...]

PR #1690, opened 2024-06-20
update time querying

sjordan29:
# Overview
I was getting a No Data error when requesting a single datetime in EDR. 

# Related Issue / discussion
#1608

# Additional information
Note this will return the **closest** timestep. Would we want to throw a No Data error if there is not an exact match on datetime?

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute this bugfix to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/pygeoapi/provider/xarray_edr.py
+++ b/pygeoapi/provider/xarray_edr.py
@@ -115,23 +115,18 @@ class XarrayEDRProvider(BaseEDRProvider, XarrayProvider):
                 data = self._data
 
             if self.time_field in query_params:
-                remaining_query = {
-                    key: val for key, val in query_params.items()
-                    if key != self.time_field
-                }
                 if isinstance(query_params[self.time_field], slice):
+                    remaining_query = {
+                        key: val for key, val in query_params.items()
+                        if key != self.time_field
+                    }
                     time_query = {
                         self.time_field: query_params[self.time_field]
                     }
+                    data = data.sel(
+                        time_query).sel(remaining_query, method='nearest')
                 else:
-                    time_query = {
-                        self.time_field: (
-                                data[self.time_field].dt.date ==
-                                query_params[self.time_field]
-                        )
-                    }
-                data = data.sel(
-                    time_query).sel(remaining_query, method='nearest')
+                    data = data.sel(query_params, method='nearest')
             else:
                 data = data.sel(query_params, method='nearest')
         except KeyError:


sjordan29:
Hi @webb-ben and @tomkralidis, I'm not sure why the build is failing with this tweak to the `xarray_edr` provider. It fails on use ubuntuGIS unstable ppa and the local build of the image from dockerfile in the check vulnerabilities check.

webb-ben:
> Hi @webb-ben and @tomkralidis, I'm not sure why the build is failing with this tweak to the `xarray_edr` provider. It fails on use ubuntuGIS unstable ppa and the local build of the image from dockerfile in the check vulnerabilities check.

Re-ran, should be all good now

sjordan29:
Hm, I see from the tests now that this change is resulting in undesired behavior. I had made an update where you'd pull the nearest time for a single timestep (relevant code from tests below), but now I see you'd want to return empty no content instead.

```
    # some data
    req = mock_api_request({
        'coords': 'POINT(11 11)', 'datetime': '2000-01-16'
    })
    rsp_headers, code, response = get_collection_edr_query(
        api_, req, 'icoads-sst', None, 'position')
    assert code == HTTPStatus.OK

    # no data
    req = mock_api_request({
        'coords': 'POINT(11 11)', 'datetime': '2000-01-17'
    })
    rsp_headers, code, response = get_collection_edr_query(
        api_, req, 'icoads-sst', None, 'position')
    assert code == HTTPStatus.NO_CONTENT

```
It appears the issue I was trying to address with this is actually coming from the format of the datetimes in the underlying datasets I'm working with, which are formatted as floats of the days since a certain date. Seems like I will need to consider an alternate solution to this and scrap this PR.

image

Seems like I'd need to use `cftime`'s [`num2date` function](https://unidata.github.io/cftime/api.html) to make the times I'm working with queryable with a datetime syntax aligned with the EDR spec ([datetime-definition](https://docs.ogc.org/is/19-086r6/19-086r6.html#_1bf988a9-9249-466d-97b4-d583bf197335); [syntax](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6)) . Is that a case that you'd want to handle within pygeoapi (i.e., identify if there are numeric dates in the data being served and handle the request accordingly) or would it be preferable to reformat the dates in the underlying data?

sjordan29:
Coming back to this with fresh eyes and realizing that xarray should decode these times under the hood. I'll re-test the datasets I was working with to nail down what the real issue is here. Closing this PR.

PR #1709, opened 2024-07-03
Fix bullet point nesting in docs

C-Loftus:
# Overview

The nested lists on the [main page of the documentation site](https://docs.pygeoapi.io/en/stable/introduction.html) (and a few other pages) are not rendering properly. The rst for nested bullet points is not using proper syntax of adding an extra newline before and after the nested level. Thus, the * are rendered inline. For an example see the bottom of the attached image. This commit just adds extra lines to make the docs render properly.

I referenced the rst syntax [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html). 
> Nested lists are possible, but be aware that they must be separated from the parent list items by blank lines:

This commit is relatively trivial but helps improve a11y and styling.

# Dependency policy (RFC2) / Issues

N/A

# Updates to public demo

N/A

# Contributions and licensing

- [x] I'd like to contribute a docs fix to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

# Before
image

# After
image

### DIFF
--- a/docs/source/data-publishing/ogcapi-coverages.rst
+++ b/docs/source/data-publishing/ogcapi-coverages.rst
@@ -100,23 +100,41 @@ Data access examples
 --------------------
 
 * list all collections
+
   * http://localhost:5000/collections
+  
 * overview of dataset
+
   * http://localhost:5000/collections/foo
+  
 * schema of dataset
+
   * http://localhost:5000/collections/foo/schema
+  
 * coverage access via CoverageJSON (default)
+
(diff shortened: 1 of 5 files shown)


C-Loftus:
And of course if I am missing something here and this is a result of an autoformatter or something else in CI feel free to close

tomkralidis:
Thanks @C-Loftus !

Issue #1719, opened 2024-07-11
Test for TinyDBManager report successful while not performing any assert


francescoingv:
**Description**
Running the test `test_tinydb_manager_for_parallel_requests` **do not perform any assert**,
because when trying to execute of the process in `_execute_process()` the following exception is raised:
`AttributeError: 'API' object has no attribute 'execute_process'`
which is not caught; therefore the dictionary `processes_out` is not populated
consequently the code never enters the loop of asserts on the contents of `processes_out`.

**Steps to Reproduce**
I was able to find the issue running the test in debug.
In debug-console, for each of the 4 processes I get the following log:
```
Traceback (most recent call last):
  File "/usr/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/[...]/pygeoapi/tests/test_tinydb_manager_for_parallel_requests.py", line 64, in _execute_process
    headers, http_status, response = api.execute_process(request, process_id)
                                     ^^^^^^^^^^^^^^^^^^^
AttributeError: 'API' object has no attribute 'execute_process'

```

**Expected behavior**
In debug mode check the processes were completed.
Add an assertion to verify the processes run.

**Environment**
 - OS:
 - Python version: 3.11
 - pygeoapi version: 0.17

**Additional context**

**Suggestion**
Substitute the following line (currently 64):
`    headers, http_status, response = api.execute_process(request, process_id)`
with a call to:
`   api.manager.execute_process()`
passing the correct parameters

francescoingv:
@MartinPontius, @tomkralidis

PR #1720, opened 2024-07-12
Update test_tinydb_manager_for_parallel_requests.py

francescoingv:
# Overview
Updated the call to execute_process
and now checks the expected number of process results are present.

# Related Issue / discussion
Should fix issue #1719 

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


francescoingv:
@MartinPontius :
I did not change `_create_request()`,
however please consider that setting `data.mode="async"` has no effect
as actually the `execution_mode` is taken from the request header `"Prefer"`.

As it is now, the test run **synchronously**.

PR #1724, opened 2024-07-15
add support for TinyDB feature provider

### LINKED ISSUE #1723: add TinyDB feature provider
Enhance the TinyDBCatalogue Provider to additionally support features.

Provision of a TinyDB provider to publish feature data.

tomkralidis:
# Overview
Add support for TinyDB as a feature provider.

Additionally updates CITE test setup and docs.

# Related Issue / discussion
Fixes #1723 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


tests/test_tinydb_provider.py
```diff
@@ -0,0 +1,191 @@
+# =================================================================
+#
+# Authors: Tom Kralidis <tomkralidis@gmail.com>
+#
+# Copyright (c) 2023 Tom Kralidis
```
webb-ben, line 5: 2024?

webb-ben (approved):

tomkralidis:
FYI the associated CITE instance [PR](https://github.com/geopython/demo.pygeoapi.io/pull/58) can be merged if/once this is approved/merge.

tests/test_tinydb_provider.py
tomkralidis: Good catch!  Updated here and in `pygeoapi/plugin.py`.

docs/source/data-publishing/ogcapi-features.rst
```diff
@@ -31,6 +31,7 @@ parameters.
    `SQLiteGPKG`_,✅/❌,results/hits,✅,❌,❌,✅,❌,❌,✅
    `SensorThings API`_,✅/✅,results/hits,✅,✅,✅,✅,❌,❌,✅
    `Socrata`_,✅/✅,results/hits,✅,✅,✅,✅,❌,❌,✅
+   `TinyDB`_,✅/✅,results/hits,✅,✅,✅,✅,,❌✅,❌
```
justb4, line 34: align comma in table for TinyDB

pygeoapi/plugin.py
```diff
@@ -60,6 +60,7 @@
         'SensorThings': 'pygeoapi.provider.sensorthings.SensorThingsProvider',
         'SQLiteGPKG': 'pygeoapi.provider.sqlite.SQLiteGPKGProvider',
         'Socrata': 'pygeoapi.provider.socrata.SODAServiceProvider',
+        'TinyDB': 'pygeoapi.provider.tinydb_.TinyDBProvider',
```
justb4, line 63: Looks like `TinyDBCatalogueProvider` has been renamed to `TinyDBProvider` ? So can be removed here or for forward compat map to `TinyDBProvider`...

justb4: (ok I see that both are defined).

pygeoapi/provider/tinydb_.py
justb4, line 1: There are quite some `if self._catalogue` statements in the base class `TinyDBProvider`, while at the same time there is a subclass `TinyDBCatalogueProvider` . My suggestion is to refactor those conditionals into the subclass.

justb4 (changes requested):
Two main comments:
- refactor Catalogue-specific statements into subclass `TinyDBCatalogueProvider`
- add also unit tests for `TinyDBCatalogueProvider`

docs/source/data-publishing/ogcapi-features.rst
tomkralidis: Fixed.

pygeoapi/plugin.py
tomkralidis: Yes, there are nuances between a feature and record provider here.

pygeoapi/provider/tinydb_.py
tomkralidis: That was my original implementation, but I decided to add this conditional (3x) in the code so as not to have to loop after search results which would affect performance.

justb4: (maybe not use return values, but like: 
```
    def _add_text_search_field(self, fields):
        fields['q'] = {'type': 'string'}
        // return fields
```

tomkralidis:
> Two main comments:
> 
>     * refactor Catalogue-specific statements into subclass `TinyDBCatalogueProvider`
> 
>     * add also unit tests for `TinyDBCatalogueProvider`

Note that these already exist in `tests/test_tinydb_catalogue_provider.py`.

tomkralidis:
@justb4 PR updated, for review.

pygeoapi/provider/tinydb_.py
```diff
@@ -306,17 +314,69 @@ def delete(self, identifier):
...
+        return f'<TinyDBProvider> {self.data}'
+
+
+class TinyDBCatalogueProvider(TinyDBProvider):
+    """TinyDB Catalogue Provider"""
+
+    def __init__(self, provider_def):
+        super().__init__(provider_def)
+
+        self._excludes = ['_metadata-anytext']
```
justb4, line 350: Major: IMO will not have desired effect: `self.fields = self.get_fields()` is already called in the base class constructor. Or `self.fields = self.get_fields()` should be called again, but see my suggestion to make `_excludes` a class var as it is in effect a property of the classes.

justb4: To consider: `self.fields` is never referenced after assignment so maybe `self.fields = self.get_fields()` can be removed and the above `self._excludes =` -statement will be ok. `get_fields()` is a Provider base.py method and the way to access fields I guess.

justb4 (changes requested):
See inline comments: think all will be resolved by simply removing `self.fields = self.get_fields()` from constructor.

pygeoapi/provider/tinydb_.py
tomkralidis: If we remove `self.fields = self.get_fields()` from the constructor then this will affect other parts of `pygeoapi.api.itemtypes`.  Suggest we leave this as is in this PR and address in a subsequent PR later on.

justb4: Ok, then the only remaining is IMO to resolve the proper `_excludes`.

justb4: Do you agree that the current order will not work? Hence my suggestion for a class var (see my example) or another solution?

justb4 (changes requested):

pygeoapi/provider/tinydb_.py
tomkralidis: The default TinyDB provider as a result of this PR has an empty `_excludes` member.  The TinyDBCatalogue provider sets `_excludes` to remove a given property that is a pygeoapi "extra".

tomkralidis:
@justb4 this order works and accomplishes the need to remove the specified fields from catalogue responses.

pygeoapi/provider/tinydb_.py
webb-ben: fwiw other providers make the call to `self.get_fields()` at the end of the init method which will either fetch self.fields if set, or set self.fields.

[esri.py](https://github.com/geopython/pygeoapi/blob/e8ce02167fd670bb515f8216d699a3e9abca21d4/pygeoapi/provider/esri.py#L70)

[sqlite.py](https://github.com/geopython/pygeoapi/blob/e8ce02167fd670bb515f8216d699a3e9abca21d4/pygeoapi/provider/sqlite.py#L82)

Happy to establish a normative behavior between the two peoples `self.fields = self.get_fields()` is present in

[postgresql.py](https://github.com/geopython/pygeoapi/blob/e8ce02167fd670bb515f8216d699a3e9abca21d4/pygeoapi/provider/postgresql.py#L127)

[elasticsearch_.py](https://github.com/geopython/pygeoapi/blob/e8ce02167fd670bb515f8216d699a3e9abca21d4/pygeoapi/provider/elasticsearch_.py#L90)

justb4:
> @justb4 this order works and accomplishes the need to remove the specified fields from catalogue responses.

Hmm, I still don't think so. Here is a similar example I used to investigate, showing both the instance and class var:

```
class Base:
    # class variable
    class_var = ['Base']
    def __init__(self):
        self.exclude = 'Exclude by Base'
        print('Parent init')
        self.show_info()
    def show_info(self):
        # Accessing class variable
        print("Base: self.exclude=", self.exclude)
        print("Base: Base.class_var=", Base.class_var)

class Derived(Base):

    def __init__(self):
        print('Derived init')
        Base.class_var.append('Derived')
        super().__init__()
        self.exclude = 'Exclude by Derived'

    def show_info(self):
        # Accessing class variable
        super().show_info()
        print("Derived: self.exclude=", self.exclude)
        print("Derived: Base.class_var=", Base.class_var)

derived = Derived()
```
`self.exclude = 'Exclude by Derived'` is called after the Base constructor. Guess the outcome...

pygeoapi/provider/tinydb_.py
tomkralidis: Good points.  Suggest we address this as part of another PR

tomkralidis:
@justb4 feel free to test the branch with/without the exclude in the TinyDBCatalogue class?

justb4:
Ok, all tests pass. But makes sense as `p.get_fields()` is called. If I make
```
def test_query(config):
    p = TinyDBCatalogueProvider(config)

    fields = p.fields
```
as it results from constructor, the field is not excluded and test fails:

```
(pygeoapi-3.10.12) just@savu:~/project/pygeoapi/pygeoapi.git$ pytest tests/test_tinydb_catalogue_provider.py 
========================================================= test session starts =========================================================
platform darwin -- Python 3.10.12, pytest-8.2.2, pluggy-1.5.0
rootdir: /Users/just/project/pygeoapi/pygeoapi.git
configfile: pytest.ini
plugins: Faker-26.0.0, cov-5.0.0, env-1.1.3, anyio-4.4.0
collected 5 items                                                                                                                     

tests/test_tinydb_catalogue_provider.py F....                                                                                   [100%]

============================================================== FAILURES ===============================================================
_____________________________________________________________ test_query ______________________________________________________________

config = {'data': PosixPath('/private/var/folders/89/ggkklkwd5d7g9sw3pfbqwcvm0000gn/T/pytest-of-just/pytest-1/test_query0/sample-records.tinydb'), 'id_field': 'externalId', 'name': 'TinyDBCatalogue', 'time_field': 'created', ...}

    def test_query(config):
        p = TinyDBCatalogueProvider(config)
    
        fields = p.fields
>       assert len(fields) == 9
E       AssertionError: assert 10 == 9
E        +  where 10 = len({'_metadata-anytext': {'type': 'string'}, 'created': {'format': 'date-time', 'type': 'string'}, 'description': {'type': 'string'}, 'externalIds': {'type': 'string'}, ...})

tests/test_tinydb_catalogue_provider.py:97: AssertionError
======================================================= short test summary info =======================================================
FAILED tests/test_tinydb_catalogue_provider.py::test_query - AssertionError: assert 10 == 9
===================================================== 1 failed, 4 passed in 0.10s =====================================================
(py`
```
'_metadata-anytext' is _not_ excluded... as can be expected.

tomkralidis:
What if we normalized all access via `get_fields()`, and remove the `.fields` property?  Again, this would be part of another PR.

justb4:
Function access is always to be preferred over exposing internal instance vars.

But harder to assess (performance) impact. As for some Providers it may mean accessing a remote source like a DB or even WFS. But then again `get_fields()` may smart-cache. Could even be implemented once in Base class e.g. with `@property` ...

tomkralidis:
> \\@property
>     def fields(self):
>         return self.get_fields()

I guess you mean:

```python
@property
def fields(self):
    return self.get_fields()
```

Suggest we address as part of another PR (given the scope of this PR is to expand TinyDB support).  While we've identified some inconsistency/issue with `.fields`/`get_fields()`, this is better address in a separate PR given it will be a substantial change across numerous feature/record providers and their calling code.

@justb4 works for you?

justb4:
Ok, yes, but in my code I meant to cache the fields on first access (or leave that to each provider). Still in the current state `TinyDBCatalogueProvider` self.fields is incorrect. But don't know how much impact that has.

tomkralidis:
> ll in the current state `TinyDBCatalogueProvider` self.fields is incorrect. But don't know how much impact that has.

I think there is no impact in the context of the codepath and how the plugin is implemented.

justb4 (approved):
Ok, we'll fix 'excluds' issue later. For sake of progress.

doublebyte1 (commented):
Nice job!

Issue #1543, opened 2024-02-21
Upgrade test suite to Python 3.9


huard:
**Is your feature request related to a problem? Please describe.**

The CI runs on 3.8, but numpy and xarray have dropped support for 3.8 last year (see https://docs.xarray.dev/en/stable/whats-new.html#v2023-02-0-feb-7-2023). This is not causing any problem right now for me, but I suspect changes in upstream libraries could go undetected. 

**Describe the solution you'd like**
Update the python version running the CI to 3.9.

ricardogsilva:
In the spirit of [pygeoapi RFC2](https://pygeoapi.io/development/rfc/2/), it seems to me that the idea is to ensure dependencies are aligned with whatever is available in the current ubuntu LTS version - which at the time of writing is version 22.04.

In light of this, it seems to me that we actually ought to move to Python 3.10, which is the default version on ubuntu 22.04.

tomkralidis:
Thanks for the report.  Associated PR in #1586.

PR #1725, opened 2024-07-15
Update setup.py to restrict python version >= 3.9

pgValero:
Release 0.17.0 is incompatible with Python < 3.8

"python_requires" added to only install this version with python >= 3.9

# Overview

# Related Issue / discussion

I with python 3.8 the lastest version (0.17.0) is install but it's incompatible. Restrict this to install v0.16.1 is python < 3.9.

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -155,6 +155,7 @@ setup(
     maintainer='Tom Kralidis',
     maintainer_email='tomkralidis@gmail.com',
     url='https://pygeoapi.io',
+    python_requires='>=3.9',
     install_requires=read('requirements.txt').splitlines(),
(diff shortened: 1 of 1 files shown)


ricardogsilva:
Hi @pgValero 

Thanks for this, indeed we missed updating `setup.py` with the correct information. However, as mentioned in #1543, and subsequently merged in #1586 our target version of Python is 3.10 and not 3.9. Can you please update your PR to mention 3.10 instead?

pgValero:
Hi @ricardogsilva 

Perfect, thanks!

PR #1727, opened 2024-07-16
Use consistent get_field ref in providers

webb-ben:
# Overview
Use consistent method to fetch fields across providers. Of note, providers fields are always queried using the `p.get_fields()` method and not the `p.fields` property

# Related Issue / discussion
As discussed in https://github.com/geopython/pygeoapi/pull/1724

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/base.py
+++ b/pygeoapi/provider/base.py
@@ -93,15 +93,19 @@ class BaseProvider:
         raise NotImplementedError()
 
     @property
-    def fields(self):
+    def fields(self) -> dict:
         """
         Store provider field information (names, types)
 
-        Example response: {'field1': 'string', 'field2': 'number'}}
+        Example response: {'field1': {'type': 'string'}}
 
-        :returns: dict of field names and their associated JSON Schema types
+        :returns: dict of dicts (field names and their associated JSON Schema definitions)
         """
-        return self.get_fields()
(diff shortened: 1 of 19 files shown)


webb-ben:
This is ready for review. 

Of note, this essentially combines the `p.fields` property with `p.get_fields()` which will populate if empty and then return `p.fields_` dict

pygeoapi/provider/base.py
```diff
@@ -92,6 +92,17 @@ def get_fields(self):
 
         raise NotImplementedError()
 
+    @property
```
tomkralidis, line 95: The response should be a dict of dicts, so:

```python
    @property
    def fields(self) -> dict:
        """
        Store provider field information (names, types)

        Example response: {'field1': {'type': 'string'}}

        :returns: dict of dicts (field names and their associated JSON Schema definitions)
        """

        if hasattr(self, '_fields'):
            return self._fields
        else:
            return self.get_fields()
```

Given the above, a provider has the option to either set `self._fields` on init or to lazy load.  Then, from the downstream caller, the `.fields` accessor is opaque/consistent.

tomkralidis (changes requested):

Issue #1677, opened 2024-06-11
Add a ProcessManager backed by PostgreSql


francescoingv:
**Is your feature request related to a problem? Please describe.**
I would like to store the job data in a PostgreSql DB

**Describe the solution you'd like**
I developed a `PostgreSqlDBManager` based on `TinyDBManager`.
I would like to submit the file for additional checking:
could it be included in pygeoapi distribution, or in some alternative repository?

I would need also advice on how/where to release the script to create the required table(s).

I assumed the configuration parameters come from the configuration file,
however they would include user/password to access the DB: is there a better approach?

vprivat-ads:
> I assumed the configuration parameters come from the configuration file, however they would include user/password to access the DB: is there a better approach?

I guess that wouldn't be a problem given the current way to [configure the postgresql provider](https://github.com/geopython/pygeoapi/blob/master/tests/pygeoapi-test-config-postgresql.yml#L128):

```yaml
        providers:
          - type: feature
            name: PostgreSQL
            data:
                host: localhost
                dbname: test
                user: postgres
                password: postgres
                search_path: [osm, public]
            options:
                # Maximum time to wait while connecting, in seconds.
                connect_timeout: 10
                # Number of *milliseconds* that transmitted data may remain
                # unacknowledged before a connection is forcibly closed.
                tcp_user_timeout: 10000
                # Whether client-side TCP keepalives are used. 1 = use keepalives,
                # 0 = don't use keepalives.
                keepalives: 1
                # Number of seconds of inactivity after which TCP should send a
                # keepalive message to the server.
                keepalives_idle: 5
                # Number of TCP keepalives that can be lost before the client's
                # connection to the server is considered dead.
                keepalives_count: 5
                # Number of seconds after which a TCP keepalive message that is not
                # acknowledged by the server should be retransmitted.
                keepalives_interval: 1
            id_field: osm_id
            table: hotosm_bdi_waterways
            geom_field: foo_geom

```

PR #1730, opened 2024-07-16
Release Process Manager backed by Postgresql

francescoingv:
# Overview
Provide a Process Manager backed by PostgreSql database.

# Related Issue / discussion
#1677

# Additional information
Provided also dump of database structure
and test suite including the configuration file.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


pygeoapi/process/manager/postgresdb_.py
```diff
@@ -0,0 +1,261 @@
...
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# =================================================================
+
+# Requires postgresql database structure.
+# Import dump:
+# psql -U postgres -h 127.0.0.1 -p 5432 postgres <
+#   tests/data/postgres_manager_full_structure.backup.sql
+
+import json
```
tomkralidis, line 35: Imports should be in the order of standard library, third party packages, and local imports, alphabetically, grouped by spaces.  So:


```python
import json
import logging
from pathlib import Path
from typing import Any, Tuple

import psycopg2
import psycopg2.extras

from pygeoapi.process.manager.base import BaseManager
from pygeoapi.process.base import (
    JobNotFoundError,
    JobResultNotFoundError,
    ProcessorGenericError
)
from pygeoapi.util import JobStatus
```

pygeoapi/process/manager/postgresdb_.py
```diff
@@ -0,0 +1,261 @@
...
+    JobNotFoundError,
+    JobResultNotFoundError,
+    ProcessorGenericError,
+)
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class PostgreSqlDBManager(BaseManager):
```
tomkralidis, line 56: Suggest to name `PostgreSQLManager`

pygeoapi/process/manager/postgresdb_.py
tomkralidis, line 1: Suggest to name file `pygeoapi/process/manager/postgresql.py`

pygeoapi/process/manager/postgresdb_.py
```diff
@@ -0,0 +1,261 @@
...
+
+        self.__database_connection_parameters = manager_def['connection']
+        try:
+            # Test connection parameters:
+            test_query = """SELECT version()"""
+            with self.get_db_connection() as conn:
+                with conn.cursor() as cur:
+                    cur.execute(test_query)
+                    cur.fetchone()
+        except (Exception) as ex:
```
tomkralidis, line 80: Suggest:

`except Exception as err`

(`err` is a convention in the codebase)

pygeoapi/process/manager/postgresdb_.py
```diff
@@ -0,0 +1,261 @@
...
+            LOGGER.fatal('Test connecting to DB failed: ' + str(ex))
```
tomkralidis, line 81: Suggest `LOGGER.error(f'Test connecting to DB failed: {err}')`

pygeoapi/process/manager/postgresdb_.py
```diff
@@ -0,0 +1,261 @@
...
+            raise ProcessorGenericError(
+                'Test connecting to DB failed.'
+            ) from ex
```
tomkralidis, line 84: What is the implication of `from ex` here?  We typically raise as per `raise ProcessorGenericError('Test connecting to DB failed')`.

pygeoapi/process/manager/postgresdb_.py
```diff
@@ -0,0 +1,261 @@
...
+                                        be returned
+        :returns: `tuple` of mimetype and raw output
+        """
+
+        job_result = self.get_job(job_id)
+        location = job_result.get('location')
+        mimetype = job_result.get('mimetype')
+        job_status = JobStatus[job_result['status']]
+
+        if not job_status == JobStatus.successful:
```
tomkralidis, line 244: `if job_status != JobStatus.successful:`

pygeoapi/process/manager/postgresdb_.py
```diff
@@ -0,0 +1,261 @@
...
+            # Job is incomplete
+            return (None,)
+        if not location:
+            LOGGER.warning(f'job {job_id!r} -  unknown result location')
+            raise JobResultNotFoundError()
+        else:
+            try:
+                location = Path(location)
+                with location.open('r', encoding='utf-8') as filehandler:
```
tomkralidis, line 253: We don't need `'r'` as it is the the default mode for file opening.

As well, suggest to replace `filehandler` with `fh` (common convention in codebase).

requirements-postgres.txt
tomkralidis, line 1: Do we need this file?  We already have `psycopg2` in `requirements-provider.txt`.

tests/pygeoapi-test-config-postgres-manager.yml
```diff
@@ -0,0 +1,111 @@
...
+        - fr-CA
+    cors: true
+    pretty_print: true
+    limit: 10
+    # templates: /path/to/templates
+    map:
+        url: https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png
+        attribution: '<a href="https://wikimediafoundation.org/wiki/Maps_Terms_of_Use">Wikimedia maps</a> | Map data &copy; <a href="https://openstreetmap.org/copyright">OpenStreetMap contributors</a>'
+    manager:
+        name: 'pygeoapi.process.manager.postgresdb_.PostgreSqlDBManager'
```
tomkralidis, line 50: To keep consistent with the definition of `manager.connection`, suggest to define as a DB string formatting as per RFC1738.  Example:`postgresql://postgres:postgres@localhost:5432/ogc_api`

tests/test_postgresql_manager.py
```diff
@@ -0,0 +1,178 @@
...
+# HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+# WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# =================================================================
+
+# See pygeoapi/process/manager/postgresdb_.py
+# for instructions on setting up database structure.
+
```
tomkralidis, line 32: Imports should be in the order of standard library, third party packages, and local imports, alphabetically, grouped by spaces. So:

```python
import json

import pytest
from werkzeug.test import create_environ
from werkzeug.wrappers import Request

from .util import get_test_file_path
from pygeoapi.api import API, APIRequest
import pygeoapi.api.processes as processes_api
from pygeoapi.util import yaml_load
```

tomkralidis (changes requested):
Great work @francescoingv! Please see change requests.

In addition, the SQL statements be safeguarded against SQL injection.  This can be accomplished via variable binding or using SQLAlchemy as an ORM layer (see `pygeoapi/provider/postgresql.py` for an example).

pygeoapi/process/manager/postgresdb_.py
francescoingv: The implication is to not give useless information back.
Ref: https://stackoverflow.com/questions/24752395/python-raise-from-usage
Not still sure if appropriate.
I will change to the `typical usage`.

requirements-postgres.txt
francescoingv: We have `requirements-provider.txt` for providers,
but we do not have requirements for process manager.
If we assume all installations have the requirements for providers, then we should move the content of `requirements-provider.txt` in `requirements.txt`.
Possibly the file naming is not so good: what about `requirements-postgresql-manager.txt` ?

tests/pygeoapi-test-config-postgres-manager.yml
francescoingv: I followed the configuration present in the test for PostgreSQL providers: `test_postgresql_provider.py`.
I find it more clear the key-value configuration rather then an URI.

The used library (`psycopg2`), on receiving an URI to make a connection, delegate the parsing to `libqp` (https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.parse_dsn). Then psycopg2 depends also on `libqp` version (9.2 is required to parse URI).

The documentation of PostgreSQL reports that URI connection strings (https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) are accepted when following [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) (an update of RFC 1738).

I will modify the code to accept **also** URI,
and the test configuration file `pygeoapi-test-config-postgres-manager.yml` to insert comment on the possibility to use URI as well.

francescoingv:
> In addition, the SQL statements be safeguarded against SQL injection. This can be accomplished via variable binding or using SQLAlchemy as an ORM layer (see `pygeoapi/provider/postgresql.py` for an example).

Unless I missed something by mistake,
all SQL statements use variable binding (ref.: https://www.psycopg.org/psycopg3/docs/basic/params.html).

Please advise where not adhering to the above.

requirements-postgres.txt
tomkralidis: Suggest to rename, then, to `requirements-manager.txt`.

Having said this, we use `psycopg2` in `requirements-provider.txt` whereas in this PR `psycopg2-binary` is proposed.

I would suggest we use the same requirement across both and have `requirements-manager.txt` use `psycopg2`.  

We can then open another PR to use `psycopg2-binary` for all PostgreSQL support, to ensure feature providers also work accordingly.

tomkralidis:
> > In addition, the SQL statements be safeguarded against SQL injection. This can be accomplished via variable binding or using SQLAlchemy as an ORM layer (see `pygeoapi/provider/postgresql.py` for an example).
> 
> Unless I missed something by mistake, all SQL statements use variable binding (ref.: https://www.psycopg.org/psycopg3/docs/basic/params.html).
> 
> Please advise where not adhering to the above.

Thanks for the confirmation.  I am a bit uncomfortable with codepaths like https://github.com/geopython/pygeoapi/pull/1730/files#diff-84c6a5ee200ada2c0b338b21fe4d1429b70fdf9cb7470f5401b90bee087c0671R150-R163.  Is using SQLAlchemy possible?  It would make our PostgreSQL support consistent across the codebase (and not require the need to assemble SQL query strings).

requirements-postgres.txt
francescoingv: If you are the maintainer of a published package depending on psycopg2 you shouldn’t use psycopg2-binary as a module dependency. For production use you are advised to use the source distribution.

francescoingv: If you are the maintainer of a published package depending on psycopg2 you shouldn’t use psycopg2-binary as a module dependency. For production use you are advised to use the source distribution.

tests/data/postgres_manager_full_structure.backup.sql
tomkralidis, line 1: Taking a deeper look at this setup, can we update such that it includes, only the definitions to support job management?

- database is called `pygeoapi-manager`
- main jobs table is called `jobs`

This will keep the DB setup scoped to the job manager only.

tomkralidis (changes requested):

requirements-postgres.txt
francescoingv: As reported in psycopg2 documentation (https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary)

> The `psycopg2-binary` package is meant for beginners to start playing with Python and PostgreSQL without the need to meet the build requirements
> If you are the maintainer of a published package depending on **`psycopg2`** you shouldn’t use `psycopg2-binary` as a module dependency. **For production use you are advised to use the source distribution**.

I have no idea on how to make a complete build of `psycopg2`, removing `psycopg2-binary` from `requirements-manager.txt`.

For sure I would recommend **not** to use `psycopg2-binary` for all PostgreSQL.

tomkralidis: Can we keep this PR to `psycopg2` then, and issue another PR to use `psycopg2-binary` across the installation?  We need more eyes/review on this change, to ensure no impacts to build/deploy.  I'm happy to issue the next PR to harmonize.

francescoingv:
> Thanks for the confirmation. I am a bit uncomfortable with codepaths like https://github.com/geopython/pygeoapi/pull/1730/files#diff-84c6a5ee200ada2c0b338b21fe4d1429b70fdf9cb7470f5401b90bee087c0671R150-R163. Is using SQLAlchemy possible? It would make our PostgreSQL support consistent across the codebase (and not require the need to assemble SQL query strings).

It is surely possible to use SQLAlchemy,
but it is not something I know how to do it.

I would consider another PR for porting to SQLAlchemy.

requirements-postgres.txt
francescoingv: > Can we keep this PR to `psycopg2` then, and issue another PR to use `psycopg2-binary` across the installation? We need more eyes/review on this change, to ensure no impacts to build/deploy. I'm happy to issue the next PR to harmonize.

@tomkralidis I am not sure this is the right approach

I _guess_ in the Dockerfile of `geopython/pygeoapi` there is the **build** of `psycopg2`.
In that case it is correct to not use `psycopg2-binary` even in PostgreSQL manager.

On a machine where `psycopg2` **was not built  from source**,
having in `requirements-manager.txt` the following line only:
`psycopg2`


produce the following output:
```
(.venv) $ python -m pip install -r **requirements-manager.txt** 
Collecting psycopg2 (from -r requirements-manager.txt (line 1))
  Using cached psycopg2-2.9.9.tar.gz (384 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: psycopg2
  Building wheel for psycopg2 (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for psycopg2 (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [38 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.linux-x86_64-cpython-311
      creating build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/errors.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/extras.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/errorcodes.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/pool.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/extensions.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/_json.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/_range.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/__init__.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/_ipaddress.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/tz.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      copying lib/sql.py -> build/lib.linux-x86_64-cpython-311/psycopg2
      running build_ext
      building 'psycopg2._psycopg' extension
      creating build/temp.linux-x86_64-cpython-311
      creating build/temp.linux-x86_64-cpython-311/psycopg
      x86_64-linux-gnu-gcc -pthread -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -fPIC "-DPSYCOPG_VERSION=2.9.9 (dt dec pq3 ext lo64)" -DPSYCOPG_DEBUG=1 -DPG_VERSION_NUM=160003 -DHAVE_LO64=1 -DPSYCOPG_DEBUG=1 -I/home/francesco/Progetti/OGC_API/clone_pygeoapi/.venv/include -I/usr/include/python3.11 -I. -I/usr/include/postgresql -I/usr/include/postgresql/16/server -I/usr/include/libxml2 -c psycopg/adapter_asis.c -o build/temp.linux-x86_64-cpython-311/psycopg/adapter_asis.o -Wdeclaration-after-statement
      In file included from psycopg/adapter_asis.c:28:
      ./psycopg/psycopg.h:35:10: fatal error: Python.h: No such file or directory
         35 | #include <Python.h>
            |          ^~~~~~~~~~
      compilation terminated.
      
      It appears you are missing some prerequisite to build the package from source.
      
      You may install a binary package by installing 'psycopg2-binary' from PyPI.
      If you want to install psycopg2 from source, please install the packages
      required for the build and try again.
      
      For further information please check the 'doc/src/install.rst' file (also at
      <https://www.psycopg.org/docs/install.html>).
      
      error: command '/usr/bin/x86_64-linux-gnu-gcc' failed with exit code 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for psycopg2
Failed to build psycopg2
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (psycopg2)
(.venv) $

```

IMO an alternative solution for **not using** `psycopg2-binary` would be to have documented the requirement for building `psycopg2` from source.

francescoingv: Solved:
it was an issue in the configuration of my PC.

tomkralidis:
@francescoingv a few more comments upon deeper review:

#### Test data and CI
Thanks for updating the test data. We need to add these tests to CI, so in `.github/workflows/main.yml` (around [here](https://github.com/geopython/pygeoapi/blob/master/.github/workflows/main.yml#L106)) add:

```bash
psql postgresql://postgres:${{ secrets.DatabasePassword || 'postgres' }}@localhost:5432/test -f tests/data/postgres_manager_full_structure.backup.sql
```

and then, in https://github.com/geopython/pygeoapi/blob/master/.github/workflows/main.yml#L132, add:

```bash
pytest tests/test_postgresql_manager.py
```

Note: the PostgreSQL instance in CI already creates a database called `test`, so you need to remove any database creation/etc. in your SQL test data dump (see [example](https://github.com/geopython/pygeoapi/blob/master/tests/data/dummy_types_data.sql) of PostgreSQL test data for CI).

Then, in `tests/pygeoapi-test-config-postgres-manager.yml`, set the DB parameters accordingly to match the CI setup:

```yaml
    manager:
        name: 'pygeoapi.process.manager.postgresql.PostgreSQLManager'
        connection:
          host: localhost
          port: 5432
          database: test
          user: postgres
          password: ${POSTGRESQL_PASSWORD:-postgres}
        # Alternative accepted connection definition:
        # connection: postgresql://postgres:${POSTGRESQL_PASSWORD:-postgres}@localhost:5432/test
        output_dir: /tmp
```

#### Plugin registry integration

In https://github.com/geopython/pygeoapi/blob/master/pygeoapi/plugin.py#L78, add:

```python
'PostgreSQL': 'pygeoapi.process.manager.postgresql.PostgreSQLManager',
```

Following this, configuring a manager would look like (note `manager.name`):

```yaml
    manager:
        name: PostgreSQL
        connection:
          host: localhost
          port: 5432
          database: test
          user: postgres
          password: ${POSTGRESQL_PASSWORD:-postgres}
        # Alternative accepted connection definition:
        # connection: postgresql://postgres:${POSTGRESQL_PASSWORD:-postgres}@localhost:5432/test
        output_dir: /tmp
```

Issue #1735, opened 2024-07-17
promote process manager concept to the application level


tomkralidis:
**Is your feature request related to a problem? Please describe.**
The process manager was initially developed as a means to manage OGC API - Processes jobs.  It would be valuable to enable the manager to be able to provide the ability for pygeoapi downstream applications to manage application-level artifacts as needed.

**Describe the solution you'd like**
- move `pygeoapi.process.manager` to `pygeoapi.manager`
- build out documentation accordingly

**Describe alternatives you've considered**
None

**Additional context**
None

PR #1745, opened 2024-07-20
port PostgreSQL process manager to SQLAlchemy

tomkralidis:
# Overview
Updates PostgreSQL workflow to SQLALchemy for consistency with PostgreSQL feature provider.

# Related Issue / discussion
Follow up to https://github.com/geopython/pygeoapi/pull/1730#issuecomment-2233003293

# Additional information
cc @francescoingv for review/comment.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/process/manager/postgresql.py
+++ b/pygeoapi/process/manager/postgresql.py
@@ -45,15 +45,16 @@ import logging
 from pathlib import Path
 from typing import Any, Tuple
 
-import psycopg2
-import psycopg2.extras
+from sqlalchemy import insert, update, delete
+from sqlalchemy.orm import Session
 
-from pygeoapi.process.manager.base import BaseManager
 from pygeoapi.process.base import (
     JobNotFoundError,
     JobResultNotFoundError,
-    ProcessorGenericError,
+    ProcessorGenericError
 )
+from pygeoapi.process.manager.base import BaseManager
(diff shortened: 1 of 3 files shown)


webb-ben (approved):
Doesn't need to be a part of this PR but can we get a block documenting this configuration in https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-processes.html? 

Thanks for the port @tomkralidis

tomkralidis:
> Doesn't need to be a part of this PR but can we get a block documenting this configuration in https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-processes.html? 
> 
> Thanks for the port @tomkralidis 

Yes, I will be doing this in support of #1735.

francescoingv:
@tomkralidis It is fine for me
however the porting makes the manager not specific for PostgreSQL, but generic for all DB supported by SQLAlchemy.
Therefore I would suggest to change the filenames and class to something related to SQL or SQLAlchemy instead of PostgreSQL
If it were possible, in the test would be nice to verify also against different DB.

tomkralidis:
Thanks @francbartoli, good point.  I guess the same can be said for the PostgreSQL feature provider.  We can do as part of another PR when the time comes and the requirement arises.

PR #1764, opened 2024-07-29
fix EDR HTML breadcrumbs

tomkralidis:
# Overview
Fixes breadcrumbs and JSON link in EDR query HTML rendering.

# Related Issue / discussion
None

# Additional information
None
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/locale/en/LC_MESSAGES/messages.po
+++ b/locale/en/LC_MESSAGES/messages.po
@@ -726,3 +726,9 @@ msgstr ""
 
 msgid "Radius"
 msgstr ""
+
+msgid "Locations"
+msgstr ""
+
+msgid "Instances"
+msgstr ""
--- a/locale/en/LC_MESSAGES/messages.po
+++ b/locale/en/LC_MESSAGES/messages.po
@@ -708,3 +708,21 @@ msgstr ""
 
 msgid "not specified"
 msgstr ""
+
+msgid "Position"
+msgstr ""
+
+msgid "Cube"
+msgstr ""
+
+msgid "Area"
+msgstr ""
+
+msgid "Corridor"
+msgstr ""
+
+msgid "Trajectory"
+msgstr ""
+
+msgid "Radius"
+msgstr ""
--- a/pygeoapi/api/__init__.py
+++ b/pygeoapi/api/__init__.py
@@ -94,7 +94,6 @@ FORMAT_TYPES = OrderedDict((
     (F_HTML, 'text/html'),
     (F_JSONLD, 'application/ld+json'),
     (F_JSON, 'application/json'),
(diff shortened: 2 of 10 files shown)


locale/en/LC_MESSAGES/messages.po
webb-ben, line 1: Any reason we are omitting Locations, Items, and  Instances?

pygeoapi/api/__init__.py
```diff
@@ -80,6 +80,7 @@
 
 CHARSET = ['utf-8']
 F_JSON = 'json'
+F_COVERAGEJSON = 'json'
```
webb-ben, line 83: Why is this not in `FORMAT_TYPES`?

pygeoapi/api/environmental_data_retrieval.py
```diff
@@ -195,6 +198,31 @@ def get_collection_edr_query(api: API, request: APIRequest,
...
+        data['dataset_path'] = '/'.join(uri.split('/')[:-1])
+        data['collections_path'] = api.get_collections_url()
+
+        data['links'] = [{
+            'rel': 'collection',
+            'title': collections[dataset]['title'],
+            'href': data['dataset_path']
+        }, {
+            'type': 'application/prs.coverage+json',
+            'rel': request.get_linkrel(F_COVERAGEJSON),
```
webb-ben, line 221: EDR will not always be CovJSON

webb-ben (commented):
Small questions. Another one is about the JSON-LD representation of EDR. At the moment the way the HTML pages are rendered, they do not account for query args when requesting the JSON-LD content. Ass such, any `/cube/...?bbox=...` HTML representation will request a JSON-LD representation without providing the bbox, producing an error in pygeoapi.

pygeoapi/api/__init__.py
tomkralidis: Adding to `FORMAT_TYPES` will overwrite F_JSON (same key) in the subsequent dict.  We can address in a subsequent PR.

pygeoapi/api/environmental_data_retrieval.py
tomkralidis: Are we safe to assume, from a pygeoapi perspective, that CoverageJSON will be the baseline?

tomkralidis:
> Small questions. Another one is about the JSON-LD representation of EDR. At the moment the way the HTML pages are rendered, they do not account for query args when requesting the JSON-LD content. Ass such, any `/cube/...?bbox=...` HTML representation will request a JSON-LD representation without providing the bbox, producing an error in pygeoapi.

Good catch @webb-ben!  JSON-LD HTML link updated with query parameters.

webb-ben (approved):

Issue #1767, opened 2024-07-30
Add wallet authentication for Session Pools for Oracle Provider


Moritz-Langer:
**Is your feature request related to a problem? Please describe.**
Add wallet auth for session pools in oracle provider for addressing security concerns in production. Storing DB passwords in env vars or as plaintext in yml should be avoided.

**Describe the solution you'd like**
Adjust session pooling for checking against `external_auth: wallet` in provider definition under data. When wallet is configured use `externalauth=True` as session pool creation option. 

**Describe alternatives you've considered**

**Additional context**

PR #1768, opened 2024-07-30
Add Wallet for Session pool connections in oracle.py

Moritz-Langer:
# Overview
This pull request is linked to session pooling in the oracle provider. There should be an option for wallet authentication.

# Related Issue / discussion
Relates to https://github.com/geopython/pygeoapi/issues/1767

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


totycro (approved):

pygeoapi/provider/oracle.py
```diff
@@ -67,16 +67,31 @@ def create_pool(cls, conn_dict, oracle_pool_min, oracle_pool_max):
            Lock is implemented before function call at __init__"""
         dsn = cls._make_dsn(conn_dict)
         # Create the pool
+        if conn_dict.get("external_auth") == "wallet":
```
tomkralidis, line 70: Suggest optimizing as follows:


```python

import logging as LOGGER


@classmethod
def create_pool(cls, conn_dict, oracle_pool_min, oracle_pool_max):
    """Initialize the connection pool for the class
       Lock is implemented before function call at __init__"""

    dsn = cls._make_dsn(conn_dict)

    connect_kwargs = {
        'dsn': dsn,
        'min': oracle_pool_min,
        'max': oracle_pool_max,
        'increment': 1
    }

    # Create the pool
    if conn_dict.get("external_auth", "") == "wallet":
        # If Auth is via Wallet you need to save a wallet under
        # the directory returned by this bash command if apache is used
        # cat /etc/passwd |grep apache
        # except another directory is specified in the sqlnet.ora file
        LOGGER.debug("Connection pool from wallet.")
        connect_kwargs["externalauth"] = True
        connect_kwargs["homegeneous"] = False

    else:
        LOGGER.debug("Connection pool from user and password.")
        connect_kwargs["user"] = conn_dict["user"]
        connect_kwargs["password"] = conn_dict["password"]

    p = oracledb.create_pool(**connect_kwargs)
    LOGGER.debug("Connection pool created successfully")

    return p

```

tomkralidis (changes requested):
Thanks for this contribution @Moritz-Langer.  See change request.

As well, do we need any updates to the [Oracle provider docs](https://github.com/geopython/pygeoapi/blob/master/docs/source/data-publishing/ogcapi-features.rst#oracle)?

pygeoapi/provider/oracle.py
Moritz-Langer: Hi Tom, thank you for the feedback. I updated the provider with your feedback and added a brief reference to the docs for the configuration of wallet usage.

tomkralidis (approved):

PR #1770, opened 2024-08-02
fix breadcrumbs again (follow on of #1769)

tomkralidis:
# Overview
Mea culpa; forgotten commit of #1769
# Related Issue / discussion
#1769 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/templates/collections/schema.html
+++ b/pygeoapi/templates/collections/schema.html
@@ -3,7 +3,7 @@
 {% block crumbs %}{{ super() }}
 / <a href="{{ data['collections_path'] }}">{% trans %}Collections{% endtrans %}</a>
 / <a href="{{ data['dataset_path'] }}">{{ data['title'] | truncate( 25 ) }}</a>
-/ <a href="./{{ data['id'] }}schema">{% trans %}Schema{% endtrans %}</a>
+/ <a href="{{ data['dataset_path'] }}/schema">{% trans %}Schema{% endtrans %}</a>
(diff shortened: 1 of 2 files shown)


pygeoapi/templates/collections/schema.html
```diff
@@ -3,7 +3,7 @@
 {% block crumbs %}{{ super() }}
 / <a href="{{ data['collections_path'] }}">{% trans %}Collections{% endtrans %}</a>
 / <a href="{{ data['dataset_path'] }}">{{ data['title'] | truncate( 25 ) }}</a>
-/ <a href="./{{ data['id'] }}schema">{% trans %}Schema{% endtrans %}</a>
+/ <a href="{{ data['dataset_path'] }}/schema">{% trans %}Schema{% endtrans %}</a>
```
doublebyte1, line 6: @tomkralidis same comment as for queryables.

doublebyte1 (commented):
I don't see the need for the change, but maybe I am missing something here? (:

When I click the "schema" or "queryables" breadcrumbs they seem to be already working.

tomkralidis:
@doublebyte1 previous to #1769, the breadcrumb href was empty because of missing properties in the data object passed. In some cases, this would cause a 404 when clicked depending on how pygeoapi was deployed (subpath, etc.). This PR (as well as #1769) follows the same HTML rendering pattern as .../items, to safeguard the href.

Issue #1780, opened 2024-08-12
EDR Xarray covjson reordering y


pzaborowski:
**Description**
in 
provider/xarray_.py
gen_covjson function
following code:

```
minx, miny, maxx, maxy = metadata['bbox']
        mint, maxt = metadata['time']
        try:
            tmp_min = data.coords[self.y_field].values[0]
        except IndexError:
            tmp_min = data.coords[self.y_field].values
        try:
            tmp_max = data.coords[self.y_field].values[-1]
        except IndexError:
            tmp_max = data.coords[self.y_field].values

        if tmp_min > tmp_max:
            LOGGER.debug(f'Reversing direction of {self.y_field}')
            miny = tmp_max
            maxy = tmp_min
...
                    'y': {
                        'start': maxy,
                        'stop': miny,
```
Is reversing only axis north-south but not data. It breaks resulting coveragejson with inconsistent axis-data.
Either it shall be followed by the data reordering (not cheap so maybe not recommended) or axis start/stop shall be read from data.
Reading start/stop from data would also partly address the issue when data grid does not fill the query bbox and resulting array would not fill current axis.
**Steps to Reproduce**
Use reference configuration from https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-edr.html
View the results on the map - see it is upside down.

**Expected behavior**
1. Reordering of axis shall be removed or followed by data reordering - in case data is not expected organisation exception could be rised optionally.
2. start/stop of axis shall be read from data to make it consistent with variables (not bbox min/max which could make data stretched inproperly

**Screenshots/Tracebacks**

**Environment**
 - OS:
 - Python version:
 - pygeoapi version:

**Additional context**
Test data from https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-edr.html
 - tests/data/coads_sst.nc
 looks like longitude is in fact 0-360 not as in coords -180:180

pzaborowski:
same problem in demo deployment https://demo.pygeoapi.io/master/collections/icoads-sst/cube?bbox=-180,-90,180,90&datetime=2000-01-16%2F2000-02-01&select_properties=AIRT

PR #1781, opened 2024-08-12
EDR true axis y start/stop

pzaborowski:
Potential fix to 
https://github.com/geopython/pygeoapi/issues/1780

Reading from query bbox and reversing the order would not work anyway if not followed by reversing data (which is not cheap and would not do on default. It is visible on the sample data tests/data/coads_sst.nc where the start-stop swapping turn the data upside down.

# Overview
In the original implementation, start/stop for y axis are associated with max/min respectively read from the y data.
It looks like in case data is organised south-->north, they are reversed manually.
While it is probably correct for the only supported CRS to keep starty<stopy, the implementation probably does not work anyway as it is not followed by variables data reordering.
Test with tests/data/coads_sst.nc reference data shows the data would be turned upside down.

The proposed change does not influence behavior in case data is organized along descending lat (tested with https://psl.noaa.gov/data/gridded/data.coads.1deg.html as test/data/coads_sst.nc is anyway corrupted - values looks 0:360 and lat declared -180:180). Can cause  in case data is not organised in expected way it shall rise exception, but this change at least makes troubleshooting easier.

# Related Issue / discussion

# Additional information

# Dependency policy (RFC2)

- [X ] I have ensured that this PR meets [RFC2](https://pygeoapi.io/development/rfc/2) requirements

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/xarray_.py
+++ b/pygeoapi/provider/xarray_.py
@@ -317,7 +317,6 @@ class XarrayProvider(BaseProvider):
                               metadata['time_steps']]
                 }
                 cj['ranges'][key]['values'] = data[key].values.flatten().tolist()  # noqa
-                #cj['ranges'][key]['values'] = data[key].transpose('lat','lon','time').values.flatten().tolist()  # noqa
         except IndexError as err:
             LOGGER.warning(err)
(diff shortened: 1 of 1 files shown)


pzaborowski:
will update

Issue #1787, opened 2024-08-16
OGC API - Coverages collection with xarray provider fails to return CoverageJSON when a field is selected


JamesVarndell:
**Description**
This bug occurs when:
- Working with an OGC API - Coverages collection with an [xarray provider](https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-coverages.html#xarray-provider)
- Making a query with a [field selection](https://docs.ogc.org/DRAFTS/19-087.html#rc_fieldselection) (e.g. `parameters=t2m`), with the format set to JSON (`f=json`).

In the process of converting to CoverageJSON, the xarray backend fails when there is a field selection because this selection is not properly propagated down into the `gen_covjson` method. Let's say, for example, we have the following schema for our collection, which has dimensions of latitude, longitude and time:

```
{
    "type":"object",
    "title":"ERA5 temperature and pressure",
    "properties":{
        "t2m":{
            "type":"number",
            "title":"2 metre temperature",
            "x-ogc-unit":"K"
        },
        "msl":{
            "type":"number",
            "title":"Mean sea level pressure",
            "x-ogc-unit":"Pa"
        },
    },
    ...
}
```

And then we might make a request for just the temperature coverage in JSON, like this:
`localhost:5000/collections/era5-tp/coverage?f=json&properties=t2m`

Which raises the following error from xarray in the backend:
`KeyError: "No variable named 'msl'. Variables on the dataset include ['t2m', 'latitude', 'longitude', 'time']"`

This happens because the field is selected from the data successfully, but when converting to CoverageJSON, there is an attempt to extract all of the original fields from the resulting file - but `msl` is not in the result file, so the above error is raised.

The bug is caused by the [`fields` argument to `gen_covjson`](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/xarray_.py#L234) never being read - instead, the [full list of fields from the collection schema](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/xarray_.py#L297-L318) is used.

**Steps to Reproduce**
- Deploy an OGC API - Coverages collection with an [xarray provider](https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-coverages.html#xarray-provider) with more than one field available
- Make a query with a [field selection](https://docs.ogc.org/DRAFTS/19-087.html#rc_fieldselection) (e.g. `parameters=t2m`), with the format set to JSON (`f=json`).

**Expected behavior**
[...]

PR #1788, opened 2024-08-16
OGC API - Coverages: Propagate selected fields into covjson conversion

JamesVarndell:
# Overview

This PR fixes a bug which causes coverage collections with an xarray provider to fail to return CoverageJSON responses when a field selection is included in a query.

# Related Issue / discussion

#1787 

# Additional information

THis issue is caused by the `fields` argument to `gen_covjson` never being used to subset the fields when building the CoverageJSON. This argument contains the fields which have been selected in the request, and therefore the only fields which exist in the resulting xarray object - so only these fields should be added to the resulting CoverageJSON.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/xarray_.py
+++ b/pygeoapi/provider/xarray_.py
@@ -237,7 +237,7 @@ class XarrayProvider(BaseProvider):
 
         :param metadata: coverage metadata
         :param data: rasterio DatasetReader object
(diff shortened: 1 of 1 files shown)


tomkralidis:
Thanks for the fix @JamesVarndell ! 👍

PR #1794, opened 2024-08-20
Fix indentation in yaml example

PascalLike:
# Overview

Copying and pasting from the example in the documentation results in a broken YAML. A small triviality that can cost time :crying_cat_face: 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [X] I'd like to contribute docs to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution


doublebyte1 (approved):
Thanks for spotting and fixing this, @PascalLike ! 👍🏽

PR #1797, opened 2024-08-21
Display numberMatched in HTML view

webb-ben:
# Overview
This PR allows the pygeoapi templates to take advantage of the `numberMatched` property of pygeoapi items HTML view. It is a pleasant way to improve the richness of the HTML view by using properties of the items response that has already been generated. Additionally it allows for `/items?resulttype=hits` to serve an informative response on the HTML view.

It moves this content below the map as well.

# Related Issue / discussion

# Additional information
Hits HTML view
image

Normal items HTML view
image

items HTML with no numberMatched

Hits HTML view with no default numberMatched
image

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/locale/bs/LC_MESSAGES/messages.po
+++ b/locale/bs/LC_MESSAGES/messages.po
@@ -311,7 +311,7 @@ msgid "Items"
 msgstr "Stavke"
 
 #: pygeoapi/templates/collections/items/index.html:25
-msgid "items in this collection"
+msgid "Items in this collection"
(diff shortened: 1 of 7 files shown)


tomkralidis:
So as to keep the translations consistent, as well as not having to worry about plurals (1 item, 0 items, etc.), suggest:

`Items in this collection: 5`

webb-ben:
+1

tomkralidis (approved):

PR #1800, opened 2024-08-27
Improvements for xarray provider

barbuz:
# Overview

Using the xarray provider to manage Zarr data extensively, I have noticed and fixed a few bugs and added a couple of features that could be useful, so I would like to contribute these changes.

# Additional information

Summary of changes:
- Avoid crashes when the time dimension is not cf-compliant by trying to reload the data with `decode_times=False`
- Allow managing data without a time dimension at all
- The provider was already automatically managing subsets on coordinates sorted in reverse order, now it can do so also when applying a bounding box
- Fixed bug where some values where not converted to float64 before generating json output
- Used a named temporary file to generate netcdf output to allow use of netcdf4 engine  
- A couple of minor changes
- Allow requesting for netcdf or zarr output format regardless of the data format

This last change is the trickiest. I had to patch the APIRequest class to advertise the capability of producing more than one kind of output format for coverage data. At the moment I have hardcoded the two formats supported by the xarray provider, but eventually it would be nice to allow each provider to advertise an arbitrary list of output formats; I'm proposing this change as an initial implementation and a future PR could be able to expand on it and make this available to all providers. Please let me know what you think about this.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api/__init__.py
+++ b/pygeoapi/api/__init__.py
@@ -88,7 +88,6 @@ F_PNG = 'png'
 F_JPEG = 'jpeg'
 F_MVT = 'mvt'
 F_NETCDF = 'NetCDF'
-F_ZARR = 'zarr'
 
 #: Formats allowed for ?f= requests (order matters for complex MIME types)
 FORMAT_TYPES = OrderedDict((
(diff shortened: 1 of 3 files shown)


tomkralidis:
@barbuz are you able to rebase this PR?  I'd rather cover the format handling as part of another issue/discussion.  Are you able to reduce this PR to the pure Xarray fixes?

barbuz:
> @barbuz are you able to rebase this PR? I'd rather cover the format handling as part of another issue/discussion. Are you able to reduce this PR to the pure Xarray fixes?

I have rebased this reverting the changes to the format handling. I have opened a separate issue to discuss format handling for coverage providers in general: https://github.com/geopython/pygeoapi/issues/1821

Issue #1782, opened 2024-08-12
EDR time axis compliance with coverageJSON


pzaborowski:
**Description**
Current implementation of the XarrayEDRProvider generate time axis with label read from the source file into time_field with default value of 'time'. Time axes is defined by start, stop and steps.
But https://schemas.opengis.net/covjson/1.0/coveragejson.json
referred in https://docs.ogc.org/cs/21-069r2/21-069r2.html#_d5c16418-1a20-4dbf-bf7a-8e685062df97
expect 't' as time axis with 'values'.

This way it is implemented in https://github.com/Reading-eScience-Centre/leaflet-coverage to properly visualise coverage with >1 time slice (it does not serve other levels and start-stop definition).

**Steps to Reproduce**
open:
https://demo.pygeoapi.io/master/collections/icoads-sst/cube?f=html&bbox=-180,-90,180,90&datetime=2000-01-16/2000-03-17
select AIRT and get this 
<img width="1293" alt="image" src="https://github.com/user-attachments/assets/043bde3f-bd54-4461-9aa3-133ddbea631a">
It looks like the template issue, which is partly true as it could be solved with changes in query template like in this commit:
https://github.com/geopython/pygeoapi/commit/f4f24f05f4d11c774c086e2fda56c879a5c874e6
but these will not will not work with LeafletCoverage library without chaging time axis to t or changes in the JS libraries to support non-standard encoding.

**Expected behavior**
Coverage with >1 time value shall be presented as separate layers.
With current state it can be done as in the commit above but potentially breaking changes or changes in coveragejson schema and leaflet plugin. something in the middle is switch in the configuration that could render 'values' instead of 'start-stop-steps' but not sure if preferable. let me know< I can PR what is done already.

PR #1814, opened 2024-09-20
EDR CoverageJSON conformance fixes: time axis and demo data

pzaborowski:
# Overview
changes to confirm to https://schemas.opengis.net/covjson/1.0/coveragejson.json
- time axis changed to values (was interval)
- time axis first in range as expected by https://github.com/Reading-eScience-Centre/leaflet-coverage and dependancies
- parameter description as object - worked with JS visualisation but not validator and covjson playground
- range dataType as float (was number and allowed are float, integer, string) - worked with JS visualisation but not validator and covjson playground

Other:
 - typo in edr/query.html template causing issues with first parameter layer
 -  - Changes from https://github.com/geopython/pygeoapi/pull/1811 included

Changes shall solve issues using https://covjson.org libraries, also on pygeoapi demo site.

# Related Issue / discussion
mainly https://github.com/geopython/pygeoapi/issues/1833
but pull depends on this
https://github.com/geopython/pygeoapi/issues/1782

# Additional information
changes is not fully backward compatible with potential non-standard client using time as interval (which is breaking coveragejson-schema).

In addition to code changes, coads_sst.nc is fixed. Demo file had COADSX shifted west but not data which looks like this:
image

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


tests/api/test_environmental_data_retrieval.py
```diff
@@ -148,29 +151,36 @@ def test_get_collection_edr_query(config, api_):
...
-    assert time_dict['stop'] == '2000-12-16T01:20:05.999999996'
-    assert time_dict['num'] == 7
+    assert time_dict['values'] == ['2000-08-16T07:23:42.000000000',
+                                   '2000-09-15T17:52:48.000000000',
+                                   '2000-10-16T04:21:54.000000000',
+                                   '2000-11-15T14:51:00.000000000',
+                                   '2000-12-16T01:20:05.999999996']
+    # assert time_dict['start'] == '2000-08-16T07:23:42.000000000'
+    # assert time_dict['stop'] == '2000-12-16T01:20:05.999999996'
+    # assert time_dict['num'] == 7
```
webb-ben, line 179:
```suggestion
```

pygeoapi/templates/collections/edr/query.html
webb-ben, line 1: This should be already addressed in upstream pygeoapi

tests/api/test_environmental_data_retrieval.py
```diff
@@ -148,29 +151,36 @@ def test_get_collection_edr_query(config, api_):
...
 
-    assert time_dict['start'] == '2000-01-16T06:00:00.000000000'
-    assert time_dict['stop'] == '2000-06-16T10:25:30.000000000'
-    assert time_dict['num'] == 6
+    assert time_dict['values'] == ['2000-06-16T10:25:30.000000000',
+                                   '2000-07-16T20:54:36.000000000',
+                                   '2000-08-16T07:23:42.000000000']
+    # assert time_dict['start'] == '2000-06-16T10:25:30.000000000'
+    # assert time_dict['stop'] == '2000-08-16T07:23:42.000000000'
+    # assert time_dict['num'] == 3
```
webb-ben, line 159:
```suggestion
```

tests/api/test_environmental_data_retrieval.py
```diff
@@ -132,14 +132,17 @@ def test_get_collection_edr_query(config, api_):
...
 
-    assert time_dict['start'] == '2000-02-15T16:29:05.999999999'
-    assert time_dict['stop'] == '2000-06-16T10:25:30.000000000'
-    assert time_dict['num'] == 5
+    assert time_dict['values'] == ['2000-06-16T10:25:30.000000000',
+                                   '2000-07-16T20:54:36.000000000',
+                                   '2000-08-16T07:23:42.000000000']
+    # assert time_dict['start'] == '2000-06-16T10:25:30.000000000'
+    # assert time_dict['stop'] == '2000-08-16T07:23:42.000000000'
+    # assert time_dict['num'] == 3
```
webb-ben, line 140:
```suggestion
```

pygeoapi/provider/xarray_.py
```diff
@@ -92,6 +92,8 @@ def __init__(self, provider_def):
             self.axes = [self._coverage_properties['x_axis_label'],
                          self._coverage_properties['y_axis_label'],
                          self._coverage_properties['time_axis_label']]
+            self.time_axis_covjson = provider_def.get('time_axis_covjson') \
+                or self.time_field
```
webb-ben, line 96: What is the point of adding this? Can we add a test case that uses this field. Suggest to potentially rename in pygeoapi, but will defer to @tomkralidis. This addition should also be noted in the docs ([OACoverages](https://github.com/geopython/pygeoapi/blob/master/docs/source/data-publishing/ogcapi-coverages.rst) and [OAEDR](https://github.com/geopython/pygeoapi/blob/master/docs/source/data-publishing/ogcapi-edr.rst)

webb-ben (commented):
Looks good. Please rebase!

pygeoapi/provider/xarray_.py
pzaborowski: used it to configure the time axis label (see https://github.com/geopython/pygeoapi/issues/1833, coverage schema expect 't' as the time axis) and keep an option for backward compatibility in case the implemented client needs the other one.

Issue #1661, opened 2024-05-15
Allow using sqlalchemy > 2


ricardogsilva:
Now that [pygeofilter supports sqlalchemy v2](https://github.com/geopython/pygeofilter/issues/67), and with the changes merged in #1520, it seems pygeoapi can now lift the current restriction on using sqlalchemy<2, as seen in:

https://github.com/geopython/pygeoapi/blob/6c31a8e371d10750f265357bab62b45bfa9abc8f/requirements.txt#L17

This is related to #1123, but I am not able to re-open it - feel free to close this and re-open that if appropriate.

tomkralidis:
cc @kalxas 

@ricardogsilva as discussed a couple of weeks ago, looks like SQLAlchemy is still at 1.x on [jammy](https://packages.ubuntu.com/jammy/python3-sqlalchemy).  We'll need to monitor accordingly and update once we have some updates from upstream.

ricardogsilva:
@tomkralidis 

Just a small nitpick, current Ubuntu LTS is now 24.04 (Noble Numbat) - but your point is still valid, the `python3-sqlalchemy` package is still at version 1.4.50 as per:

https://packages.ubuntu.com/noble/python3-sqlalchemy

According to [sqlalchemy docs](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#overview), v1.4 is compatible with version 2 and in #1520 I had already fixed existing sqlalchemy transition to v2 warnings in the codebase.

This to say that I think we are now in a position whete it is possible to simply remove the `sqlalchemy<v2` pin in our requirements files and things should work OK both on Ubuntu LTS, as per RFC2 and would also allow other installs to move forward with using sqlalchemy v2+.

When I find some time I will check whether everything works OK with the pin removed, most likely by trying to run the test suite inside some ubuntu:24.04 container.

kalxas:
keeping this one open

PR #1832, opened 2024-10-21
Lifted pin on sqlalchemy in order to be able to use v2+

ricardogsilva:
# Overview

This PR lifts the pin on the `sqlalchemy` dependency, removing the requirement to use a version lower than `v2`.

# Related Issue / discussion

- #1661

# Additional information

WRT RFC2, and as mentioned in my previous comment at the issue:

https://github.com/geopython/pygeoapi/issues/1661#issuecomment-2249817527

The current code base should be compatible with both sqlalchemy `v1.4` ([which is what is included in current ubuntu LTS repos](https://packages.ubuntu.com/noble/python3-sqlalchemy)) and with `v2`, which means we can continue to use v1.4 from the current ubuntu LTS repos when doing a 'bare metal' install and also means we can use the current version of sqlalchemy when installing via other methods, like pip.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,7 +94,6 @@ jobs:
         pip3 install -r requirements-manager.txt
         pip3 install -r requirements-django.txt
         python3 setup.py install
-        pip3 install --upgrade "sqlalchemy<2"
         pip3 install --global-option=build_ext --global-option="-I/usr/include/gdal" GDAL==`gdal-config --version`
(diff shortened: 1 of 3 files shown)


ricardogsilva:
@tomkralidis CI error seems unrelated to the changes in the PR - seems like some trivy limit has been reached. I think it would likely go away if the job is re-run.  I am not able to ask for a re-run of the failed CI JOB though

PR #1841, opened 2024-11-05
Adds extensible possibilities to the API class to support dynamic resources loading and hot-reload logic mechanisms

Alex-NRCan:
# Overview

This PR adds 2 methods and 2 overridable methods in the API class to allow for:
- (1) loading resources from somewhere else than the yaml file and;
- (2) hot-reload the resources dynamically, through multiple pygeoapi instances (e.g. in a load balanced environment) without having to restart any pygeoapi instances.

This PR also adds a decorative function to make sure the resources are always up to date with the possibly dynamic resources material.

With the overridable functions implementations, this PR doesn't alter the native way pygeoapi works.

# Related Issue / discussion

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/crs.rst
+++ b/docs/source/crs.rst
@@ -30,7 +30,7 @@ Metadata
 The conformance class `http://www.opengis.net/spec/ogcapi-features-2/1.0/conf/crs` is present as a `conformsTo` field
 in the root landing page response.
 
-The configured CRSs, or their defaults, `crs` and `storageCRS` and optionally `storageCrsCoordinateEpoch` will be present in the "Describe Collection" response.
+The configured CRSs, or their defaults, `crs` and `storageCrs` and optionally `storageCrsCoordinateEpoch` will be present in the "Describe Collection" response.
 
 Parameters
 ----------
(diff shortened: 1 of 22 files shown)


Alex-NRCan:
Hey totycro,
`last_loaded_resources` is used to keep track of the last time the resources were loaded in the pygeoapi instance, indeed.

That way, in the custom implementation override (i.e in `on_load_resources_check` function), one can verify if that date is "too old" compared to whatever principle one has in place to determine that.

For example, in our implementation, we have all resources stored in a database (not the yaml file), because we have multiple pygeoapi instances running in parallel (load balanced) and we wanted them all to use the resources centralized in a database and those resources can change any time. When the resources change, we flag a date in the database about it. Therefore, in our `on_load_resources_check` implementation, we check if the running pygeoapi instance's `last_loaded_resources` date is older than the date in the database and when so, that particular pygeoapi instance reloads itself automatically.

That being said, one can have different ways of checking for a reload flag, it doesn't have to be a central database. That is why the `on_load_resources_check` simply returns a boolean. Up to the developer to implement their custom check. Similarly, it's also up to the developer to implement `on_load_resources` which can be anything, not necessarily a central database either.

Alex-NRCan:
@tomkralidis I've noticed that this PR has been closed. Was it implemented or was this PR closed by mistake? I don't seem to find hot-reloading mechanisms merged in the last couple weeks. We've been maintaining that PR synchronized for many months in hope of getting it merged one day :)

Issue #1840, opened 2024-11-04
Continue refactoring of API #1405 in remaining files


totycro:
The api refactoring #1405 introduced a new way of writing APIs. The following two files have not been transformed yet since the change set was already massive:
https://github.com/geopython/pygeoapi/blob/master/pygeoapi/api/__init__.py#L689
https://github.com/geopython/pygeoapi/blob/master/pygeoapi/admin.py#L160%7C

As mentioned in this comment in the review of the PR, these should now be addressed: https://github.com/geopython/pygeoapi/pull/1405#discussion_r1545762696

Transforming the  so would allow us to delete some code such as `@pre_process` and `APIRequest.with_data` and have a more uniform code base.

PR #1485, opened 2024-01-09
import Admin module based on runtime configuration (#1469)

tomkralidis:
# Overview
Loads Admin module conditionally.
# Related issue / discussion
Fixes #1469
# Additional information
None
# Dependency policy (RFC2)

- [x] I have ensured that this PR meets [RFC2](https://pygeoapi.io/development/rfc/2) requirements

# Updates to public demo

- [x] I have ensured that breaking changes to the [pygeoapi master demo server](https://github.com/geopython/demo.pygeoapi.io) have been addressed
  - [ ] https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_master/local.config.yml

# Contributions and licensing

(as per https://github.com/geopython/pygeoapi/blob/master/CONTRIBUTING.md#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

Issue #1469, opened 2024-01-03
ModuleNotFoundError: No module named 'jsonpatch'


krishnaglodha:
**Description**
I'm trying to execute `Install in 5 minutes` on https://pygeoapi.io/ , and getting above mentioned error. 

**Steps to Reproduce**
```
python3 -m venv pygeoapi
cd pygeoapi
. bin/activate
git clone https://github.com/geopython/pygeoapi.git
cd pygeoapi
pip3 install -r requirements.txt
python3 setup.py install
cp pygeoapi-config.yml example-config.yml
vi example-config.yml  # edit as required
export PYGEOAPI_CONFIG=example-config.yml
export PYGEOAPI_OPENAPI=example-openapi.yml
pygeoapi openapi generate $PYGEOAPI_CONFIG --output-file $PYGEOAPI_OPENAPI
pygeoapi serve

```
**Screenshots/Tracebacks**
<img width="570" alt="Screenshot 2024-01-03 at 9 53 12 PM" src="https://github.com/geopython/pygeoapi/assets/47075664/a916c462-e161-4476-8a3f-9d3a658ca5b1">

**Environment**
 - OS: Mac OS
 - Python version: 3.11
 - pygeoapi version: latest

krishnaglodha:
Should we mention it on the [admin page ](https://docs.pygeoapi.io/en/latest/admin-api.html) that to run is successfully, user should also execute `pip install -r requirements-admin.txt`

doublebyte1:
@krishnaglodha well spotted! 👍🏽  Do you want to do a PR to add that line to the docs?

https://github.com/geopython/pygeoapi/blob/d66e790df18ddcaa984984c2cb6a25c97c21bdee/docs/source/installation.rst?plain=1#L29

krishnaglodha:
Will do

Thanks & Regards,
Krishna G. Lodha
Founder
office email - ***@***.***
personal email - ***@***.***

krishnaglodha:
I propose to have this on the doc page of admin, because we are not sure that everyone will be using admin function, thus, let people know about this only on admin doc page ?

ricardogsilva:
@krishnaglodha 

If that would be the case, then the packages specified on `requirements-admin.txt` ought to be optional, which does not seem to be the case currently, which I guess is a bug? 

The admin is being turned on conditionally, depending on the pygeoapi config. However, the import of `jsonpatch` happens regardless.

ricardogsilva:
[...]

PR #1842, opened 2024-11-06
Continue refactoring of API #1405 in remaining files

totycro:
# Overview

This applies the same refactoring as in #1405 to the remaining files (admin api and `api/__init__.py`). See #1840
   
This removes nearly 300 lines without any changes in functionality.

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


totycro:
The vulnerability check fails because the respective docker image repository is overloaded. Unit tests and linting does pass.

ricardogsilva:
@totycro @tomkralidis 

Sorry to be late to the party but this introduces a regression whereby the flask admin is imported unconditionally, thus making `jsonpatch` become a hard dependency.

It is the same thing we had discussed in #1469 and which @tomkralidis had addressed with #1485.

The change now merged in:

https://github.com/geopython/pygeoapi/blob/f181d99b5756c258a7429d0fb45989e1917679c9/pygeoapi/flask_app.py#L49

results in the unconditional import of the admin module, which in turn imports `jsonpatch`. This dependency is only installed when pulling in the `requirements-admin.txt` file. 

The result of this is that the 'Install in 5 minutes' section in the docs becomes broken and that `requirements-admin.txt` effectively becomes mandatory, even if the config disables the admin.

PR #1845, opened 2024-11-13
Optional contact properties

FritzHoing:
# Overview

# Additional information

In the current version all contact information's (like a fax number e.g.) are needed.
This MR adds conditional checks for all properties and just renders the ones supplied in the config.yml.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/openapi.py
+++ b/pygeoapi/openapi.py
@@ -150,32 +150,54 @@ def gen_contact(cfg: dict) -> dict:
         'email': cfg['metadata']['contact']['email']
     }
 
-    contact['x-ogc-serviceContact'] = {
-        'name': cfg['metadata']['contact']['name'],
-        'position': cfg['metadata']['contact']['position'],
-        'addresses': [{
-            'deliveryPoint': [cfg['metadata']['contact']['address']],
-            'city': cfg['metadata']['contact']['city'],
-            'administrativeArea': cfg['metadata']['contact']['stateorprovince'],  # noqa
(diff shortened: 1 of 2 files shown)


pygeoapi/openapi.py
```diff
@@ -150,32 +150,54 @@ def gen_contact(cfg: dict) -> dict:
...
-            'type': 'fax', 'value': cfg['metadata']['contact']['fax']
-        }],
-        'emails': [{
-            'value': cfg['metadata']['contact']['email']
-        }],
-        'contactInstructions': cfg['metadata']['contact']['instructions'],
-        'links': [{
+    service_contact = {}
+    if cfg['metadata']['contact'].get('name'):
+        service_contact['name'] = cfg['metadata']['contact']['name']
```
huard, line 155: Note that the `if` here checks if the attribute exists AND evaluates as `True`. This means that if a field has value `0`, it won't be copied into the output dictionary. Granted, it's unlikely to be the case here, but I'd suggest to use `if "name" in cfg['metadata']['contact']`.
```python
def dmap(source: dict, mapping: dict, target:dict = None) -> dict:
    """Map keys from source to target using mapping."""
    if target is None:
        target = {}
    else:
        target = target.copy()

    for key, val in mapping.items():
        if key in source:
            target[val] = source[key]

    return target
```

There's apparently a complete library to make those types of conversions: https://pypi.org/project/json-converter/

tomkralidis:
Closing in lieu of #1862

PR #1861, opened 2024-12-05
Add License and Extent information to Collections

sjordan29:
# Overview

We ran into some issues where the STAC created with the Hateoas provider could not be parsed with pystac. Seems like part of our issue was due to the fact that pystac requires "license" and "extent" info at the collection level (per [STAC spec](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md)). Since this info wasn't in the content of collections in the hateoas provider, pystac would fail. Adding that information to the content returned by the hateoas provider allows the catalog to be parsed with pystac. 

# Related Issue / discussion
Discussion [here](https://code.usgs.gov/wma/nhgf/pygeoapi/-/issues/29#note_674325) with @dblodgett-usgs. Can open an issue in the pygeoapi repository as well if helpful. 

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute this bugfix to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution


pygeoapi/provider/hateoas.py
```diff
@@ -120,6 +120,8 @@ def get_data_path(self, baseurl, urlpath, entrypath):
             try:
                 jsondata = _get_json_data(f'{data_path}/collection.json')
                 resource_type = 'Collection'
+                content['license'] = jsondata['license']
```
tomkralidis, line 123: Can we safeguard? i.e.:

```python
for key in ['license', 'extent']:
    if key in jsondata:
        content[key] = jsondata[key]
```

tomkralidis (changes requested):

pygeoapi/provider/hateoas.py
sjordan29: Updated!

PR #1888, opened 2024-12-23
Admin tests without live server

totycro:
# Overview

This PR creates new admin api tests that work without a live server. The live server tests are still there and can be executed manually, they are not run in CI though.

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


webb-ben (approved):
+1 looks good. 

Doesn't need to be a part of this PR but it would also be valuable to test against common invalid configurations to make sure the Admin API both does not implement the changes but also returns informative errors as to why the change was not applied.

PR #1891, opened 2024-12-30
Implement OGC API Features Part 4 transactions for the postgres provider

totycro:
# Overview

This PR implements `create()`, `update()` and `delete()` in the postgres provider. The behavior is analogous to implementations in other providers and as defined by the standard.

The implementation is pretty basic, i can add more advanced logic if necessary (e.g. inputs which are not in WGS 84).

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


tomkralidis (changes requested):
PR looks good, thanks @totycro !

Please mark transactions as supported in PostgreSQL in the [providers table](https://docs.pygeoapi.io/en/latest/data-publishing/ogcapi-features.html#providers).

Issue #1856, opened 2024-11-20
enhance support for limits (RFC5)


tomkralidis:
Provide enhanced limit handling implementation in support of better definition and control on data query limiting.

RFC5 can be found at https://pygeoapi.io/development/rfc/5.

webb-ben:
@Youssef-Harby @tomkralidis bringing part of the conversation about "unlimited" defaultitems configuration in pygeoapi. Given the relative heterogeneity of providers, I wonder if it possible to realistically implement that on a provider level. At least I wonder if the most effective implementation would leverage the `hits` response and then querying by the literal limit. For various systems that interact with pygeoapi, this is generally the behavior I do to get "all" the items returned. See
https://github.com/cgs-earth/pygeoapi-plugins/blob/dc7fb55a2126c6a178c71e2097f1066ba35fd12b/pygeoapi_plugins/process/intersect.py#L192-L193

Is there a way to do this without consistently making two queries on a provider?  Is there a naive way to pass python infinity to providers?

webb-ben:
Interestingly, if we changed https://github.com/geopython/pygeoapi/blob/master/pygeoapi/util.py#L147 to 
```python
        if '.' in value or 'inf' in value:  # float?
```
you could get pygeoapi to cast the configured defaultlimit to pythonic infinity, and it works for at least some of the providers. At the moment this is only possible if you set this as an environment variable, but its not that far off.
```yaml
    limits:
        default_items: ${PYGEOAPI_SERVER_DEAULT_LIMIT:-inf}
```
<img width="1324" alt="image" src="https://github.com/user-attachments/assets/11adab64-8fb1-4884-985d-9877a0d28d12" />

PR #1892, opened 2024-12-31
enhance support for limits (RFC5) (#1856)

tomkralidis:
# Overview
This PR implements [RFC5](https://pygeoapi.io/development/rfc/5) to provide enhanced limit handling implementation in support of better definition and control on data query limiting.
# Related Issue / discussion
#1856 

# Additional information
This PR needs [PSC approval](https://lists.osgeo.org/pipermail/pygeoapi/2025-January/000420.html) of [RFC5](https://pygeoapi.io/development/rfc/5) in order to be merged.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -25,7 +25,7 @@ jobs:
       contents: read
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@master
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2.1.0
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,8 +23,8 @@ jobs:
         include:
           - python-version: '3.10'
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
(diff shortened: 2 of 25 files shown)


totycro (approved):
The implementation looks good to me!

With this change, the old `limit` setting is now ignored, so if anybody updates to the new version without changing the config, then the default limits apply again. We could detect this old setting and print an informational message .

webb-ben (approved):
+1

doublebyte1:
I think this is a really nice improvement! 👍🏽 

Should the pygeoapi example configuration file also be changed?

https://github.com/geopython/pygeoapi/blob/b4d3535032e04b6ca44797d011c9f71239a8bd8e/pygeoapi-config.yml#L44

I would suggest changing it to a different number on default and max:

    limits:
        defaultitems: 10  # applies to vector data
        maxitems: 100  # applies to vector data
        maxdistance: [x, y]  # applies to all datasets
        units: m  # as per UCUM https://ucum.org/ucum#section-Tables-of-Terminal-Symbols
        on_exceed: error  # one of error, throttle

PR #1858, opened 2024-11-20
support dataset specific templates (#1857)

tomkralidis:
# Overview
Adds support for dataset specific templates.
# Related Issue / discussion
Fixes #1857

# Additional information

# Dependency policy (RFC2)

- [x] I have ensured that this PR meets [RFC2](https://pygeoapi.io/development/rfc/2) requirements

# Updates to public demo

- [x] I have ensured that breaking changes to the [pygeoapi master demo server](https://github.com/geopython/demo.pygeoapi.io) have been addressed
  - [x] https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_master/local.config.yml

# Contributions and licensing

(as per https://github.com/geopython/pygeoapi/blob/master/CONTRIBUTING.md#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

PR #1927, opened 2025-02-06
Support GeoJSON-LD templating

webb-ben:
# Overview
Building on the capacity introduced in https://github.com/geopython/pygeoapi/pull/868 and https://github.com/geopython/pygeoapi/pull/1858, the PR migrates GeoJSON-LD rendering to always use Jinja2 templates. The default templates provide the same functional JSON-LD output as pygeoapi currently outputs. With the addition of dataset level templating, custom JSON-LD can be rendered for an GeoJSON Feature or GeoJSON Feature Collection by implementing the appropriate template.

This allows us to also un-next the `context` block from a resource configuration, seeeing as the only specific configuration for linked-data is now just the context block. 

Finally, serialization of geojsonld was moved inside geojson2gejsonld to reduce the number of times the json is serialized and the align with the function's docstring. 

# Related Issue / discussion

# Additional information
In theory it is possible this type of behavior could be subsumed by a jinj2 formatter plugin - not sure how plugin formatters interact with the default HTTP types pygeoapi deals with (HTML, JSON, JSON-LD)

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docker/default.config.yml
+++ b/docker/default.config.yml
@@ -100,12 +100,11 @@ resources:
         keywords:
             - observations
             - monitoring
-        linked-data:
-            context:
-                - datetime: https://schema.org/DateTime
-                - vocab: https://example.com/vocab#
-                  stn_id: "vocab:stn_id"
-                  value: "vocab:value"
+        context:
+            - datetime: https://schema.org/DateTime
+            - vocab: https://example.com/vocab#
+              stn_id: "vocab:stn_id"
+              value: "vocab:value"
         links:
             - type: text/csv
               rel: canonical
--- a/docker/default.config.yml
+++ b/docker/default.config.yml
@@ -100,11 +100,12 @@ resources:
         keywords:
             - observations
             - monitoring
-        context:
-            - datetime: https://schema.org/DateTime
-            - vocab: https://example.com/vocab#
-              stn_id: "vocab:stn_id"
-              value: "vocab:value"
+        linked-data:
+            context:
+                - datetime: https://schema.org/DateTime
+                - vocab: https://example.com/vocab#
+                  stn_id: "vocab:stn_id"
+                  value: "vocab:value"
         links:
             - type: text/csv
               rel: canonical
--- a/pygeoapi/linked_data.py
+++ b/pygeoapi/linked_data.py
@@ -226,14 +226,14 @@ def geojson2jsonld(cls, data: dict, dataset: str,
                 identifier_ = f"{ds_url}/items/{feature['id']}"  # noqa
 
             # Include multiple geometry encodings
-            if (feature.get('geometry') is not None):
+            if feature.get('geometry') is not None:
                 jsonldify_geometry(feature)
 
             data['features'][i] = {
                 '@id': identifier_,
                 'type': 'schema:Place',
                 **feature.pop('properties'),
-                **feature,
+                **feature
             }
 
     if data.get('timeStamp', False):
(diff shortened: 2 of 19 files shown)


totycro:
I would suggest to add at least a basic test for this to make sure that it's not broken by accident in the future.

webb-ben:
> I would suggest to add at least a basic test for this to make sure that it's not broken by accident in the future.

Added some general tests for `linked_data.py`. Not sure if it is worth pulling in `test_describe_collections_json_ld` and `test_root_structured_data`

pygeoapi/linked_data.py
tomkralidis, line 1: Pending #1943, then:

```python
tpl_config = get_dataset_templates(dataset)
content = render_j2_template(cls.config, tpl_config, 'collections/items/index.jsonld', data)
```

pygeoapi/linked_data.py
```diff
@@ -233,9 +228,15 @@ def geojson2jsonld(cls, data: dict, dataset: str,
...
+            # Include multiple geometry encodings
+            if (feature.get('geometry') is not None):
+                jsonldify_geometry(feature)
+
             data['features'][i] = {
                 '@id': identifier_,
-                'type': 'schema:Place'
+                'type': 'schema:Place',
+                **feature.pop('properties'),
+                **feature,
```
tomkralidis, line 239: remove trailing comma.

pygeoapi/linked_data.py
```diff
@@ -233,9 +228,15 @@ def geojson2jsonld(cls, data: dict, dataset: str,
             if not is_url(str(identifier_)):
                 identifier_ = f"{ds_url}/items/{feature['id']}"  # noqa
 
+            # Include multiple geometry encodings
+            if (feature.get('geometry') is not None):
```
tomkralidis, line 232: `if feature.get('geometry') is not None:`

tomkralidis (changes requested):
Change requests to be applied pending #1943

tests/test_linked_data.py
```diff
@@ -0,0 +1,174 @@
...
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+# HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+# WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# =================================================================
+
+
+from copy import deepcopy
```
tomkralidis, line 31:
```python
from copy import deepcopy
import json
import logging

import pytest
from shapely.geometry import (Point, MultiPoint, Polygon,
                              MultiPolygon, LineString, MultiLineString)

from pygeoapi.linked_data import (
    geojson2jsonld,
    geom2schemageo,
    jsonldify_geometry
)
```

docker/default.config.yml
```diff
@@ -100,12 +100,11 @@ resources:
         keywords:
             - observations
             - monitoring
-        linked-data:
```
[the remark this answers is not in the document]
webb-ben: With the removal of item_template from linked data, the only possible key in linked-data is context. No reason to double nest.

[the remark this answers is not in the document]
tomkralidis: I would keep the nesting for 1) existing applications 2) any other linked data type definitions that may be required over time.

tomkralidis (approved):

Issue #1930, opened 2025-02-07
Add support to tileindexing in MapScript Provider


doublebyte1:
[Tileindexes](https://mapserver.org/optimization/tileindex.html) are MapServer’s method for doing on-the-fly mosaicing. They are supported in MapScript, and it would be nice to support them also in the MapScript provider.

**Describe the solution you'd like**
I would like to support tileindexes, by introducing a simple boolean parameter `tileindex`.

```
        providers:
          - type: map
            name: MapScript
            data: /data/orto_index_all.shp
            options:
                type: MS_LAYER_RASTER
                tileindex: True
                layer: sample
            format:
                name: png
                mimetype: image/png

```

This parameter should be optional, and default to False (e.g.: no tileindex).

**Describe alternatives you've considered**
The alternative could be remain without tile indexing, and serve different files as different layers. However, that approach is less efficient than having indexing.

**Additional context**
Tile indexes require an index file, that can be generated with [gdalindex](https://gdal.org/en/stable/programs/gdaltindex.html).

CeciliaPH:
Dear all,

Joyfully I found that this issue is solved. We will be using it for sure in our maps soon.
I wonder if you have thought about adding this provider to coverages, as it would be really useful for that API also. 
We suggested this in this staled issue.
https://github.com/geopython/pygeoapi/issues/1251

Greetings,

Cecilia

CeciliaPH:
Hi! I mean to add this ability of using tileindex and MapScript as a source for the OGC-API Coverages collections.
It will be very usefull as we have our DTM data split in different files.
May be something like this:

        providers:
            - type: coverage
              name: MapScript
            data: /data/dtm_index_all.shp
            options:
                type: MS_LAYER_RASTER
                tileindex: True
                layer: sample
            format:
                name: png
                mimetype: image/png

tomkralidis:
@CeciliaPH thanks; opened a new issue in #1939

PR #1931, opened 2025-02-07
Added support to tileindexing on MapScript provider

doublebyte1:
# Overview

This PR adds the ability to create tileindexes with the MapScript provider.  For enabling tile indexing, the provider options need to include an index file (which can be generated with gdalindex) and a `tileindex` parameter that should be set to True. This parameter is completely optional, and the default value is `False` (e.g.: no tileindex). This change is backwards compatible.

```
        providers:
          - type: map
            name: MapScript
            data: /data/orto_index_all.shp
            options:
                type: MS_LAYER_RASTER
                projection: 3763
                tileindex: True
                debug: MS_DEBUGLEVEL_VVV
                layer: sample
            format:
                name: png
                mimetype: image/png

```
# Related Issue / discussion

https://github.com/geopython/pygeoapi/issues/1930

# Additional information

Example of a raster mosaic generated in pygeoapi, based on the index file:

Screenshot from 2025-02-07 19-43-47

Screenshot from 2025-02-07 19-45-22

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/data-publishing/ogcapi-maps.rst
+++ b/docs/source/data-publishing/ogcapi-maps.rst
@@ -73,9 +73,9 @@ Projections are supported through EPSG codes (`options.projection`):
 
 This parameter is optional, defaulting to WGS84 (4236).
 
-This provider also supports `tileindexing <https://mapserver.org/optimization/tileindex.html>`_,
-which let MapScript create a mosaic on the fly, piecing together a set of files. 
-In order to enable it, set `options.tileindex` to True and set the location of the index file on the `data` path.
+This provider also supports `tile indexing <https://mapserver.org/optimization/tileindex.html>`_,
+which lets MapScript create a mosaic on the fly, piecing together a set of files. 
+In order to enable it, set `options.tileindex` to `True` and set the location of the index file on the `data` path.
 
 .. code-block:: yaml
 
@@ -91,7 +91,7 @@ In order to enable it, set `options.tileindex` to True and set the location of t
                 name: png
                 mimetype: image/png
 
-The `options.tileindex` parameter is optional, defaulting to False.
+The `options.tileindex` parameter is optional, defaulting to `False`.
 
 WMSFacade
 ^^^^^^^^^
--- a/docs/source/data-publishing/ogcapi-maps.rst
+++ b/docs/source/data-publishing/ogcapi-maps.rst
@@ -73,6 +73,26 @@ Projections are supported through EPSG codes (`options.projection`):
 
 This parameter is optional, defaulting to WGS84 (4236).
 
+This provider also supports `tileindexing <https://mapserver.org/optimization/tileindex.html>`_,
+which let MapScript create a mosaic on the fly, piecing together a set of files. 
+In order to enable it, set `options.tileindex` to True and set the location of the index file on the `data` path.
+
+.. code-block:: yaml
+
+        providers:
+          - type: map
+            name: MapScript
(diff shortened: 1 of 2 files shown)


pygeoapi/provider/mapscript_.py
```diff
@@ -74,13 +74,22 @@ def __init__(self, provider_def):
 
             file_extension = self.data.split('.')[-1]
 
-            if file_extension in ['shp', 'tif']:
-                LOGGER.debug('Setting built-in MapServer driver')
-                self._layer.data = self.data
+            tileindex = False
```
tomkralidis, line 77: Suggest using `pygeoapi.util.str2bool`:

```python
            tileindex = False
            if 'tileindex' in self.options:
                if str2bool(self.options.get('tileindex', False))
                    self._layer.tileindex = self.data 
```

docs/source/data-publishing/ogcapi-maps.rst
```diff
@@ -73,6 +73,26 @@ Projections are supported through EPSG codes (`options.projection`):
 
 This parameter is optional, defaulting to WGS84 (4236).
 
+This provider also supports `tileindexing <https://mapserver.org/optimization/tileindex.html>`_,
```
tomkralidis, line 76: tileindexing -> tile indexing

docs/source/data-publishing/ogcapi-maps.rst
```diff
@@ -73,6 +73,26 @@ Projections are supported through EPSG codes (`options.projection`):
...
+which let MapScript create a mosaic on the fly, piecing together a set of files.
```
tomkralidis, line 77: let -> lets

docs/source/data-publishing/ogcapi-maps.rst
```diff
@@ -73,6 +73,26 @@ Projections are supported through EPSG codes (`options.projection`):
 
 This parameter is optional, defaulting to WGS84 (4236).
 
+This provider also supports `tileindexing <https://mapserver.org/optimization/tileindex.html>`_,
+which let MapScript create a mosaic on the fly, piecing together a set of files. 
+In order to enable it, set `options.tileindex` to True and set the location of the index file on the `data` path.
```
tomkralidis, line 78: True -> `True`

docs/source/data-publishing/ogcapi-maps.rst
```diff
@@ -73,6 +73,26 @@ Projections are supported through EPSG codes (`options.projection`):
...
+            data: /data/index.shp
+            options:
+                type: MS_LAYER_RASTER
+                tileindex: True
+                layer: index
+            format:
+                name: png
+                mimetype: image/png
+
+The `options.tileindex` parameter is optional, defaulting to False.
```
tomkralidis, line 94: False -> `False`

tomkralidis (changes requested):

pygeoapi/provider/mapscript_.py
doublebyte1: It is fine to use str2bool, but if we don't combine the two conditions, collections that have `tileindex=False` will never enter the `else` code, and therefore won't have a layer assigned. I propose this:

https://github.com/doublebyte1/pygeoapi/blob/17323a9e98633149d207b10deaa5870b8830399e/pygeoapi/provider/mapscript_.py#L77-80

tomkralidis (approved):

Issue #1936, opened 2025-02-12
pygeoapi fails to gracefully catch missing tiles


webb-ben:
**Description**
Per the [OGC API - TIles](https://docs.ogc.org/is/20-057/20-057.html#:~:text=A.1.2.6.%20%20Error%20conditions) spec, tiles that are not found are supposed to return a 400 or 404 status. pygeoapi does this, but at the cost of logging a full error stack for every tile requested that does not exist.

**Steps to Reproduce**
1. Run pygeoapi with a Tile provider and log level set to error.
2. View tile map, in and out of where there map be tiles.
3. View pygeoapi logs

**Expected behavior**
pygeoapi is able to return the proper HTTP status without spamming the logs for an expected behavior. 

**Screenshots/Tracebacks**
500+ errors from last 15 minutes
<img width="1508" alt="Image" src="https://github.com/user-attachments/assets/4252ae14-f843-4071-a2fb-094a0dcc5274" />

**Environment**
 - OS: all
 - Python version: 3.10
 - pygeoapi version: 0.20.dev0

**Additional context**
pygeoapi tile providers could consistently throw ``ProviderTileNotFoundError`` which can be handled in `api/tiles.py` to gracefully return such an error. Not sure it is wise to rely on `ProviderTileNotFoundError`.

PR #1937, opened 2025-02-12
Gracefully catch tile not found

webb-ben:
# Overview
This PR introduces graceful catching of `Not Found` errors for a Tile provider. When a Tile provider raises `ProviderTileNotFoundError` the API returns a 404 gracefully.

# Related Issue / discussion
https://github.com/geopython/pygeoapi/issues/1936

# Additional information
Willing to adapt the solution - but if the expected behavior is to return a 404, I think it does make sense to not log such behavior as an error.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


doublebyte1:
Thanks for this PR @webb-ben ! +:1

PR #1940, opened 2025-02-18
Fix pygeofilter reprojection tests

webb-ben:
# Overview
Not sure what changed, but [Build](https://github.com/geopython/pygeoapi/actions/workflows/main.yml) is now failing. For some reason the output geomtery has changed ever so slightly. This PR adjusts the expected geometry in the tests to correspond with their actual values. @MTachon wonder if you might have insights into what changed and what is the best long term test case.

# Related Issue / discussion

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -496,7 +496,7 @@ def test_prefetcher():
         None,
         pygeofilter.ast.GeometryIntersects(
             pygeofilter.ast.Attribute(name='geometry'),
-            Geometry({'type': 'Point', 'coordinates': (2313681.8086284213, 4641307.939955418)})  # noqa
+            Geometry({'type': 'Point', 'coordinates': (2313681.808628421, 4641307.939955416), 'crs': {'properties': {'name': 'urn:ogc:def:crs:EPSG::3004'}}}) # noqa
         ),
         id='unnested-geometry-transformed-coords-explicit-input-crs-ewkt'
     ),
@@ -507,7 +507,7 @@ def test_prefetcher():
         None,
         pygeofilter.ast.GeometryIntersects(
             pygeofilter.ast.Attribute(name='geometry'),
-            Geometry({'type': 'Point', 'coordinates': (2313681.8086284213, 4641307.939955418)})  # noqa
+            Geometry({'type': 'Point', 'coordinates': (2313681.808628421, 4641307.939955416), 'crs': {'properties': {'name': 'urn:ogc:def:crs:EPSG::3004'}}}) # noqa
         ),
         id='unnested-geometry-transformed-coords-explicit-input-crs-filter-crs'
     ),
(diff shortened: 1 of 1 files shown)


MTachon:
Hi @webb-ben, the error is so small that it looked like rounding errors to me at first. But why the transformed coordinates suddenly were different, ever so slightly, from what they used to be, was strange!

There are no type casting from double precision floating point numbers to single precision floating point by going back and forth between the *geojson* representation of a geometry, and its shapely geometrical object representation (with `shapely.geometry.shape` imported as `geojson_to_geom`, and `shapely.geometry.mapping` imported as `geom_to_geojson`).

What I saw that did change between the builds is the version of `pyproj` used. The version used before the builds started to fail was 3.7.0. When the builds started to fail, `pyproj`'s version was upgraded to 3.7.1. Looking at the [changelog of the pyproj package](https://pyproj4.github.io/pyproj/stable/history.html), the wheel for version 3.7.1 now uses PROJ 9.5.1, while the wheel for version 3.7.0 uses PROJ 9.4.1. I am not sure what changed between PROJ 9.4.1 and PROJ 9.5.1, but pinning pyproj to version 3.7.0 help as you can see in the successful build [here](https://github.com/MTachon/pygeoapi/actions/runs/13453971493/job/37593789572).

A solution to this issue would be to pin the version of `pyproj` in the `requirements.txt` file as I just did in the test above. But that is unfortunate if we want to get updates from the `pyproj`/`PROJ` library. We can update the tests as you just did whenever an upgrade of `pyproj` results in similar problems later. Or we could use a lower precision for the coordinates of the test geometries (e.g. rounding them to nearest millimeter for CRSs with meter as unit before comparing result and expected geometries). I think the last solution would be best, as one never needs to use as many decimals as in the examples in the real world, no sensor or mapping method has that kind of accuracy etc. We could refactor the tests by removing some decimals in the geometries' coordinates passed to `pytest.param()`, and rounding the coordinates of the geometries that have been transformed by `_inplace_transform_filter_geometries()`, before the test comparison (` assert result == expected`).

totycro:
I'm not an expert on reprojection but rather on tests, and from that perspective i would say that the test should pass if the output is as desired, and for calculation with floating point numbers, this can easily be a range.

Many small factors can determine the precise output, just reordering operations in a mathematically correct way can change them. Also the tests should run locally for everyone, so they should be robust enough for different setups.

So ideally the tests could be rewritten to pass if the numbers are close enough.
python has [`math.isclose`](https://docs.python.org/3/library/math.html#math.isclose) to deal with this, and there might even be more convenient functions that deal with geometries.

MTachon:
One could isolate the test for the geometry. Converting the GeoJSON geometry to a *Shapely* geometrical object would allow using the `.equals_exact()` method (with a specified [tolerance](https://shapely.readthedocs.io/en/2.0.6/reference/shapely.Polygon.html#shapely.Polygon.equals_exact)) to compare the geometries' coordinates. May be a good idea to [normalize](https://shapely.readthedocs.io/en/2.0.6/reference/shapely.Polygon.html#shapely.Polygon.normalize) the geometries before comparing them.

doublebyte1 (approved):
Thank You for this fix, @webb-ben!

Issue #1948, opened 2025-02-28
pygeoapi does not respect header q values


webb-ben:
**Description**
When sending a request to pygeoapi with multiple values in accept parameter, pygeoapi takes the first item instead of evaluating the q values of all options sent in the header. 

**Steps to Reproduce**
1. Send a request to pygeoapi with variable q values, and a lower q value as your first header:
`Accept: application/json;q=0.5,application/ld+json;q=1`
2. View response content format

**Expected behavior**
pygeoapi evaluates the full header context to decide the most appropriate header to send in the response. 

**Screenshots/Tracebacks**

**Environment**
 - OS: linux
 - Python version: 3.10/3.12
 - pygeoapi version: 0.20.dev0

**Additional context**

tomkralidis:
Untested:

```diff
diff --git a/pygeoapi/api/__init__.py b/pygeoapi/api/__init__.py
index 7ce3ec4..854d667 100644
--- a/pygeoapi/api/__init__.py
+++ b/pygeoapi/api/__init__.py
@@ -339,7 +339,7 @@ class APIRequest:
         h = headers.get('accept', headers.get('Accept', '')).strip() # noqa
         (fmts, mimes) = zip(*FORMAT_TYPES.items())
         # basic support for complex types (i.e. with "q=0.x")
-        for type_ in (t.split(';')[0].strip() for t in h.split(',') if t):
+        for type_ in sorted(h.split(','), key=lambda x: x.split('q=')[-1], reverse=True):  # noqa
             if type_ in mimes:
                 idx_ = mimes.index(type_)
                 format_ = fmts[idx_]
```

webb-ben:
Its worth noting we technically allow the type of preference setting for the following headers `Accept`, `Content-Encoding`, `Accept-Language`. I am in favor on consolidating the logic for extracting out the choice of header to a common util function:

```python
def get_choice_from_headers(headers: dict, 
                            header_name: str,
                            all: bool = False) -> Any:
    """
    Gets choices from a request dictionary,
    considering numerical ordering of preferences.
    Supported are complex preference strings (e.g. "fr-CH, fr;q=0.9, en;q=0.8")

    :param headers: `dict` of request headers.
    :param header_name: Name of request header.
    :param all: bool to return one or all header values.

    :returns: Sorted choices from header
    """
    try:
        # Clean headers and select header of interest
        cleaned_headers = {k.lower(): v for k, v in headers.items()}
[...]

Issue #1591, opened 2024-03-11
gzip compression not applied in starlette (invalid header handling)


totycro:
**Description**
When using `pygeoapi --starlette` and adding the request header `Accept-Encoding: gzip`, the response is not actually gzipped.

The problem lies in these two lines:
https://github.com/geopython/pygeoapi/blob/master/pygeoapi/api.py#L640  
https://github.com/geopython/pygeoapi/blob/master/pygeoapi/api.py#L625

In starlette, header names are case insensitive (as by HTTP spec). So you can do `accept-encoding in request.headers` or `Accept-Encoding in request.headers` and both will return `True`. However in `get_request_headers`, the headers are listed, and in this operation the names are lower cased.

Thus in `get_response_headers`, `Accept-Encoding` can never actually be present in the headers dict, only `accept-encoding`.

Note that this bug probably doesn't cause more problems since the request headers only seem to be used for gzip here. 

Here's the starlette documentation:
https://www.starlette.io/requests/#headers

A fix could be to lowercase all request headers in `get_request_headers` 

**Expected behavior**
The response should be gzipped.

**Screenshots/Tracebacks**
An example set of request headers as printed in starlette:
```json
{'host': 'localhost:5000', 'user-agent': 'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:122.0) Gecko/20100101 Firefox/122.0', 'accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8', 'accept-language': 'en-US,en;q=0.5', 'accept-encoding': 'gzip', 'connection': 'keep-alive', 'cookie': '...;, 'upgrade-insecure-requests': '1', 'sec-fetch-dest': 'document', 'sec-fetch-mode': 'navigate', 'sec-fetch-site': 'cross-site', 'pragma': 'no-cache', 'cache-control': 'no-cache'}
```

**Environment**
 - OS:
 - Python version: 
 - pygeoapi version: 0.16 (from git)

PR #1952, opened 2025-03-05
Implement q priority sorting of requests

webb-ben:
# Overview
Create common util functions `get_choice_from_headers` and `get_from_headers` to get value from header in case insensitive way and implement q priority sorting for header values. This logic is then applied to all headers that pygeoapi accepts q value ranking for: `Accept` and `Accept-Lang`.

# Related Issue / discussion
Addresses: https://github.com/geopython/pygeoapi/issues/1948 and https://github.com/geopython/pygeoapi/issues/1591

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


webb-ben:
@GeoSander I moved the header parsing and q priority sorting to common util functions. Could you confirm everything looks good? I added additional tests to expand validate nothing broke in the translations while supporting the same seeking in a more concise and readable sort of acceptable values.

pygeoapi/l10n.py
```diff
@@ -131,12 +131,12 @@ def best_match(accept_languages, available_locales) -> Locale:
...
 
-    :param accept_languages: A Locale or string with one or more languages.
+    :param accept_languages: A Locale or list of one or more languages.
                              This can be as simple as "de" for example,
                              but it's also possible to include a territory
                              (e.g. "en-US" or "fr_BE") or even a complex
-                             string with quality values, e.g.
-                             "fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5".
+                             list sorted by quality values, e.g.
+                             ["fr-CH, "fr", "en", "de", "*"].
```
GeoSander:
```suggestion
                             ["fr-CH", "fr", "en", "de", "*"].
```

Not important, but missing a quote there.

GeoSander (approved):
Hi @webb-ben,

Thanks for case-insensitive header value retrieval, and for adding the prioritization logic to the `q` parameter. That was one of those things that we did not fully implement back then, because we were running out of time/budget.

I haven't tested this tbh, but the code looks good and the tests make sense 👍 

Sander

PR #1982, opened 2025-03-27
Don't generate paging links if provider created them

### LINKED ISSUE #1981: Let providers handle paging links to enable other paging methods

We want to move away from limit-offset paging for the data we're serving from Postgres, as deep pages (requiring large offsets) are inefficient and can result in users downloading duplicate records and missing new records if data is appended to an earlier page. We're hoping to move to keyset paging instead, which means users will provide a (non-numeric) key with their queries to get the next page.

Unfortunately limit-offset paging is baked into the itemtypes API here:
https://github.com/geopython/pygeoapi/blob/c976bd7f61d60a50536c84a2663e506328749789/pygeoapi/api/itemtypes.py#L548-L574

Rather than assuming providers are using limit-offset paging, and therefore the last page was always at `offset - limit` and the next one at `offset + limit`, it'd be nice if providers could pass their own `next` links that would skip this entire chunk of code. I think a `next` link in the response should disable the generation of both `prev` and `next` links, as the Features standard mentions that `prev` links may be skipped based on implementation:

> Providing prev links supports navigating back and forth between pages, but depending on the implementation approach it may be too complex to implement.

Potentially each provider should be responsible for implementing a sensible default pagination strategy, but that seems like a much higher lift than assuming (1) the provider is right if it provides a `next` link (2) the existing strategy can be used otherwise.

mikemahoney218-usgs:
# Overview

This PR changes itemtypes.py to allow providers to generate their own `'next'` links, which then disable the limit-offset paging built into pygeoapi. As discussed in #1981, we want to move away from limit-offset paging for the data we're serving from Postgres and move towards keyset paging instead, and I believe this change would let us handle that from within our provider.

We just check for a next link in the response before disabling the generation of both prev and next links, as the Features standard mentions that prev links may be skipped based on implementation. As such, if the provider has generated a `next` link but not a `prev` link, we assume that was intentional.

# Related Issue / discussion

Closes #1981 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api/itemtypes.py
+++ b/pygeoapi/api/itemtypes.py
@@ -545,34 +545,40 @@ def get_collection_items(
         'href': f'{uri}?f={F_HTML}{serialized_query_params}'
     }])
 
-    if 'next' not in [link['rel'] for link in content['links']]:
-        if offset > 0:
-            prev = max(0, offset - limit)
-            content['links'].append(
-                {
-                    'type': 'application/geo+json',
-                    'rel': 'prev',
-                    'title': l10n.translate('Items (prev)', request.locale),
-                    'href': f'{uri}?offset={prev}{serialized_query_params}'
-                })
+    next_link = False
+    prev_link = False
 
-        next_link = False
-
+    if 'next' in [link['rel'] for link in content['links']]:
+        LOGGER.debug('Using next link from provider')
+    else:
         if content.get('numberMatched', -1) > (limit + offset):
             next_link = True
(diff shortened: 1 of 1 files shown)


mikemahoney218-usgs:
I wanted to bump this to see if there's any interest in the PR! We're trying to determine now if we'll need to post-process the response object somehow, or if this change will let us handle things entirely in the provider.

webb-ben (approved):

PR #1986, opened 2025-03-28
Clean feature properties regarding resource setting for OGR provider

vvmruder:
# Overview

Adds a clean out of properties on returned features for the OGR provider similar to the approach from Postgres provider. This way the configured properties as they can be defined in the [docs](https://docs.pygeoapi.io/en/0.19.0/data-publishing/ogcapi-features.html#controlling-the-order-of-properties) are respected.

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute this PR to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/pygeoapi/provider/ogr.py
+++ b/pygeoapi/provider/ogr.py
@@ -30,12 +30,12 @@
 #
 # =================================================================
 
+from copy import deepcopy
 import functools
 import importlib
 import logging
 import os
 from typing import Any
-from copy import deepcopy
 
 from osgeo import gdal as osgeo_gdal
 from osgeo import ogr as osgeo_ogr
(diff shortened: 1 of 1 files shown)


pygeoapi/provider/ogr.py
```diff
@@ -35,6 +35,7 @@
 import logging
 import os
 from typing import Any
+from copy import deepcopy
```
tomkralidis, line 38: Alphabetize, so:

```python
from copy import deepcopy
import functools
import importlib
import logging
import os
from typing import Any
```

tomkralidis (changes requested):

Issue #1985, opened 2025-03-28
Potentially using wrong tile URL when fetching vector tiles from MVT providers


ThorodanBrom:
**Description**
Hello, as part of the OGC March Code Sprint, we noticed an error in the `tileRow` and `tileColumn` values when the tile URL is created for the vector tiles preview on Leaflet. `tileRow` was replaced with `x` and `tileColumn` with `y` - this is opposite to what's expected - `tileRow` should be `y`, `tileColumn` should be `x`. 

https://github.com/geopython/pygeoapi/blob/8fc8ba3a2d65b1f44f6e9b23b645f86fcfa9bccd/pygeoapi/templates/collections/tiles/index.html#L103-L104

**Despite the switched indices, the preview still worked as intended and this was an open question.**

### Probable cause

(**x** and **y** here means the actual value of the x and y indices respectively)

- The indices are switched at the preview URL, so `/{tileMatrix}/{tileRow}/{tileColumn}` becomes /{z}/{**x**}/{**y**}, 

- The `get_tile` methods in the `mvt_*` classes become get_tile(z, **x**, **y**)

- The final call to the providers becomes /z/**x**/**y**

One thing to note is that the [Elastic MVT Provider](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-vector-tile-api.html), Tippecanoe folder structure, [pg_tileserv](https://github.com/CrunchyData/pg_tileserv/?tab=readme-ov-file#table-layer-detail-json) and [martin](https://maplibre.org/martin/using-with-leaflet.html) **all serve tiles in the /z/x/y** URL/dir-structure format by default. 

**However, the `get_tile` methods for all the `mvt_*` providers sets the URL as `/{z}/{y}/{x}`** ([see here](https://github.com/search?q=repo%3Ageopython%2Fpygeoapi+path%3A%2F%5Epygeoapi%5C%2Fprovider%5C%2F%2F++%2F%7Bz%7D%2F%7By%7D%2F%7Bx%7D&type=code))

Hence (unless the y and x indices are switched around somewhere else in the code) I think that the reason the preview is working is because the switched indices actually get the correct tiles from all the MVT providers. 

But we noticed that:
[...]

PR #1989, opened 2025-04-05
Correct MVT tile indices when fetching tiles from providers + fix tile indices in MVT UI preview

ThorodanBrom:
# Overview

This PR fixes the tile indices in the Elastic, Proxy and Tippecanoe MVT providers, and corrects the tile indices in the MVT tiles pygeoapi UI. All providers now use `/{z}/{x}/{y}` to fetch tiles from their original sources. 

A combination of the switched-around indices in the preview and the switched-around indices in the providers allowed tiles to be visible on the preview, but external clients needed to use `/{z}/{x}/{y}` in their tile URL to serve tiles - this is contrary to the OGC Tiles specification which is `/{z}/{y}/{x}`.

# Related Issue / discussion

This PR is related to #1985 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/mvt_elastic.py
+++ b/pygeoapi/provider/mvt_elastic.py
@@ -176,7 +176,7 @@ class MVTElasticProvider(BaseMVTProvider):
                 with requests.Session() as session:
                     data = {'fields': ['*']}
                     session.get(base_url)
-                    resp = session.get(f'{base_url}/{layer}/{z}/{y}/{x}{url_query}', json=data)  # noqa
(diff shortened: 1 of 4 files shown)


doublebyte1 (approved):
Thanks for this PR, @ThorodanBrom !

No need to do changes in the tests and the documentation.

If you want to create a PR to fix the client indexes on the dive into pygeoapi workshop, that would be very much appreciated: https://github.com/geopython/diving-into-pygeoapi/issues/192 

Thank You! 🙏🏽

doublebyte1:
@ThorodanBrom please also check the box on "Updates to public demo"

PR #1991, opened 2025-04-10
Fix response type for tile request

joepke:
# Overview

Set content-type to mimetype instead of format_type, so e.g. `image/png` instead of `png`.

# Related Issue / discussion

No related issue.

# Additional information

When testing a get tile request for either vector tiles or WMTS, the content-type in the response is either `pbf` or `png`. This should be replaced by the relevant mimetype, so the reponse type can be detected properly and previewed in case of png.

Example of request:

https://demo.pygeoapi.io/master/collections/lakes/tiles/WebMercatorQuad/1/1/1?f=pbf

returns response header with `content-type: pbf`.

A tile request to a WMTSFacade provider:

localhost:5000/collections/streets/tiles/WebMercatorQuad/3/2/4?f=png

returns a response header with `content-type: png`. 

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api/tiles.py
+++ b/pygeoapi/api/tiles.py
@@ -238,7 +238,7 @@ def get_collection_tiles_data(
         p = load_plugin('provider', t)
 
         format_ = p.format_type
-        headers['Content-Type'] = format_
+        headers['Content-Type'] = t['format']['mimetype']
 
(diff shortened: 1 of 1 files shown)


doublebyte1 (approved):
Thanks @joepke . This makes sense!

PR #2004, opened 2025-04-24
MySQL OAF provider

C-Loftus:
# Overview

- Adds a MySQL OAF provider with associated tests 
    - MySQL is very similar to postgis but doesn't support the same geospatial functions so it needs some different underlying behavior
- This PR keeps all existing postgres functionality; in doing so it puts all core sql functionality into a base class that both postgres and mysql providers can inherit from; (this could be used for other sql providers in the future if desired)
- Fixed formatting in existing postgis provider to follow the rest of the project (single quotes etc)

# Related Issue / discussion

N/A

# Additional information

N/A

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


pygeoapi/provider/sql.py
```diff
@@ -193,15 +231,17 @@ def query(self, offset=0, limit=10, resulttype='results',
                 'type': 'FeatureCollection',
                 'features': [],
                 'numberMatched': matched,
-                'numberReturned': 0
+                'numberReturned': 0,
```
webb-ben, line 234:
```suggestion
                'numberReturned': 0

File: pygeoapi/provider/sql.py [hunk 2]
```diff
@@ -193,15 +231,17 @@ def query(self, offset=0, limit=10, resulttype='results',
...
-            if resulttype == "hits" or not results:
+            if resulttype == 'hits' or not results:
                 return response
 
             crs_transform_out = self._get_crs_transform(crs_transform_spec)
 
-            for item in results.order_by(*order_by_clauses).offset(offset).limit(limit):  # noqa
+            for item in (
+                results.order_by(*order_by_clauses).offset(offset).limit(limit)
+            ):  # noqa
```
webb-ben, line 244:
```suggestion
            ):

File: pygeoapi/provider/sql.py [hunk 3]
```diff
@@ -269,7 +309,9 @@ def _column_format_to_json_schema_format(column_type):
 
                 self._fields[str(column.name)] = {
                     'type': _column_type_to_json_schema_type(column.type),
-                    'format': _column_format_to_json_schema_format(column.type)
+                    'format': _column_format_to_json_schema_format(
+                        column.type
+                    ),
```
webb-ben, line 314:
```suggestion
                    )
```

pygeoapi/provider/sql.py
```diff
@@ -550,40 +602,38 @@ def _get_crs_transform(self, crs_transform_spec=None):
...
-        user: str,
-        password: str,
-        **connection_options
+    driver_name: str,
+    host: str,
+    port: str,
+    database: str,
+    user: str,
+    password: str,
+    **connection_options,
```
webb-ben, line 611:
```suggestion
    **connection_options
```

pygeoapi/provider/sql.py
```diff
@@ -123,29 +148,39 @@ def __init__(self, provider_def):
...
         self._store_db_parameters(provider_def['data'], options)
         self._engine = get_engine(
+            driver_name,
             self.db_host,
             self.db_port,
             self.db_name,
             self.db_user,
             self._db_password,
-            **(self.db_options or {})
+            **{**(self.db_options or {}), **(extra_conn_args or {})}
```
webb-ben, line 157: What is going on here?

pygeoapi/provider/sql.py
```diff
@@ -550,40 +602,38 @@ def _get_crs_transform(self, crs_transform_spec=None):
...
 @functools.cache
 def get_table_model(
-        table_name: str,
-        id_field: str,
-        db_search_path: tuple[str],
-        engine,
+    table_name: str,
+    id_field: str,
+    db_search_path: tuple[str],
+    engine,
```
webb-ben, line 636:
```suggestion
    engine
```

tests/test_mysql_provider.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def test_fields(config):
+    p = MySQLProvider(config)
+    fields = p.get_fields()
+    expectedFields = [
+        'locationID',
+        'locationName',
+        'description',
+        'created_at',
```
webb-ben, line 115:
```suggestion
        'created_at'
```

pygeoapi/provider/sql.py
C-Loftus: This is the unpacked union of `self.db_options` and `extra_conn_args`.  It expands the `self.db_options` dictionary but then in order to add the extra items, you have to unpack it again after unioning it with `,`. So it is the same resultant data type and structure.

webb-ben:
```suggestion
            **(self.db_options or {}) | (extra_conn_args or {})
```

Issue #1766, opened 2024-07-29
validate pydantic version when testing on Debian


tomkralidis:
As discussed at FOSS4G Europe 2024 with @EdwardBetts and @xamanu (in support of pygeoapi on Debian):

https://hachyderm.io/@edward/112728302761775798

@EdwardBetts @xamanu can you confirm the Debian version?

On jammy (our current baseline):

```bash
dpkg -l | grep -i pydantic
ii  python3-pydantic            1.8.2-1                                 all          Data validation and settings management using Python type hinting
```

And a small script to test/reproduce:

```python
from pydantic import BaseModel, Field

class Foo(BaseModel):
    api_version: str = Field(regex=r'^\d+\.\d+\..+$',
                             description="Semantic API version number.")

f = Foo(api_version='1.0.0')
```

pantierra:
The current version in Debian `sid` is the latest v1 version `1.10.17`, which already has the renaming from `regex` to `pattern` included before v2.

EdwardBetts:
Debian has Pydantic 2.9.2.

Can we change pygeoapi to use the `pattern` parameter instead of `regex` for `Field()`, so it will work with Pydantic 2?

kalxas:
I would wait for pydantic 2.x to hit the next Ubuntu LTS before moving

pantierra:
It feels like a chicken-and-egg situation. If we follow the latest Debian (`sid`) version, `pygeoapi` could be included in the next Ubuntu LTS. But if we wait, it will be delayed by another release, and we may always run into the risk of incompatibilities.

kalxas:
Is the goal to include pygeoapi in Debian on the next stable release?
What is the timeline?

tomkralidis:
Let's discuss at the next pygeoapi [meeting](https://github.com/geopython/pygeoapi/wiki/Meeting-2025-02-28) on 2025-02-28.

tomkralidis:
We discussed this at [today's](https://github.com/geopython/pygeoapi/wiki/Meeting-2025-02-28) pygeoapi meeting.

pygeoapi inclusion in Debian is of value and importance.  We also need to consider the importance of Ubuntu LTS to the community at large.  We also discussed that a safe strategy would be to have version checks in the pygeoapi pydantic codepaths to be able to shim between v1 and v2.

@francbartoli is this possible to update for #1941?  I believe @webb-ben has done some work in this direction.

kalxas:
@francbartoli good point.
[...]

Issue #1664, opened 2024-05-17
Compatibility with Pydantic >= 2 through pydantic.v1


SarahG-579462:
**Is your feature request related to a problem? Please describe.**

Recently, pygeoapi's pydantic was downgraded from 2.0-compatible to pydantic < 1. This removed compatibility with two of our libraries, [xscen](https://github.com/Ouranosinc/xscen) and [xclim](https://github.com/Ouranosinc/xclim), for climate data processing, due to [intake-esm's current requirement on pydantic > 2](https://github.com/intake/intake-esm/blob/60766304f86173d8cad970e8979693e281f80658/requirements.txt#L10)

Because we require these in the use of our pygeoapi process plugins, this change forced us to use either very outdated versions of xscen (and thus xarray, pandas, xclim, etc.), or a very outdated version of pygeoapi.

Thankfully, a simple fix is available by the use of pydantic 2's backwards-compatible pydantic.v1 module. 

It is not entirely clear to me why pydantic was downgraded, the best explanation I can see in the discussion was due to "versionitis". 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Pydantic issues:
#1573 
#1341
#1584

tomkralidis:
Given RFC2, and that our current base OS is Ubuntu jammy, we are bound to OS specific versions.  AFAICT, Pydantic is at [version 1.8.2](https://packages.ubuntu.com/source/jammy/amd64/pydantic) on jammy.

ridoo:
@francbartoli thanks for giving more background! BTW, I got notified that #1547 [got selected](https://github.com/geopython/pygeoapi/issues/1547#issuecomment-2143681422) to be closed by github-actions.

tomkralidis:
Thanks @francbartoli.  @kalxas let us know if this works/any issues.

kalxas:
The package seems to be in Unstable: https://packages.debian.org/sid/python3-pydantic-core

kalxas:
Lets wait until this ends up in a Debian/Ubuntu stable release

kalxas:
Not yet available in Noble

kalxas:
@francbartoli it is not available for jammy, so I looked at noble to see if it exists there

ridoo:
one general question: Why pygeoapi depends on python packages managed by the OS? What about creating a `venv` and install whatever package version you want?

ridoo:
Thanks and sorry for the noise

[...]

PR #2014, opened 2025-05-16
Pydantic v1/2 compatibility

webb-ben:
# Overview
This PR backports Pydantic v1 compatibility to better allow long term package management.

# Related Issue / discussion
https://github.com/geopython/pygeoapi/issues/1766
https://github.com/geopython/pygeoapi/issues/1664
https://github.com/geopython/pygeoapi/issues/1573

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,6 +69,7 @@ ARG ADD_DEB_PACKAGES="\
     python3-netcdf4 \
     python3-pandas \
     python3-psycopg2 \
+    python3-pydantic \
     python3-pymongo \
     python3-pyproj \
     python3-rasterio \
--- a/pygeoapi/models/config.py
+++ b/pygeoapi/models/config.py
@@ -32,11 +32,22 @@
 # =================================================================
 
 from pydantic import BaseModel, Field
+import pydantic
+
+# Handle Pydantic v1/v2 compatibility
+if pydantic.VERSION.startswith("1"):
+    model_validator = "parse_obj"
+    model_fields = "__fields__"
+    regex_param = "regex"
+else:
+    model_validator = "model_validate"
+    model_fields = "model_fields"
+    regex_param = "pattern"
 
 
 class APIRules(BaseModel):
     """ Pydantic model for API design rules that must be adhered to. """
-    api_version: str = Field(pattern=r'^\d+\.\d+\..+$',
+    api_version: str = Field(**{regex_param: r'^\d+\.\d+\..+$'},
                              description="Semantic API version number.")
     url_prefix: str = Field(
         "",
@@ -62,11 +73,12 @@ class APIRules(BaseModel):
         """ Returns a new APIRules instance for the current API version
         and configured rules. """
         obj = {
-            k: v for k, v in rules_config.items() if k in APIRules.model_fields
+            k: v for k, v in rules_config.items() if k in getattr(APIRules, model_fields)
         }
         # Validation will fail if required `api_version` is missing
         # or if `api_version` is not a semantic version number
-        return APIRules.model_validate(obj)
+        model_validator_ = getattr(APIRules, model_validator)
+        return model_validator_(obj)
 
     @property
     def response_headers(self) -> dict:
--- a/pygeoapi/models/config.py
+++ b/pygeoapi/models/config.py
@@ -1,5 +1,3 @@
-# ****************************** -*-
-# flake8: noqa
 # =================================================================
 #
 # Authors: Sander Schaminee <sander.schaminee@geocat.net>
@@ -75,11 +73,10 @@ class APIRules(BaseModel):
         """
         Returns a new APIRules instance for the current API version
         and configured rules.
-
-        :param 
         """
         obj = {
-            k: v for k, v in rules_config.items() if k in getattr(APIRules, model_fields)
+            k: v for k, v in rules_config.items()
+            if k in getattr(APIRules, model_fields)
         }
         # Validation will fail if required `api_version` is missing
         # or if `api_version` is not a semantic version number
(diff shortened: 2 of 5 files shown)


kalxas (approved):

kalxas:
Thank you @webb-ben for this contribution!
This PR unblocks usage/packaging of pydantic 1.x in some production environments.

kalxas:
@francbartoli yes, that was the plan but unfortunately nothing has yet been ported to the stable branch.
If we could support both pydantic versions it would be easier to release from master.
No changes are expected for Dockerfile, it would remain as is (pydantic 2.x)

Another option is keep this PR just for the stable branch.

Regarding the breakage you point to, should we add some test to be able to detect?

GeoSander:
Wow, I haven't checked in on the project for a while 😅 
Don't know when this was introduced, but I am happy to see that Pydantic is now being used in pygeoapi! 👍

tomkralidis:
> I'd like to make you aware that
> 
> ```python
> class OAPIFormat(BaseModel):
>     root: SupportedFormats = SupportedFormats.YAML
> ```
> 
> will break for example
> 

Looks like CI is passing.  Do you have a use case/workflow to reproduce the error?

webb-ben:
We don't explicitly test against pydantic 1 and 2 anywhere. Hesitant to naively add to the current workflow out of fear it might become an uncontrollable matrix of various dependencies.

webb-ben:
@francbartoli I made that additional change. Could you take a look/review?

doublebyte1:
I think this change is breaking the GetTile metadata function on MVT-tippecanoe:

https://github.com/geopython/pygeoapi/blob/45cac60bc887ddbbb62d59f0d28a25f7265d18d3/pygeoapi/provider/mvt_tippecanoe.py#L301-L306

To reproduce the error, try to get [the tile set metadata on the dive into pygeoapi workshop](https://github.com/geopython/diving-into-pygeoapi/tree/main/workshop/exercises), using the latest pygeoapi docker image:
http://localhost:5000/collections/hyderabad/tiles/WebMercatorQuad/metadata

> pygeoapi  |     return self.get_html_metadata(dataset, server_url, layer,
pygeoapi  |   File "/pygeoapi/pygeoapi/provider/mvt_tippecanoe.py", line 305, in get_html_metadata
pygeoapi  |     metadata['metadata'] = content.model_dump()
pygeoapi  | AttributeError: 'MVTTilesJson' object has no attribute 'model_dump'

If you try the same thing with pygeoapi:0.20.0, it works.

webb-ben:
I believe the issue is not caused by this PR so much as what this PR did not address. It appears we are using a mix of Pydantic v1 https://github.com/geopython/pygeoapi/blob/45cac60bc887ddbbb62d59f0d28a25f7265d18d3/pygeoapi/provider/mvt_proxy.py#L262 and v2 https://github.com/geopython/pygeoapi/blob/45cac60bc887ddbbb62d59f0d28a25f7265d18d3/pygeoapi/provider/mvt_tippecanoe.py#L367 terminology. #2056 should address this for the remaining pieces that use Pydantic

doublebyte1:
> I believe the issue is not caused by this PR so much as what this PR did not address. It appears we are using a mix of Pydantic v1
> 
> https://github.com/geopython/pygeoapi/blob/45cac60bc887ddbbb62d59f0d28a25f7265d18d3/pygeoapi/provider/mvt_proxy.py#L262
> 
> and v2
> https://github.com/geopython/pygeoapi/blob/45cac60bc887ddbbb62d59f0d28a25f7265d18d3/pygeoapi/provider/mvt_tippecanoe.py#L367
> 
> terminology. #2056 should address this for the remaining pieces that use Pydantic

@webb-ben  I am a bit confused, because the code works on pygeoapi master but not in the latest docker image. That is why I thought it was related to one of the dependencies in the Dockerfile. In any case, the safeguards you introduced in your PR address the problem of multiple versions of pydantic.

doublebyte1:
@webb-ben  This is the pydantic version installed in the Docker image of pygeoapi:
pydantic==1.8.2

francbartoli:
@doublebyte1 you might have the version v1 of pydantic in your virtual environment. Can you try again with a brand new virtualenv?

doublebyte1:
> @doublebyte1 you might have the version v1 of pydantic in your virtual environment. Can you try again with a brand new virtualenv?

@francbartoli  That was the version in the docker image. In my virtual environment I have:
pydantic==2.11.7

doublebyte1:
```
pip freeze | grep pydantic
pydantic==2.11.7
pydantic_core==2.33.2

docker build -t pygeoapi-new .
docker run -p 5000:80 -v $(pwd)/example-config.yml:/pygeoapi/local.config.yml -it pygeoapi-new

docker exec -it strange_kare bash
root@02ea09fd3e1b:/pygeoapi# pip freeze | grep pydantic
pydantic==1.8.2

```

tomkralidis:
@doublebyte1 this is because the pygeoapi Docker image is built on Ubuntu jammy, which uses `python3-pydantic` 1.8.2: https://launchpad.net/ubuntu/jammy/i386/python3-pydantic.  If you are building master off of pip3/etc., then pydantic is unpinned in current master.

The purpose of this PR was to ensure compatibility, and it seems that we have some edge cases.  Perhaps we can have some tests which cover these edge cases?

webb-ben:
As noted in this PR, we could easily add a matrix for different versions of pydantic to test across pygeoapi but I worry that incorrectly opens the door to testing against multiple versions of _any_ python package.

francbartoli:
> > @doublebyte1 you might have the version v1 of pydantic in your virtual environment. Can you try again with a brand new virtualenv?
> 
> @francbartoli That was the version in the docker image. In my virtual environment I have: pydantic==2.11.7

Exactly @doublebyte1, in the docker image the methods from pydantic v2 across `mvt` classes cannot be executed because we have pydantic v1 and those methods don't exist anymore

PR #2016, opened 2025-05-22
Allow PYGEOAPI_ env variables in the docker container to be overriden

perok:
# Overview

Hi! We have a need to be able to override the environment variables for where the pygeoapi container is writing the openapi file and reading the config file.

The following change allows opt-in changes to those environment variables.

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -36,9 +36,9 @@ echo "START /entrypoint.sh"
 
 set +e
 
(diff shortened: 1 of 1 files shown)


tomkralidis:
@perok thanks for this.  Can you update the documentation (https://docs.pygeoapi.io/en/latest/running-with-docker.html) as well?  Thanks.

PR #2040, opened 2025-06-11
Add some Python types to the base provider

C-Loftus:
# Overview

- this PR adds Python types to the base provider
- this makes it so users who are writing custom providers can more clearly understand and validate that their providers are correct
- this should not change any runtime behavior

# Related Issue / discussion

N/A

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/base.py
+++ b/pygeoapi/provider/base.py
@@ -51,12 +51,14 @@ class SchemaType(Enum):
     update = 'update'
     replace = 'replace'
 
+
 # All the potential properties on a field
 # as defined by the OGC API features spec
 FieldProperties = TypedDict(
     "FieldProperties",
     {
-        "type": Literal["string", "number", "integer", "boolean", "object", "array"],
+        "type": Literal["string", "number", "integer",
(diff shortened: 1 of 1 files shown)


pygeoapi/provider/base.py
```diff
@@ -43,11 +44,29 @@ class SchemaType(Enum):
...
+    "FieldProperties",
+    {
+        "type": Literal["string", "number", "integer", "boolean", "object", "array"],
+        "title": str,
+        "description": str,
+        "format": NotRequired[str],
+        "x-ogc-unit": NotRequired[str],
+        "x-ogc-role": NotRequired[str],
+        "enum": NotRequired[list[str]],
+    },
```
C-Loftus: @webb-ben does this seem like the correct interpretation of the queryables spec? https://docs.ogc.org/is/19-079r2/19-079r2.html#queryables 

It appears there are some other properties related to uuid's / uri's but I have never actually seen these in practice. So as such not sure if they should be on the properties type.
image

PR #2045, opened 2025-06-17
Add devcontainer config

C-Loftus:
# Overview

It can be difficult for a new user to set up a dev environment for pygeoapi. For instance, if they are on Windows, if they can't install the proper version of gdal for some reason, or have something unusual about their system's Python environment. Additionally, for some folks who install dependencies locally, what they install may end up having differences from what eventually gets deployed in the final Dockerfile. This can lead to bugs or reduce the amount of potential contributors.

A devcontainer helps to solve these issues. A devcontainer is a cross ide specification for launching a reproducible dev environment. It works with Jetbrains IDEs, VSCode, [Neovim](https://github.com/esensar/nvim-dev-container), and is not bound to a specific platform / company.

In the configuration I added, I only used general settings related to launching the dev environment and nothing related to extensions in the IDE.
- In a future PR, if desired, it could be possible to setup the IDE for specific linters / formatters to match project style guidelines or set up an associated docker compose file to launch the services needed for specific provider integrations. 
    - (i.e. so a user doesn't need to manually launch postgis and load it with data every time a postgres integration needs to be developed locally)

# Related Issue / discussion

Devcontainers were listed by other community members as a useful config in #593 

# Additional information

You can launch and test the container in VSCode by going to the command pallet and typing in `reopen in container`.

image

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


tomkralidis:
Thanks for this.  Suggest to move this PR to https://github.com/geopython/pygeoapi-examples instead, which could have this example as well as the follow up improvements mentioned above.

C-Loftus:
Thanks for your reply!

Personally I feel that the devcontainer config belongs in this repo, if you feel it is appropriate, since it is intended for development with pygeoapi directly. I think it would be out of place in the examples repo since it isn't an example for deployment or integration but rather a tool for setting up a dev environment for this codebase itself and is IDE agnostic. If you feel differently please let me know! Thanks

tomkralidis:
Does this setup work with `/bin/vi` or `/usr/bin/vim` directly?  Does it work for non-Docker development workflows?  If this is valuable for certain developers great, but not in the codebase, hence pygeoapi-examples or elsewhere.

C-Loftus:
>  Does it work for non-Docker development workflows?

Yes you can use devcontainers with podman or another container runtime.  

> Does this setup work with /bin/vi or /usr/bin/vim directly? 

Yup since devcontainers are just an open spec you can use it however you wish. You can launch the devcontainer both directly with an IDE like VSCode/Jetbrains, or the [cli](https://github.com/devcontainers/cli) and then do whatever you want inside. Ultimately it is a convenience wrapper that I feel would help make it easier for more folks to contribute and it shares the same base Dockerfile so there is low chance for configuration drift.

My original goal was to not add anything specific to particular IDE but it would be easy to add vim into the preinstalled features so you automatically enter into vim or a similar cli editor/toolchain.

PR #2056, opened 2025-07-03
Pydantic v1/2 compatibility for provider models

webb-ben:
# Overview
Extending https://github.com/geopython/pygeoapi/pull/2014 to ensure pydantic models are safely used across pygeoapi.
This PR makes `model_dump` from Pydantic v2 the standard vocabulary - patching the models that are created with pydantic v1 to wrap `dict` in an execution of `model_dump`

cc: @doublebyte1

# Related Issue / discussion

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/models/provider/base.py
+++ b/pygeoapi/models/provider/base.py
@@ -35,8 +35,8 @@ from datetime import datetime
 from enum import Enum
 from typing import List, Optional
 
-from pydantic import BaseModel
 import pydantic
+from pydantic import BaseModel
 
 
 class TilesMetadataFormat(str, Enum):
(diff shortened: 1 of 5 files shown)


doublebyte1 (changes requested):
@webb-ben Thanks for addressing this! 

Could you also add  

`if pydantic.VERSION.startswith('1'): on mvt_tippecanoe.py? 🙏🏽 
`

https://github.com/geopython/pygeoapi/blob/45cac60bc887ddbbb62d59f0d28a25f7265d18d3/pygeoapi/provider/mvt_tippecanoe.py#L305

doublebyte1:
@webb-ben Thank you for addressing this issue! 🙏🏽 With these guards, it works with both versions of pydantic

doublebyte1 (approved):

PR #2052, opened 2025-06-30
update contributor guidelines and dev docs (#2032)

tomkralidis:
# Overview
This PR updates contributing guidelines as well as additional details for PRs, testing, linting and documentation.

# Related Issue / discussion
Fixes #2023

# Additional information
cc @danielsparing
# Dependency policy (RFC2)

- [x] I have ensured that this PR meets [RFC2](https://pygeoapi.io/development/rfc/2) requirements

# Updates to public demo

- [x] I have ensured that breaking changes to the [pygeoapi master demo server](https://github.com/geopython/demo.pygeoapi.io) have been addressed
  - [ ] https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_master/local.config.yml

# Contributions and licensing

(as per https://github.com/geopython/pygeoapi/blob/master/CONTRIBUTING.md#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

PR #2057, opened 2025-07-04
fix link format (mix of Markdown and restructured text)

jmckenna:
- followup to recent changes in #2052


tomkralidis:
Thanks @jmckenna!  For the record, Jeff has already previously agreed to the pygeoapi Contributions and Licensing Guidelines in #1518

Issue #2000, opened 2025-04-23
Behaviour on bbox queries for providers that do not support them


doublebyte1:
This issue is just to flag the fact that bbox queries issued against providers that do not support them, simply ignore the bbox parameter.

For example, in the case of the GeoJSON provider:

https://github.com/geopython/pygeoapi/blob/451f1f0fcf33bb3a2996530ee37a260f3ba9fab5/pygeoapi/provider/geojson.py#L156-L158

Would it be more appropriated to throw a warning to the user, stating that the bbox filter is not implemented?

I am not sure if there is a rule on how to deal with this, or if this change would be consistent with the pygeoapi behaviour in other occasions, but I would love to hear more thoughts about this.

tomkralidis:
Options:
- 1: emit `LOGGER.warning` (which would not emit to the client)
- 2: throw HTTP 501
- other options?

Having said this, somewhat related PR in #2058

doublebyte1:
> Options:
> 
> * 1: emit `LOGGER.warning` (which would not emit to the client)
> * 2: throw HTTP 501
> * other options?
> 
> Having said this, somewhat related PR in [#2058](https://github.com/geopython/pygeoapi/pull/2058)

I find 501 a bit extreme, but silently failing is also not ideal. Maybe option 1 would be fine? It could even be LOGGER.error

mikemahoney218-usgs:
I think a 501 is preferable -- otherwise as a user I'd assume that my query argument had done something. My other thought would be to not document the bbox parameter for these endpoints and then return a 400, indicating the user was using unavailable query arguments.

tomkralidis:
Thanks @mikemahoney218-usgs for the feedback.

@doublebyte1 seeing that #2058 is now merged, do we have any remaining providers that do not support bbox?  If so, we can close this issue.

doublebyte1:
I think we can close it. Thank you @tomkralidis

PR #2058, opened 2025-07-04
add bbox support for CSV and GeoJSON feature providers

tomkralidis:
# Overview
This PR adds bbox support for CSV and GeoJSON providers using Shapely (which is already part of the core dependency chain).

# Related Issue / discussion
Somewhat related to #2000

# Additional information
None
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/csv_.py
+++ b/pygeoapi/provider/csv_.py
@@ -136,6 +136,7 @@ class CSVProvider(BaseProvider):
                         [p[prop[0]] == prop[1] for prop in properties]), data_)
 
             if bbox:
+                LOGGER.debug('processing bbox parameter')
                 data_ = filter(
                     lambda f: all(
                         [self._intersects(f, bbox)]), data_)
@@ -214,7 +215,7 @@ class CSVProvider(BaseProvider):
         """
 
         if None in [data.get(self.geometry_x), data.get(self.geometry_y)]:
-            return False
+            return True
 
         point = Point(data[self.geometry_x], data[self.geometry_y])
         bbox2 = box(*bbox)
--- a/pygeoapi/provider/csv_.py
+++ b/pygeoapi/provider/csv_.py
@@ -32,6 +32,8 @@ import csv
 import itertools
 import logging
 
+from shapely import box, Point
+
 from pygeoapi.provider.base import (BaseProvider, ProviderInvalidQueryError,
                                     ProviderItemNotFoundError,
                                     ProviderQueryError)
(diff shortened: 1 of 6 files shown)


tests/api/test_itemtypes.py
```diff
@@ -263,20 +263,29 @@ def test_get_collection_items(config, api_):
...
+    })
+    rsp_headers, code, response = get_collection_items(api_, req, 'obs')
+    features = json.loads(response)
+
     assert len(features['features']) == 1
 
     links = features['links']
     assert len(links) == 6
-    assert '/collections/obs/items?f=json&limit=1&bbox=-180,90,180,90' in \
+    assert '/collections/obs/items?f=json&limit=1' in \
```
doublebyte1: If the bbox tests are removed from here, does it mean we are going to create separate tests for each features (and records) provider?

pygeoapi/provider/csv_.py
```diff
@@ -131,10 +135,16 @@ def _load(self, offset=0, limit=10, resulttype='results',
                     lambda p: all(
                         [p[prop[0]] == prop[1] for prop in properties]), data_)
 
+            if bbox:
```
doublebyte1, line 138: I think I would add a log message here:

`LOGGER.debug('processing bbox parameter')`

pygeoapi/provider/geojson.py
```diff
@@ -121,6 +124,11 @@ def _load(self, skip_geometry=None, properties=[], select_properties=[]):
             data['features'] = [f for f in data['features'] if \
                 all([str(f['properties'][p[0]]) == str(p[1]) for p in properties])]  # noqa
 
+        # filter by bbox if set
+        if bbox:
+            data['features'] = [f for f in data['features'] if \
```
doublebyte1, line 130: `LOGGER.debug('processing bbox parameter')`

doublebyte1 (commented):
+1 for these improvements! 👍🏽 

Please check some minor comments.

tests/api/test_itemtypes.py
tomkralidis: I removed this test because of previous behaviour of WxS (if property is null, it doesn't evaluate to true n a query/match).  Looking at OAFeat [Req 24C](https://docs.ogc.org/is/17-069r4/17-069r4.html#_parameter_bbox), I've updated this to true, hence restorng the existing behaviour in `tests/api/test_itemtypes.py` (the remaining changes being updating the miny value in the request and the resulting links).

doublebyte1 (approved):

PR #2060, opened 2025-07-07
Send url details to provider

### LINKED ISSUE #2055: Make uri and serialized_query_parameters available to providers in itemtypes?

Related to #1981 , we've been working to move away from limit-offset paging for our Postgres provider, as deep pages using large offsets are inefficient and can cause users to download duplicate records if earlier pages have been updated during a query. In #1982 , we altered itemtypes so that "next" links could be returned by a provider without pygeoapi then appending its own offset-based next links to the response.

The thing we missed was that pygeoapi uses `uri` and `serialized_query_params` to generate the `next_href` object, and these objects aren't available from inside a provider:
https://github.com/geopython/pygeoapi/blob/c976bd7f61d60a50536c84a2663e506328749789/pygeoapi/api/itemtypes.py#L567

This makes it challenging to generate next links inside a provider.

We'd like `uri` and `serialized_query_params` to be passed to providers to be used to generate next links if desired.

We could alternatively have some way in itemtypes to pass our "page key" from the provider and have itemtypes use that to generate the link. If `offset` wasn't validated to be numeric, we'd just need some way for the provider to control `next_` here:
https://github.com/geopython/pygeoapi/blob/c976bd7f61d60a50536c84a2663e506328749789/pygeoapi/api/itemtypes.py#L567

I think that approach is a bit less friendly than letting giving the provider these objects.

allenj:
# Overview
Send the URI query params and base URI to the provider so it can generate full nextlinks.

Resolves #2055 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute this feature to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/pygeoapi/api/itemtypes.py
+++ b/pygeoapi/api/itemtypes.py
@@ -557,12 +557,12 @@ def get_collection_items(
                           select_properties=select_properties,
                           crs_transform_spec=crs_transform_spec,
                           q=q, language=prv_locale, filterq=filter_,
-                          serialized_query_params=serialized_query_params, uri=uri)
+                          serialized_query_params=serialized_query_params,
+                          uri=uri)
     except ProviderGenericError as err:
         return api.get_exception(
             err.http_status_code, headers, request.format,
             err.ogc_exception_code, err.message)
-
 
     if 'links' not in content:
         content['links'] = []


allenj:
@doublebyte1 How do these latest changes look? Also, I've [commented](https://github.com/geopython/pygeoapi/issues/2055#issuecomment-3136745771) on the upstream issue regarding a comment from @webb-ben

allenj:
Another bump on this.

PR #2066, opened 2025-07-17
OATiles: remove popup and id_field

tomkralidis:
# Overview
This PR fixes hardcoded id values for OATiles HTML rendering.
# Related Issue / discussion

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/data-publishing/ogcapi-tiles.rst
+++ b/docs/source/data-publishing/ogcapi-tiles.rst
@@ -157,6 +157,7 @@ This code block shows how to configure pygeoapi to render Mapbox vector tiles fr
              user: postgres
              password: postgres
              search_path: [osm, public]
+         id_field: osm_id
          table: hotosm_bdi_waterways
          geom_field: foo_geom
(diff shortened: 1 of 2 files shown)


docs/source/data-publishing/ogcapi-tiles.rst
```diff
@@ -157,7 +157,6 @@ This code block shows how to configure pygeoapi to render Mapbox vector tiles fr
              user: postgres
              password: postgres
              search_path: [osm, public]
-         id_field: osm_id
```
webb-ben: `id_field` is required by the GenericSQLProvider. Because it is render the tiles on the fly from vector layer, it also has to set the feature id. You'll notice the tests should fail if you also remove https://github.com/geopython/pygeoapi/blob/master/tests/test_postgresql_mvt_provider.py#L57

tomkralidis: Fixed.

doublebyte1 (approved):

PR #2068, opened 2025-07-18
Fix NoData values for Coverage JSON output

geographika:
# Overview

As seen during the Diving into pygeoapi workshop the JSON coverage output was broken due to NaN values in the output. 

error

This PR ensures NaN are converted to Python `None` values, which are then converted to `null` values in the JSON output, using the same approach as in the xarray driver - https://github.com/geopython/pygeoapi/blob/master/pygeoapi/provider/xarray_.py#L395-L398

image

# Dependency policy (RFC2)

- [X ] I have ensured that this PR meets [RFC2](https://pygeoapi.io/development/rfc/2) requirements

# Contributions and licensing

- [X] I'd like to contribute this bugfix to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/pygeoapi/provider/rasterio_.py
+++ b/pygeoapi/provider/rasterio_.py
@@ -29,6 +29,7 @@
 
 import logging
 
+import numpy as np
 from pyproj import CRS, Transformer
 import rasterio
 from rasterio.io import MemoryFile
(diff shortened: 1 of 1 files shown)


geographika:
The issue was that the NumPy [tolist](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tolist.html) function originally used converts to *to the nearest compatible builtin Python type*, however it does not convert `nan` to Python `None` as `nan` is a valid Python float:

```python
>>> type(float('nan'))
<class 'float'>
```

So any `nan` values are left as-is, and persisted to the JSON output causing invalid JSON. 

>>> import numpy as np
>>> arr = np.array([1.0, np.nan, 3.0])
>>> arr.tolist()
[1.0, nan, 3.0]

This PR ensures `nan` are converted to `None` and then when converted to JSON are returned as `null`. `NaN` is not a valid JSON value. `tolist` takes care of converting other NumPy types such as `np.uint32(2)` (so no changes are required in `json_serial` in `util.py`. 

cc @tomkralidis

geographika:
Maybe estonia_light.tif can be added as a test (it is 11KB).

PR #2071, opened 2025-07-27
make support for features/records extra query parameters configurable

tomkralidis:
# Overview
This PR enables pygeoapi, by default, to ignore extra query parameters for features/records queries.

An added provider level `include_extra_query_parameters` directive (`true|false`, default `false`) can control whether to enable provider support.

The update in #1673 results in all query parameters being passed and evaluated.  pygeoapi's OpenAPI definition declares `additionalProperties: true` for all `.../items` endpoints by default, which allows us to evaluate extra query parameters as desired.

As a result, extra query parameters are always ignored unless set in provider config.

# Related Issue / discussion
#1673 

# Additional information
Some CITE config fixes and removal of leftover print statements are also included in this PR.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api/itemtypes.py
+++ b/pygeoapi/api/itemtypes.py
@@ -415,16 +415,16 @@ def get_collection_items(
 
     LOGGER.debug('processing property parameters')
     for k, v in request.params.items():
-        add_param = False
+        include_query_param = False
         if k not in reserved_fieldnames:
             if k in list(p.fields.keys()) or p.include_extra_query_parameters:
-                add_param = True
+                include_query_param = True
 
-        if add_param:
-            LOGGER.debug(f'Adding property filter {k}={v}')
+        if include_query_param:
+            LOGGER.debug(f'Including query parameter {k}={v}')
             properties.append((k, v))
         else:
-            LOGGER.debug(f'Discarding additional property filter {k}={v}')  # noqa
+            LOGGER.debug(f'Discarding query parameter {k}={v}')
 
     LOGGER.debug('processing sort parameter')
     val = request.params.get('sortby')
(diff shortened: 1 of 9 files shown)


pygeoapi/api/itemtypes.py
```diff
@@ -415,13 +415,16 @@ def get_collection_items(
...
         if k not in reserved_fieldnames:
-            if k in list(p.fields.keys()):
-                LOGGER.debug(f'Adding property filter {k}={v}')
-            else:
-                LOGGER.debug(f'Adding additional property filter {k}={v}')
+            if k in list(p.fields.keys()) or p.include_extra_query_parameters:
+                add_param = True
 
+        if add_param:
+            LOGGER.debug(f'Adding property filter {k}={v}')
```
doublebyte1: In the case of add_param=True because of `p.include_extra_query_parameters` is not really a property filter, is it? I find this comment to be a bit misleading

doublebyte1 (approved):
+1 for making the extra parameters functionality more tightly integrated in the code. If there are any providers where enabling this is particularly important for supporting a particular behaviour, I think that should be mentioned in the documentation.

tests/cite/cite.config.yml
```diff
@@ -10,7 +10,7 @@ server:
     pretty_print: true
     limits:
         default_items: 10
-        max_items: 10
+        max_items: 50
```
doublebyte1, line 13: Thank you for fixing this! We can always use this file as reference: https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_cite/pygeoapi/cite.config.yml

PR #2077, opened 2025-08-11
Externalizing tilematrixsets to resources/tilematrixsets dir

lindakarlovska:
# Overview
- Moves the hardcoded TMS definitions to the dedicated directory in resources/tilematrixsets
- Uses the definition from OGC registry: https://github.com/opengeospatial/2D-Tile-Matrix-Set/tree/master/registry/json
- Maps the OGC registry tags to the tag used in pygeoapi

# Additional info
Preparation for upcoming changes to support custom TMSs

# Contributions and licensing


lindakarlovska:
In https://github.com/opengeospatial/2D-Tile-Matrix-Set/blob/master/registry/json/, the number of tile matrices is different - 24 for WorldCRS84Quad.json and 25 for WebMercatorQuad.json. Currently, pygeoapi assumes both tilematrixsets have 30 tile matrices. I have updated the tests accordingly.

pygeoapi/models/provider/base.py
```diff
@@ -33,12 +33,18 @@
 
 from datetime import datetime
 from enum import Enum
+from pathlib import Path
 from typing import List, Optional
+import json
```
tomkralidis: Alphabetize imports

pygeoapi/models/provider/base.py
```diff
@@ -33,12 +33,18 @@
...
 
 import pydantic
 from pydantic import BaseModel
 
 
+BASE_DIR = Path(__file__).resolve().parents[2]
```
tomkralidis: I think we can use `pygeoapi.util:THISDIR` here:

```python
...
from pygeoapi.util import THISDIR
...
...
TMS_DIR = THISDIR / 'resources' / 'tilematrixsets'
```

pygeoapi/models/provider/base.py
```diff
@@ -81,687 +87,36 @@ class TileMatrixSetEnumType(BaseModel):
...
-      }
-    ]
-    )
+class TileMatrixSetLoader:
+    def __init__(self, directory: Path):
+        self.directory = directory
+
+    def load_from_file(self, filename: str) -> TileMatrixSetEnumType:
+        """Load a single TMS JSON file."""
+        with open(self.directory / filename, encoding="utf-8") as f:
```
tomkralidis:
```python
        filepath = self.directory / filename
        with filepath.open(encoding='utf-8') as fh:
            data = json.load(fh)
```

tomkralidis (changes requested):

Issue #2078, opened 2025-08-11
Item attributes out of order on PostgreSQL data sources


cmotadev:
**Description**
When registering a vector data source (tested on PostgreSQL provider), the order of table columns rendered in pygeoapi doesn't respect the database order. This could be a problem for human clients

**Steps to Reproduce**
Register a vector data source with PostgreSQL provider, as in documentation

**Expected behavior**
The order of columns must respect the data source

**Screenshots/Tracebacks**
- Pygeoapi
<img width="1343" height="862" alt="Image" src="https://github.com/user-attachments/assets/50672a49-39f9-4ec4-83f3-bae3ad4f70db" />

- Database
<img width="1343" height="862" alt="Image" src="https://github.com/user-attachments/assets/968ec8df-96b2-46f7-84f9-b98659d4f35e" />

**Environment**
 - OS: Docker (running on OpenShift )
 - Python version: as in container
 - pygeoapi version: 0.21.0

**Additional context**

PR #2085, opened 2025-08-14
maintain order of PostgreSQL table columns (#2078)

tomkralidis:
# Overview
This PR preserves the order of PostgreSQL columns when emitting feature properties.
# Related Issue / discussion
#2078 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/sql.py
+++ b/pygeoapi/provider/sql.py
@@ -242,7 +242,8 @@ class GenericSQLProvider(BaseProvider):
             ):
                 response['numberReturned'] += 1
                 response['features'].append(
-                    self._sqlalchemy_to_feature(item, crs_transform_out)
+                    self._sqlalchemy_to_feature(item, crs_transform_out,
+                                                select_properties)
                 )
 
         return response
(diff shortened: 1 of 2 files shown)


pygeoapi/provider/sql.py
```diff
@@ -447,18 +448,16 @@ def _store_db_parameters(self, parameters, options):
...
-        item_dict.pop('_sa_instance_state')  # Internal SQLAlchemy metadata
-        feature['properties'] = item_dict
-        feature['id'] = item_dict.pop(self.id_field)
+        # set feature id
+        feature['id'] = item.__dict__[self.id_field]
 
         # Convert geometry to GeoJSON style
-        if feature['properties'].get(self.geom):
-            wkb_geom = feature['properties'].pop(self.geom)
+        if item.__dict__.get(self.geom) is not None:
```
webb-ben: Is there a reason not to do 

`hasattr(item, self.geom)`?

tomkralidis: We are testing for not None, not the existence of the member.

webb-ben: I mean is there a reason we are using the dictionary entity if we are accessing kv pairs that can be done on the item itself?

The dictionary representation should be the same keys as the ones used in 
hasattr and getattr, right? Why are we using the dictionary accessor at all?

webb-ben (approved):

pygeoapi/provider/sql.py
tomkralidis: I've aliased the item to a dictionary again and accessing accordingly, as well added a docstring.

Issue #1753, opened 2024-07-24
update Docker container offerings


tomkralidis:
Currently, pygeoapi produces Docker images via [GitHub Actions](https://github.com/geopython/pygeoapi/blob/master/.github/workflows/containers.yml) to the following container registries:

- [Docker Hub](https://hub.docker.com/r/geopython/pygeoapi) (on release, and latest based on master branch)
- [GitHub Container Registry](https://github.com/geopython/pygeoapi/pkgs/container/pygeoapi) (latest based on master branch)

Currently, the latest image is ~1.5GB containing numerous optional dependencies.  It would be valuable to offer various images based on different deployment profiles/environments.

- [ ] determine images to produce (e.g. slim, default, full)
- [ ] add/update `Dockerfile` setups accordingly
- [ ] update GitHub Action for container publication
- [ ] update [documentation](https://docs.pygeoapi.io/en/latest/running-with-docker.html)

cc @justb4 @francbartoli @webb-ben @ricardogsilva

kalxas:
I have been experimenting with a slim version of our Docker image. I agree it would be good to have different images for different use cases.

mikebelanger:
I'd be interested in what you've found too @kalxas.

cmotadev:
Hey guys, I’ve already did it with Debian Bookworm and python. With some
deep modifications. I’m testing some stuff but the results are promising

Greetings

Carlos Eduardo Mota


Em ter., 15 de abr. de 2025 às 13:24, Mike ***@***.***>
escreveu:

cmotadev:
> Hey guys, I’ve already did it with Debian Bookworm and python. With some
> deep modifications. I’m testing some stuff but the results are promising
> 
> Greetings
> 
> Carlos Eduardo Mota
> 
> 
> Em ter., 15 de abr. de 2025 às 13:24, Mike ***@***.***>
> escreveu:
> […](#)

@tomkralidis @kalxas @mikebelanger 

PyGeoAPI`s debian based image, multi-staged, rootless and virtualenv-based (experimental)

https://github.com/nds-cprm/pygeoapi.git

webb-ben:
> * determine images to produce (e.g. slim, default, full)

- `default` is the current Dockerfile and requirements-docker.txt to constitute the default image.

- `slim` would constitute anything needed to run core excluding the dependencies for `providers/*`, `processes/*`, and `/pubsub/*`.

- `full` would include django and starlette deps?

[...]

PR #2088, opened 2025-08-14
Make PyGeoAPI to run rootless (Tested in Openshift)

cmotadev:
# Overview
This PR aims to modify mininal things in entrypoint script to allow the container to run as a non-root user. This is specially important to Openshift because it was denied by default to run rooted containers. 

this PR could be summarized in two items:
1. Allow to modify PYGEOAPI_CONFIG e PYGEOAPI_OPENAPI in runtime.
2. Remove the necessity to chmod python files on hot reload execution, due of using less privileged uids.

This is important because the entrypoint script tries to generate OpenAPI yaml, and the pygeoapi executable can't create on his own dir. We set the env to write in /tmp dir.

# Related Issue / discussion
Add/Update container offerings: https://github.com/geopython/pygeoapi/issues/1753

# Additional information
This two modifications, in addition to set CONTAINER_PORT > 1024 (we used 8080) allowed us to run pygeoapi in OpenShift with default security configurations. 

docker run example:
```bash
# Using a local-builded image
docker run \
  -p 5000:8080 \
  -d \
  --rm \
  -v ./pygeoapi-config.yml:/tmp/local.config.yml \
  -e "PYGEOAPI_CONFIG=/tmp/local.config.yml" \
  -e "PYGEOAPI_OPENAPI=/tmp/local.openapi.yml" \
  -e "CONTAINER_PORT=8080" \
  -u 1000 
ndscprm/pygeoapi:dev run-with-hot-reload
```

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -121,10 +121,12 @@ case ${entry_cmd} in
 
 	# Run pygeoapi server with hot reload
 	run-with-hot-reload)
-		# Lock all Python files (for gunicorn hot reload)
-		# NOTE: Unecessary when running with non-privileged user. 
-		# If running as root, we can change permissions back, with a dummy shell
-		# find . -type f -name "*.py" | xargs chmod 0444
+		# Lock all Python files (for gunicorn hot reload), if running with user root
+		if [[ $(id -u) -eq 0 ]]
+		then
+			echo "Running pygeoapi as root"
+			find . -type f -name "*.py" | xargs chmod 0444
+		fi
 
 		# Start with hot reload options
 		start_gunicorn --reload --reload-extra-file ${PYGEOAPI_CONFIG}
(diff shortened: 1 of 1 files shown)


cmotadev:
This [commit](https://github.com/geopython/pygeoapi/pull/2088/commits/4e2175906d0a52cf950af1babd6fbe79345aa544) does the same as this PR: https://github.com/geopython/pygeoapi/pull/2016

docker/entrypoint.sh
```diff
@@ -115,7 +122,9 @@ case ${entry_cmd} in
 	# Run pygeoapi server with hot reload
 	run-with-hot-reload)
 		# Lock all Python files (for gunicorn hot reload)
-		find . -type f -name "*.py" | xargs chmod 0444
+		# NOTE: Unecessary when running with non-privileged user.
```
tomkralidis: We need to keep this line for as long as the Dockerfile is running as root, to ensure that any Python file-based changes do not trigger an unintended reload.

Peferably, pygeoapi's Dockerfile should be rootless as well.

cmotadev: No problem. I've added an if statement to check if the app is running as root

PR #2100, opened 2025-09-10
don't remove id property

dblodgett-usgs:
# Overview

Notice that you can query for a feature by its id but the attribute you queried for isn't in the response anymore.

https://labs-beta.waterdata.usgs.gov/api/fabric/pygeoapi/collections/nhdplusv2-huc12/items?huc12=170101010306

Or in terms of the pygeoapi demo:

https://demo.pygeoapi.io/master/collections/dutch_windmills/items?gml_id=Molens.1

Notice that the query is for an id that doesn't get returned in the output. 

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/sql.py
+++ b/pygeoapi/provider/sql.py
@@ -487,8 +487,6 @@ class GenericSQLProvider(BaseProvider):
             if key in item_dict:
                 feature['properties'][key] = item_dict[key]
 
-        feature['properties'].pop(self.id_field, None)
-
         return feature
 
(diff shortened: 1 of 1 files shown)


webb-ben:
Of note, the `id_field` column is included in the [queryables](https://demo.pygeoapi.io/master/collections/dutch_windmills/queryables) and [schema](https://demo.pygeoapi.io/master/collections/dutch_windmills/schema)

webb-ben:
Related issue #2103

PR #2118, opened 2025-09-25
Catch EDR `AttributeError` errors more fine grained

C-Loftus:
# Overview

- Currrently in the EDR base provider there is a `try` `except` over every `AttributeError` in the `query` function
- Thus, if a provider throws an `AttributeError` it will mistakenly say that the query is not implemented and it is possible to get confusing errors like these, when it reality is was coming from an `AttributeError` in the provider

```
NotImplementedError: Query "cube" not implemented in list ['locations', 'cube', 'area']
```

## Fix

This PR fixes this behavior to:
- put the try catch on only the `getattr` call so other `AttributeErrors` are preserved
- add the names of the registered queries which is useful for debugging purposes
- explicitly make sure `query_type` is passed in. Otherwise you would try to do `getattr` with `None` which could in the future cause other confusing issues

# Related Issue / discussion

N/A

# Additional information

N/A

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/base_edr.py
+++ b/pygeoapi/provider/base_edr.py
@@ -28,7 +28,6 @@
 # =================================================================
 
 import logging
-from typing import Optional
 
 from pygeoapi.provider.base import (BaseProvider, ProviderInvalidDataError,
                                     ProviderQueryError)


C-Loftus:
I am unclear why the build is failing. The tests all pass locally including the tile test in question that is failing. Is this a quirk of the ci environment in some way that I am overlooking? I can't think of any reason the change to EDR queries would affect tiles

doublebyte1:
> I am unclear why the build is failing. The tests all pass locally including the tile test in question that is failing. Is this a quirk of the ci environment in some way that I am overlooking? I can't think of any reason the change to EDR queries would affect tiles

I think this test is unrelated to your PR. Let me double check and come back to you. Thanks

C-Loftus:
@doublebyte1 It seems there was a fix in the main branch. I have synced this PR so it should be all set now 👍

pygeoapi/provider/base_edr.py
```diff
@@ -116,8 +118,16 @@ def query(self, **kwargs):
 
         :returns: coverage data as `dict` of CoverageJSON or native format
         """
-
+        query_type: Optional[str] = kwargs.get('query_type')
```
tomkralidis:
```python
        query_type = kwargs.get('query_type')

        if query_type is None:
            raise ProviderQueryError('Missing query_type parameter is required')
```

pygeoapi/provider/base_edr.py
```diff
@@ -116,8 +118,16 @@ def query(self, **kwargs):
...
+            raise ProviderQueryError(
+                'query_type parameter is required'
+            )
         try:
-            return getattr(self, kwargs.get('query_type'))(**kwargs)
+            query_function = getattr(self, query_type)
         except AttributeError:
-            raise NotImplementedError('Query not implemented!')
+            raise NotImplementedError(
+                f'Query "{query_type}" not implemented in registered query types: {self.query_types}' # noqa
```
tomkralidis: A client can always autodiscover the query types from collection links.  Suggest, then:

`f'Query type not implemented'`.

tomkralidis (changes requested):

pygeoapi/provider/base_edr.py
C-Loftus: If it works for you, I think the format string might be useful to keep. 

In situations like this bug, the UI had the query registered but it was throwing a not implemented error. In this situation, it is helpful to know what queries are registered and what we requested as it helps to resolve the bug a bit easier. It is also possible for a user to forget the `register` decorator on a function for the provider or make a typo and not have info about what was registered when things go wrong, making it harder to debug.

Let me know if you think otherwise though. Happy to do what is best

pygeoapi/provider/base_edr.py
C-Loftus: Thank you for your comment. If possible, could you clarify why we would want to use `is None`? 

Don't we want to have `not` instead of `is None` to catch the case in which `query_type` is an empty string or something otherwise falsy? Let me know if I am missing something

pygeoapi/provider/base_edr.py
tomkralidis: We need to keep the Robustness Principle here ("be liberal in what you accept, and conservative in what you send").

pygeoapi/provider/base_edr.py
tomkralidis: `None` is what is returned from a dictionary `.get()` when a given property is not found.

C-Loftus:
@tomkralidis thanks for your comments; I have made the changes you requested and believe this should be all set now.

pygeoapi/provider/base_edr.py
tomkralidis: `query_type = kwargs.get('query_type')`

pygeoapi/provider/base_edr.py
```diff
@@ -116,8 +118,16 @@ def query(self, **kwargs):
 
         :returns: coverage data as `dict` of CoverageJSON or native format
         """
-
+        query_type: Optional[str] = kwargs.get('query_type')
+        if query_type is None:
+            raise ProviderQueryError(
```
tomkralidis: `raise ProviderQueryError('Query type required')`

pygeoapi/provider/base_edr.py
```diff
@@ -116,8 +118,16 @@ def query(self, **kwargs):
...
+                'query_type parameter is required'
+            )
         try:
-            return getattr(self, kwargs.get('query_type'))(**kwargs)
+            query_function = getattr(self, query_type)
         except AttributeError:
-            raise NotImplementedError('Query not implemented!')
+            raise NotImplementedError(
```
tomkralidis: `raise ProviderQueryError('Query type not implemented')`

tomkralidis (changes requested):

pygeoapi/provider/base_edr.py
tomkralidis: Can this be on one line please (assuming it does not break flake8)?

pygeoapi/provider/base_edr.py
tomkralidis: As well, here, can this be on one line please (assuming it does not break flake8)?

pygeoapi/provider/base_edr.py
C-Loftus: Done in one line

pygeoapi/provider/base_edr.py
tomkralidis: @C-Loftus ^^

C-Loftus: Thanks for your comment, can you please explain further? Beyond this PR I wasn't sure on pygeoapi's goal with typing.

Without type info, `query_type` has an unknown type and thus no tools such as editors or static analysis tools which use PEP 484 – Type Hints can do any sort of validation. For instance in the docstring it says `:param query_type: query type` which makes sense semantically but does not specify if it is an enum, a string, or something else nor be used for static checks.

Since Python typing is gradual and has no runtime affect I would suggest keeping this.

tomkralidis: We can address typing strategies in a later GitHub issue.  For now, for consistency, suggest changing so we can merge this PR.  Thanks.

C-Loftus: Ok sounds good; I have removed for time being. Yes it would be great to align on that. I have been wondering about that for other future PRs so it would be very useful.

tomkralidis: > Thanks for your comment, can you please explain further? Beyond this PR I wasn't sure on pygeoapi's goal with typing.
> 
> Without type info, `query_type` has an unknown type and thus no tools such as editors or static analysis tools which use PEP 484 – Type Hints can do any sort of validation. For instance in the docstring it says `:param query_type: query type` which makes sense semantically but does not specify if it is an enum, a string, or something else nor be used for static checks.
> 
> Since Python typing is gradual and has no runtime affect I would suggest keeping this.

Note that editors are not a priority/driver for pygeoapi.  As mentioned, we can in an issue/discussion if you like.

C-Loftus: Sure I can make an issue. I am referring to PEP 484 – Type Hints generally which can be used by editors as an example, but are more broadly useful and aren't coupled to a specific tool.

PR #2129, opened 2025-10-11
add support for STAC API v1.0.0

tomkralidis:
# Overview
This PR provides a minimal implementation of STAC API 1.0.0.  The design pattern is to leverage resources already defined via `feature` or `record` providers, and updating the resource `type` to `stac-collection`, which will get picked up by the STAC search functions in `pygeoapi.api.stac:search`.

# Related Issue / discussion
#221

# Additional information
cc @dblodgett-usgs 
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api/stac.py
+++ b/pygeoapi/api/stac.py
@@ -8,7 +8,7 @@
 #          Ricardo Garcia Silva <ricardo.garcia.silva@geobeyond.it>
 #          Bernhard Mallinger <bernhard.mallinger@eox.at>
 #
-# Copyright (c) 2024 Tom Kralidis
+# Copyright (c) 2025 Tom Kralidis
 # Copyright (c) 2025 Francesco Bartoli
 # Copyright (c) 2022 John A Stevenson and Colin Blackburn
 # Copyright (c) 2023 Ricardo Garcia Silva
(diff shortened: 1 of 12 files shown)


kalxas (approved):

pygeoapi/api/stac.py
```diff
@@ -217,6 +234,210 @@ def get_stac_path(api: API, request: APIRequest,
...
+        msg = 'No STAC API collections configured'
+        return api.get_exception(HTTPStatus.INTERNAL_SERVER_ERROR, headers,
+                                 request.format, 'NotApplicable', msg)
+
+    if request.data:
+        LOGGER.debug('Intercepting STAC POST request into query args')
+        request_data = json.loads(request.data)
+        request_params = deepcopy(dict(request.params))
+
+        for qp in ['bbox', 'datetime', 'limit', 'offset']:
```
luipir, line 342: I suppose it's a intended limitation to limit query to space/time instead of parse parameters basing on result of queriable entrypoint of the STAC collection/catalog. Am I right? If so IMHO it should be documented and ask for contribution about this aspect.
In any case, great addition :)

tomkralidis: Thanks for the feedback @luipir.  The PR implements the STAC API Core and Item Search components in STAC API.  I've put forth a PR in #2132 to help clarify for users.

If we need additional functionality for STAC API, this should be put forth in an issue for discussion as a next step.

PR #2131, opened 2025-10-18
update MapScript map provider output formats

tomkralidis:
# Overview
MapServer outputformat object names have changed since 7.0.  Though a safeguard / migration exists to migrate them, this PR updates to the current support to avoid compatibility breakage in the future.

Also some minor doc housekeeping included.

# Related Issue / discussion
None

# Additional information
Thanks to @geographika for the discussion/collaboration.
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/publishing/stac.rst
+++ b/docs/source/publishing/stac.rst
@@ -153,7 +153,7 @@ So, the following rules must be respected:
 File examples
 *************
 
-**Structure of the catalog.json file**
+Structure of the ``catalog.json`` file:
 
 .. code-block:: json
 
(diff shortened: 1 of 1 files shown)


geographika:
As discussed with @tomkralidis the GD renderer was [removed in MapServer 7.0 ](https://mapserver.org/development/rfc/ms-rfc-99.html). Since then any OUTPUTFORMATS starting with `GD/` have been converted in MapServer to `AGG/PNG8` which is why the MapScript code continued to work. 

However, `GD/JPEG` would also map to `AGG/PNG8` which would mean requests assuming JPEG responses would get PNGs. The new `AGG` mappings in this PR fixes this and is future-proof. 

Other possible default MapServer output types which could be useful are:

```
    {"jpegpng", "AGG/MIXED", "image/vnd.jpeg-png"},
    {"jpegpng8", "AGG/MIXED", "image/vnd.jpeg-png8"},
    {"GTiff", "GDAL/GTiff", "image/tiff"},

```

PR #2136, opened 2025-10-21
Add early support for JSON-FG

francbartoli:
# Overview

This PR adds support to the most recent JSON-FG [specification](https://portal.ogc.org/files/107269#_patterns) `0.3`. It returns JSON-FG formatted payload with the following conformance classes:

- `http://www.opengis.net/spec/json-fg-1/0.3/conf/core`
- `http://www.opengis.net/spec/json-fg-1/0.3/conf/types-schemas`

A new formatter has been added to the core for that scope and must rely on `GDAL>=3.12` which supports `0.3` (not yet released as of today). As a consequence, this is a breaking change for our dependencies chain and cannot be merged until `rasterio` will support that version as well.
This potentially might be relaxed after the merge of this branch https://github.com/geopython/pygeoapi/tree/formatters to have the formatter configurable at resource level and so the GDAL requirement needed only if it is configured.

# Related Issue / discussion

- [Configurable formatter](https://github.com/geopython/pygeoapi/tree/formatters)

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/formatter/jsonfg.py
+++ b/pygeoapi/formatter/jsonfg.py
@@ -1,0 +1,128 @@
+# =================================================================
+#
+# Authors: Francesco Bartoli <xbartolone@gmail.com>
+#
+# Copyright (c) 2025 Francesco Bartoli
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation
+# files (the "Software"), to deal in the Software without
+# restriction, including without limitation the rights to use,
+# copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following
+# conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+# OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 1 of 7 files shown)


webb-ben:
How does this provider differ from a (generic) GDAL based formatter capable of formatting to any OGR data type?

francbartoli:
It is very close potentially, what are you imaging?

pygeoapi/formatter/jsonfg.py
```diff
@@ -0,0 +1,137 @@
...
+
+    :returns: dict of converted GeoJSON (JSON-FG)
+    """
+    gdal.UseExceptions()
+    LOGGER.debug("Dump GeoJSON content into a data source")
+    try:
+        with gdal.OpenEx(json.dumps(data)) as srcDS:
+            tmpfile = f"/vsimem/{uuid.uuid1()}.json"
+            LOGGER.debug("Translate GeoJSON into a JSONFG memory file")
+            gdal.VectorTranslate(tmpfile, srcDS, format="JSONFG")
```
webb-ben, line 123: Well once you read it into GDAL, any vector driver transform can occur. Wondering if would be possible to make _any_ output format in the same spirit of the OGR provider - formatter config only barrier to add Shapefile, KML, etc. as an output format for pygeoapi GeoJSON.

tomkralidis:
@francbartoli given #2144 looks like this can be now implemented as a standalone formatter.  Are you able to update accordingly?

tomkralidis:
pygeoapi meeting 2026-05-15
- updating to GDAL 3.12 will facilitate JSON-FG output support
- needs packaging (@kalxas to review)
- consider dependencies/requirements in core (may be a moot point given rasterio is required).

webb-ben:
Will note:

The current CRS implementation in pygeoapi transforms geometry before reaching a formatter. Furthermore, the request CRS is not passed to formatters. This means that the declared CRS of the JSONFG is always defined as WGS84 regardless of the returned projection.

Issue #1997, opened 2025-04-17
Custom Formatter plugin issues


webb-ben:
**Description**
The Formatter Plugin architecture does not appear to be fully integrated. Following the documentation in https://docs.pygeoapi.io/en/latest/plugins.html#example-custom-pygeoapi-formatter, but there is no way for me to use the formatter plugin as an output in pygeoapi.

**Steps to Reproduce**
1. Produce custom formatter plugin
2. Add plugin to `pygeoapi.plugins.PLUGINS.formatters`
3. Query using custom format
4. Results are JSON.

**Expected behavior**
I would expect for there to be a clear way to deploy a custom formatter plugin in pygeoapi. I would expect appropriately registered formatters to be loaded consistent with pygeaopi's plugin architecture. 

**Screenshots/Tracebacks**
Unlike the provider, the formatters are never specified in the configuration to load an external plugin. It appears the formatters are only ever validated to ensure against the keys in `pygeoapi.plugins.PLUGINS.formatters`, see https://github.com/search?q=repo:geopython/pygeoapi+PLUGINS['formatter']&type=code

![Image](https://github.com/user-attachments/assets/d2f78c6c-7c9d-4b56-8953-e47998b2741b)

Furthermore the only formatter plugin in core, CSV is hardcoded https://github.com/geopython/pygeoapi/blob/a90f545978028f029331c115f90e8dc5d93c1239/pygeoapi/api/itemtypes.py#L650-L652

**Environment**
 - OS: All
 - Python version: 3.10
 - pygeoapi version: 0.21.dev0

**Additional context**
It seems a large portion of the formatting can be also done in a Jinja2 template. I think it would benefit pygeoapi to have a clearer picture of how the different ways to format outputs between the Formatter architecture and Jinja2 templates.

IngenieroGeomatico:
I have the same problem, I am creating a custom provider :

example ->  

```
providers:
            - type: feature
              name: pygeoapi.pygeoapi_complementos.Vector_provider.OGR_DataProvider.OGR_DataProvider
              data: https://......zip
            #   id_field: id
            #   title_field: id
```

and a custom output format, defined in the same level as server, metadata, logging,...:

example -> 
```

format:
  topojson:
    mimetype: application/topo+json
    name: pygeoapi.pygeoapi_complementos.Vector_format.topojson_DataFormat.TopoJSON_DataFormatter
    title: TopoJSON

```
 and I can't register it from the configuration yml

[...]

PR #2144, opened 2025-10-24
fix support for custom formatters (#1997)

tomkralidis:
# Overview
This PR updates formatters support to allow for custom formatters for item types and EDR.
# Related Issue / discussion
#1997 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api/__init__.py
+++ b/pygeoapi/api/__init__.py
@@ -1237,6 +1237,15 @@ def describe_collections(api: API, request: APIRequest,
                     'href': f'{api.get_collections_url()}/{k}/{qt}?f={F_HTML}'
                 })
 
+                for key, value in get_dataset_formatters(v).items():
+                    title3 = f'{qt} query for this collection as {key}'
+                    collection['links'].append({
+                        'type': value.mimetype,
+                        'rel': 'data',
+                        'title': title3,
+                        'href': f'{api.get_collections_url()}/{k}/{qt}?f={value.f}'  # noqa
+                    })
+
         if dataset is not None and k == dataset:
             fcm = collection
             break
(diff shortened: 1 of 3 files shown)


pygeoapi/formatter/base.py
```diff
@@ -39,23 +39,23 @@ def __init__(self, formatter_def: dict):
...
+        param formatter_def: formatter definition
 
         :returns: pygeoapi.formatter.base.BaseFormatter
         """
 
+        self.extension = None
         self.mimetype = None
-        self.geom = False
 
         self.name = formatter_def['name']
```
webb-ben: Should this be in a try/catch?

docs/source/plugins.rst
```diff
@@ -435,7 +435,10 @@ The below template provides a minimal example (let's call the file ``mycooljsonf
            """Inherit from parent class"""
 
            super().__init__({'name': 'cooljson', 'geom': None})
-           self.mimetype = 'application/json; subtype:mycooljson'
+           self.f = 'cooljson'  # f= value
```
webb-ben, line 438: Does the formatter need to response both `f=` and the `Accept` Header?

pygeoapi/api/itemtypes.py
```diff
@@ -656,9 +669,9 @@ def get_collection_items(
                                      'collections/items/index.html',
                                      content, request.locale)
         return headers, HTTPStatus.OK, content
-    elif request.format == 'csv':  # render
-        formatter = load_plugin('formatter',
-                                {'name': 'CSV', 'geom': True})
+    elif request.format in [df.f for df in dataset_formatters.values()]:
+        formatter = [v for k, v in dataset_formatters.items() if
```
webb-ben:
```suggestion
        formatter = [v for v in dataset_formatters.values() if
```
Are we using the key in this?

pygeoapi/formatter/base.py
webb-ben: For some reason this is formatted to the wrong line. In the base provider, all required fields are in a try/catch. Should

```
        self.name = formatter_def['name']
```

follow suit?

Issue #2142, opened 2025-10-24
update leaflet code to use ogcapi plugin for maps preview


pvgenuchten:
**Is your feature request related to a problem? Please describe.**
currently pygeoapi mis-uses a wms layer to display ogcapi-maps

```
L.tileLayer.wms("https://demo.pygeoapi.io/master/collections/mapserver_world_map/map?f=png", {"opacity": .7, "transparent": true, "crs": L.CRS.EPSG4326}).addTo(map);
```

**Describe the solution you'd like**
a ogcapi plugin is available these days to add proper ogcapi to leaflet

https://gitlab.com/IvanSanchez/leaflet.imageoverlay.ogcapi

tomkralidis:
Thanks @pvgenuchten (cc @IvanSanchez)

PR #2145, opened 2025-10-24
use ogcapi leaflet plugin, instead of wms

pvgenuchten:
# Overview

so far we were using a wms layer in leaflet to display ogcapi-maps, these days a proper ogcapi plugin is available, switching

# Related Issue / discussion

#2142

# Additional information

this introduces base_url as a property on the response object (provided to jinja), which is required by the ogcapi library 

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/templates/collections/collection.html
+++ b/pygeoapi/templates/collections/collection.html
@@ -148,11 +148,11 @@
     ));
 
    {# if this collection has a map representation, add it to the map #}
-   {% for link in data['links'] %}{% 
-    if link['rel'] == 'http://www.opengis.net/def/rel/ogc/1.0/map' and link['href'] %} 
-        L.imageOverlay.ogcapi("{{ data['base_url'] }}", {collection: "{{ data['id'] }}", "opacity": .7, "transparent": true}).addTo(map);{% 
-    endif %}{% endfor %}
-
+   {% for link in data['links'] %}
+     {% if link['rel'] == 'http://www.opengis.net/def/rel/ogc/1.0/map' and link['href'] %} 
+        L.imageOverlay.ogcapi("{{ data['base_url'] }}", {collection: "{{ data['id'] }}", "opacity": .7, "transparent": true}).addTo(map);
+     {% endif %}
+   {% endfor %}
 
     var bbox_layer = L.polygon([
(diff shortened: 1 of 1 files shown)


pygeoapi/api/__init__.py
```diff
@@ -628,6 +628,9 @@ def get_format_exception(self, request) -> Tuple[dict, int, str]:
             HTTPStatus.BAD_REQUEST, headers,
             request.format, 'InvalidParameterValue', msg)
 
+    def get_base_url(self) -> str:
```
tomkralidis: We already have `pygeoapi.util:get_base_url` (which is used to set `self.base_url`, suggest to remove this function.

pygeoapi/api/__init__.py
```diff
@@ -1347,6 +1350,7 @@ def describe_collections(api: API, request: APIRequest,
         })
 
     if request.format == F_HTML:  # render
+        fcm['base_url'] = api.get_base_url()
```
tomkralidis: `fcm['base_url'] = self..base_url`

pygeoapi/templates/collections/collection.html
```diff
@@ -147,11 +148,10 @@ <h3>{% trans %}Storage CRS{% endtrans %}</h3>
     ));
 
    {# if this collection has a map representation, add it to the map #}
-   {% for link in data['links'] %}
-      {% if link['rel'] == 'http://www.opengis.net/def/rel/ogc/1.0/map' and link['href'] %}
-        L.tileLayer.wms("{{ link['href'] }}", {"opacity": .7, "transparent": true, "crs": L.CRS.EPSG4326}).addTo(map);
-      {% endif %}
-    {% endfor %}
+   {% for link in data['links'] %}{%
```
tomkralidis:
```jinja
{% for link in data['links'] %}
    {% if link['rel'] == 'http://www.opengis.net/def/rel/ogc/1.0/map' and link['href'] %} 
        L.imageOverlay.ogcapi("{{ data['base_url'] }}", {collection: "{{ data['id'] }}", "opacity": .7, "transparent": true}).addTo(map);
    {% endif %}
{% endfor %}
```

Issue #2164, opened 2025-11-21
Position query UI breaks due to non-spatiotemporal parameters being treated as parameters


kieranbartels:
## **Description**

When performing **position queries** on datasets that include **non-spatiotemporal parameters** (e.g., `crs`, `lat_bounds`, or other auxiliary variables), pygeoapi incorrectly interprets these variables as **queryable parameters**.

As a result:

1. pygeoapi attempts to collect **timeseries data** for parameters like `crs`, which do not contain actual geospatial/temporal values.
2. The **position UI popup fails to render**, showing a blank panel.
3. In the **collection Parameters table**, these non-queryable metadata fields appear with `"None"` values for name and units

Filtering out non-geospatial parameters resolves the UI issue and produces a correct parameter list and query result.

---

## **Steps to Reproduce**

1. Use a dataset containing parameters such as `crs`, `lat_bounds`, etc.  
`Example: PRISM dataset at USGS.`

2. Run a position query:  
`https://api.water.usgs.gov/gdp/pygeoapi/collections/PRISM/position?coords=POINT(-92.97663954153481
 32.92670592660468)`

3. Note:
- The JSON response contains valid data for all variables.
- The position UI popup remains empty.

4. If all valid geospatial parameters are explicitly provided: `&parameter-name=ppt,tmn,tmx`, the UI works correctly.

5. Inspect the Parameters list for the collection and note the presence of non-geospatial parameters (`crs`, `lat_bounds`, etc.) with `"None"` values.  
`https://api.water.usgs.gov/gdp/pygeoapi/collections/PRISM`  
`https://api.water.usgs.gov/gdp/pygeoapi/collections/GMO_New`  

---

## **Expected behavior**

- pygeoapi should **ignore non-geospatial parameters** such as `crs`, `lat_bounds`, and other auxiliary dataset fields.
- Only true **spatial/temporal queryable variables** should:
    - Appear in the **Parameters** table.
    - Be used by the position query.

---

### Broken UI when CRS is present:
<img width="904" height="461" alt="Image" src="https://github.com/user-attachments/assets/2aaf4d07-86ed-44df-926f-7d2ee122a22a" />

### Non geospatial fields in the Parameters table
<img width="833" height="659" alt="Image" src="https://github.com/user-attachments/assets/76a74a3e-35b8-4b92-b2ff-297714cfd0e4" />

### Working UI when non geospatial fields are removed/filtered out:
[...]

PR #2165, opened 2025-11-21
Filtering Geospatial Data

kieranbartels:
# Overview

This PR filters out **non-spatiotemporal parameters** (e.g., `crs`, `lat_bounds`, other auxiliary variables) from collections so they are not treated as queryable parameters during position queries or displayed in the collection's Parameters table.  

This resolves the issue where pygeoapi attempts to collect timeseries data for fields like `crs`, causing:

- The **position query UI popup** to render blank.
- **Invalid parameters** (with `"None"` values) to appear in the collection's Parameters table.

After filtering these fields, position queries behave correctly, and the Parameter table only lists actual geospatial/temporal variables. 

---

# Related Issue / discussion

- **Primary Issue:** [geopython/pygeoapi#(insert number when created)](https://github.com/geopython/pygeoapi/issues/2164)
- **USGS original issue:** https://code.usgs.gov/wma/nhgf/pygeoapi/-/issues/65  
- **LimnoTech related issue + fix:** https://github.com/LimnoTech/pygeoapi/pull/11

---

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/xarray_.py
+++ b/pygeoapi/provider/xarray_.py
@@ -108,6 +108,8 @@ class XarrayProvider(BaseProvider):
         if not self._fields:
             for key, value in self._data.variables.items():
                 if key not in self._data.coords:
+                    if not {self.time_field, self.x_field, self.y_field}.issubset(value.dims):
+                        continue
                     LOGGER.debug('Adding variable')
                     dtype = value.dtype
                     if dtype.name.startswith('float'):


webb-ben:
I am not sure of the relative performance differences but I had implemented https://github.com/internetofwater/pygeoapi/blob/e6b12b0224163f4eaf5c58f37049f44fc8dc77e8/pygeoapi/provider/xarray_.py#L120-L122 to continue when no units were found. 

Not understanding OACov enough to say confidently if this is a patch for `xarray_.py` or `xarray_edr.py`. I assume even though the error surfaces in an EDR query, the logics stands that these are also required fields in a coverage?

PR #2167, opened 2025-11-24
Fix WMTSFacade provider

Dolite:
'style' query parameter is mandatory according to WMTS 1.0.0 standard. For backwards compatibility, empty string remains as default value.

# Overview

# Related Issue / discussion

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute bugfix to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/docs/source/publishing/ogcapi-tiles.rst
+++ b/docs/source/publishing/ogcapi-tiles.rst
@@ -196,7 +196,7 @@ This code block shows how to configure pygeoapi to read map tiles from a WMTS.
          options:
              wmts_layer: camb:hex350_grid_mental_1920 # the layer name of the wmts
              wmts_tile_matrix_set: WebMercatorQuad  # the name of the tile matrix set of the wmts.
-             wmts_style: camb:hex350_grid_mental_1920  # the style identifier of the wmts. Default to empty
+             wmts_style: camb:hex350_grid_mental_1920  # the style identifier of the wmts. If empty or this key is missing, it falls back to the default style.
              scheme: WebMercatorQuad  # the aligning scheme in pygeoapi.
              zoom:
                  min: 0


docs/source/publishing/ogcapi-tiles.rst
```diff
@@ -196,6 +196,7 @@ This code block shows how to configure pygeoapi to read map tiles from a WMTS.
          options:
              wmts_layer: camb:hex350_grid_mental_1920 # the layer name of the wmts
              wmts_tile_matrix_set: WebMercatorQuad  # the name of the tile matrix set of the wmts.
+             wmts_style: camb:hex350_grid_mental_1920  # the style identifier of the wmts. Default to empty
```
doublebyte1: "Default to empty"-> "if empty or this key is missing, it falls back to the default style."

pygeoapi/provider/wmts_facade.py
```diff
@@ -161,7 +161,7 @@ def get_tiles(self, layer=None, tileset=None,
                 'tileMatrix': z,
                 'tileRow': y,
                 'tileCol': x,
-                'style': ''
```
doublebyte1, line 164: The current code base is passing the mandatory "style" parameter to WMTS, but not exposing it to the user.

doublebyte1 (changes requested):
@Dolite Thank You for this enhancement to the provider, by exposing the style parameter, which currently is set to ' ' . When this parameter is empty, the server always fallbacks to the default style; In my comment I suggest you update the description on the documentation to make this behaviour more clear.

Please note that this is not a bug, as the provider does not necessarily need to expose all the mandatory parameters of WMTS; I think keeping it optional, as you did, is a good idea; to support backwards compatibility and also to make configuration easier for the users.

docs/source/publishing/ogcapi-tiles.rst
doublebyte1: @Dolite would you mind just updating the comment here, so we can merge the PR? Thank You 🙏🏽

doublebyte1:
Vulnerability check seems to be unrelated

Dolite:
Hi !

Sorry I just see your comments. I had totally forgotten this PR. When I tried your server, requests to source WMTS server were sent without STYLE query parameter and an error occured. That's why I added the style configuration. Thanks for fixing my documentation and for merging the PR.

Have a nice day and a nice year !

Issue #1678, opened 2024-06-12
Support custom JSON-LD context in generic JSON output


rob-metalinkage:
**Is your feature request related to a problem? Please describe.**
Documenting the meaning of feature attributes decreases ambiguity and increases interoperability. JSON-LD provides a means to do this with a reference to a context document, or an embedded context. The current JSON-LD option supports a different model - representing geometry elements in a (set of) particular structures, but does not address semantics of feature properties. 

**Describe the solution you'd like**
Allow definition of feature properties via published JSON-LD contexts - e.g. based on the featureType (for FG-JSON). 

This would be implementable in incremental steps:
1. allow specification of a URL to inject into an @context element in the JSON
2. allow bundling of a context document to be served as a resource and referenced by URL as per case 1
3. Import related documentation from an OGC BuildingBlock to inject into API description
4. Exploitation of JSON-LD context to provide end-users with descriptions of feature properties from linked ontologies.

step 1 provides the key enabler and is a non-breaking change to JSON outputs. Other functionality addresses infrastructure design - e.g. the concept of Feature Type Catalogues and could be designed separately. Code sprints for these would test potential.

**Describe alternatives you've considered**
Existing JSON-LD wrapper - but this changes the JSON which may cause problems and is both a less general solution and less expressive.  Injecting context for the feature properties is a _complementary_ solution, but one that can be safely applied to both the normal JSON and the extended JSON-LD feature model.

**Additional context**
The OGC BuildingBlocks model provides for feature schema definitions with JSON-LD contexts ready to use (and independently testable). JSON-LD contexts for OGC API data schemas are provided, along with ontologies (definitions) of structural elements, so that only feature-specific properties need to be defined.

An example of how this can be leveraged using a domain model - by carefully bundling the Features schema with the relevant properties model is here: https://ogcincubator.github.io/bblocks-examples/bblock/ogc.bbr.examples.feature.externalSchema

ksonda:
Worth checking if this cpaability was implemented in https://github.com/geopython/pygeoapi/pull/868

For example, https://reference.geoconnex.us/collections/gages/items/1000085?f=jsonld implements a complex RDF representation of feature properties

jmckenna:
> 
> For example, https://reference.geoconnex.us/collections/gages/items/1000085?f=jsonld 

Thanks for sharing.  For my quick review of the JSON-LD: I am surprised to see that `"property":"value"` works in the JSON-LD, as it is proper instead to have `"property": "value"`

[...]

PR #2172, opened 2025-12-01
JSON-LD UI enhancements (HTML pages)

avillar:
# Overview

This PR depends on GH-2171. It provides a series of UI components to enhance the HTML views of feature and stac items when a JSON-LD context is available, by mapping the object properties (and linked data values) to their corresponding semantic terms, resolving the terms, and retrieving a human-friendly label and description.

# Additional information

Arrow functions are used, so ES6 is required. They could be replaced and the code made fully ES5 compatible, though.

`<script src="...">` dependencies is to [rdflib.js](https://github.com/linkeddata/rdflib.js) and to [jsonld-ui-utils](https://github.com/avillar/jsonld-ui-utils) are included in the HTML.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [X] I'd like to contribute feature "JSON-LD UI enhancements (HTML pages)" to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

### DIFF
--- a/pygeoapi/templates/collections/items/item.html
+++ b/pygeoapi/templates/collections/items/item.html
@@ -80,6 +80,13 @@
           </div>
         </div>
         <div class="col-md-6 col-sm-12">
+            {% if linked_data and linked_data.get('context') | length %}
+              <div class="text-end mb-2">
+                <a href="#" class="btn btn-primary" id="semantic-lookup-btn" data-target="#item-properties-table">
+                  {% trans %}Semantic lookup{% endtrans %}
+                </a>
+              </div>
+            {% endif %}
             <table class="table table-striped table-bordered object-table" id="item-properties-table">
               <thead>
               <tr>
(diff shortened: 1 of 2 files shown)


rob-metalinkage:
an example in action can be seen here: https://defs-dev.opengis.net/bblocks-pygeoapi/collections/ogc.bbr.examples.observation.vectorObservationFeature/items/vector-obs-1

rob-metalinkage:
also here: https://defs-dev.opengis.net/bblocks-pygeoapi/collections/ogc.osc.geodcat-stac-earthcode.experiments/items/polarwarp  

in this case you will notice that much of the STAC and records element mapping is not fully realised as Linked Data - but that is why this capability is very important as part of a testing regime.

docs/source/configuration.rst
```diff
@@ -667,6 +667,45 @@ templates using dataset level templating. To learn more about Jinja2 templates,
...
+      inject_verbatim_context: true
+
+With ``inject_verbatim_context`` enabled, both JSON and JSON-LD item resources will have
+a ``@context`` property with the provided linked data context entries, and no JSON-LD
+manipulation will be performed by pygeoapi.
+
+Additionally, some semantically enabled resources may provide their own ``@id`` (i.e., URI) value,
+which may be different from the one used by pygeoapi. The ``replace_id_field`` setting
+inside ``linked-data`` can be used to instruct pygeoapi to override a given property
+in the object with its own item URL:
```
webb-ben, line 694: How does this differ from the `uri_field` in https://github.com/geopython/pygeoapi/blob/46eaaf5f9864f6bb611cbae02584622e6c74bfb9/docs/source/configuration.rst?plain=1#L235. Is one specific to STAC and the other specific to items?

rob-metalinkage:
> How does this interact with #916

it supercedes it, however the SEO issue is a related concern that just be addressed in a less ad-hoc way than the current JSON-LD implementation

 and #1954?

this needs further analysis to determine the exact requirements left unsatisfied.  It would appear this is a general UI contract around use of JSON-links vs static defined properties with object link values.  Strongly recommend getting the basics in place with this PR before support for such extended cases are designed, however is this analysis can be done in a timely fashion it could help with the review.

> 
> This is an implementation of #1678, right?

yes

Issue #1969, opened 2025-03-18
Slow Query Performance in Postgres Provider Due to Full count on Large Tables


webb-ben:
**Description**
The Postgres provider currently performs a full .count() on every request to populate `numberMatched`, which significantly slows down queries on large tables. This makes API responses much slower than necessary, especially when querying large datasets for a small subset of the features. 

**Steps to Reproduce**
1. Set up pygeoapi with a large Postgres table as a data source.
2. Make a request to an endpoint that queries the table.
3. Observe that the response time is impacted by the .count() operation.

**Expected behavior**
Queries should return results faster by avoiding expensive .count() operations on large tables.

Potential workarounds are: 
- Introduce an environment variable to control whether the .count() is always run.
- Implement an upper limit on .count() queries, with a fallback mechanism such as TABLESAMPLE or estimated counts from Postgres' pg_stat tables.
- Use an approximate count method when exact counts are not required. This is implemented in https://github.com/cgs-earth/pygeoapi-plugins/blob/master/pygeoapi_plugins/provider/postgresql.py

**Screenshots/Tracebacks**

**Environment**
 - OS: Mac
 - Python version: 3.10
 - pygeoapi version: 0.20.dev0

**Additional context**
Similar constraints exist in other providers, with some only doing a count of features when `resulttype=hits`

jratike80:
It is impossible to make everybody happy at the same time, see https://github.com/geopython/pygeoapi/issues/1007.
Personally I prefer speed over seeing the numberMatched value, but maybe the behavior could be configured on the server side, or with a special query parameter.

mikemahoney218-usgs:
I just want to note that numberMatched is allowed to be dropped in requirement [/req/core/fc-numberMatched](https://docs.ogc.org/is/17-069r4/17-069r4.html):

> A server MAY omit this information in a response, if the information about the number of matching features is not known or difficult to compute.

So providers are permitted to drop the count entirely if it's inefficient.

mikemahoney218-usgs:
It might be nice for the sql provider to let users configure this on a collection-by-collection basis

mikemahoney218-usgs:
Wanted to bump this; there's a PR for discussion in #2174 and I do still think disabling numberMatched on a collection basis for the postgres provider would be of broad benefit

webb-ben:
I do worry about overloading the provider block but I do agree that this issue should/can be closed with the introduction of behavior like:


```yaml
        providers:  # list of 1..n required connections information
[...]

PR #2174, opened 2025-12-03
fix: add fix for slow full count in postgresql

mdearos:
# Overview

This commit adds a fix for the slow performance of full counts in PostgreSQL ([Issue #1969](https://github.com/geopython/pygeoapi/issues/1969)).

To achieve this two new PostgreSQL provider settings have been added (postgresql_pseudo_count_enabled and postgresql_pseudo_count_start). Allowing the use of pseudo counts to be configured individually on each use of the PostgreSQL provider.

This fix then uses the PostgreSQL EXPLAIN function to "guess" the number of rows that will be returned by a given request. But this does not affect all queries equally because pseudo counts cannot be run on the following types of query:
   - Requests with a Result Type of Hits.
   - Requests with a CQL filter.
   - Requests with a BBOX filter.
   - Requests with a Temporal filter.

Also, you can use the postgresql_pseudo_count_start setting to tell the system to run a full count if the row estimate is to small meaning there is enough time for a full count to be run.

This commit also adds the required documentation and postgreSQL provider test changes. Including adding a building_type and datetime column to the dummy_data.sql file.

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute bugfix Slow Query Performance in Postgres Provider Due to Full count on Large Tables to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution
- [ ] I have already previously agreed to the pygeoapi Contributions and Licensing

### DIFF
--- a/pygeoapi/provider/base.py
+++ b/pygeoapi/provider/base.py
@@ -67,7 +67,7 @@ class BaseProvider:
             raise RuntimeError('name/type/data are required')
 
         self.editable = provider_def.get('editable', False)
-        self.count = str2bool(provider_def.get('count', 'true'))
+        self.count = str2bool(provider_def.get('count', True))
         self.options = provider_def.get('options')
         self.id_field = provider_def.get('id_field')
         self.uri_field = provider_def.get('uri_field')


webb-ben:
I wonder if a postgres specific addition to the provider block is the correct approach if we want to maintain a rigid pygeoapi config schema. This appears to port some of the logic implemented in a [Psuedocount-specfic](https://github.com/cgs-earth/pygeoapi-plugins/blob/master/pygeoapi_plugins/provider/postgresql.py) pygeoapi Postgres provider. 

I wonder if there is a configuration option / solution that could be used across all pygeoapi providers given the `numberMatched` is not _required_ by the specification. Is better to have no count, or an incorrect one

mikemahoney218-usgs:
Sharing my two cents: for our deployment, we decided that no count was the better option, so none of our postgresql-backed endpoints have counts. We evaluated the same pseudocount implementation and decided that it didn't give enough benefits (both in performance and in allowing users to predict result sizes -- because the count isn't accurate, so you'll likely still need to incrementally grow your result set) versus removing `numberMatched` in line with the specification. So our deployment has a flag in the providers block of each resource, `number_matched_for_results_enabled`, controlling if /items queries return numberMatched -- as well as a matching `number_matched_for_hits_enabled` controlling if resulttype=hits is allowed at all.

Edit to add: Though unlike Ben's comment above, I do prefer controlling this on a resource level, rather than at the server level; we have other data providers that don't have the same trouble counting records (particularly for small tables) and it's nice to enable numberMatched there. We haven't had anyone comment on the inconsistency yet. We also enable resulttype=hits for most resources, and only disable it for particularly large tables.

webb-ben:
@mdearos What is the use case for needing arbitrarily high counts that _are not accurate_?

mdearos:
@webb-ben Initially we were experiencing HTTP timeouts when accessing collections with high row counts. This was happening because the process of querying the database for the whole collection, and counting the rows returned was taking too long.

Several approaches to handling this issue where considered but the decision was made to go with our proposed solution because it fulfilled our requirements of consistent API responses (a total number of results is always returned), the ability to access any collection page without receiving a HTTP timeout and providing a count of results that is as close to the true value as possible.

We have some very large collections 100 million+ rows, but customers will always filter down the data. So, the predicted values provided for collections with high row counts are used mainly to stop HTTP timeouts and give customers some idea of how much data is in the collection. But, removing them completely would mean that when a customer has filtered the dataset, they would not receive a result count at all.

Also, while the predicted value is likely to be incorrect on collections with high row counts and databases with a high number of writes. Once the vacuuming process has completed the result should be accurate even on collections with extremely high row counts. The accuracy would then degrade again over time as more data is written to the collection, but this would be corrected periodically by the vacuuming process.

tomkralidis:
@mdearos as per https://github.com/geopython/pygeoapi/issues/1969#issuecomment-3733916539, would the proposed implementation (optional counts) work for you?  If yes, would you be willing to update this PR to this effect?

If they do not, you can certainly keep this PR in your own plugin, in which case we can close this PR.

Thoughts?

mdearos:
Hi @tomkralidis ,
Thank you for your response.

I have spoken to other members of my team, and we are going to move to optional counts as the default fix for this issue. I am happy to update the PR to reflect the changes requested and I will do my best to make this available before the February 6th deadline.

I was also wondering about the PR question "I have already previously agreed to the pygeoapi Contributions and Licensing" I have not ticked this because I have not done this. I was wondering what I need to do to be able check this off?

tomkralidis:
> Hi @tomkralidis , Thank you for your response.
> 
> I have spoken to other members of my team, and we are going to move to optional counts as the default fix for this issue. I am happy to update the PR to reflect the changes requested and I will do my best to make this available before the February 6th deadline.
> 

Thanks @mdearos.  Looking forward to the updates!

> I was also wondering about the PR question "I have already previously agreed to the pygeoapi Contributions and Licensing" I have not ticked this because I have not done this. I was wondering what I need to do to be able check this off?

For a contributor's first PR, contributors typically do not check off this option.  For subsequent PRs, contributors would then check off this box.  Having said this, if anyone knows if/how GitHub Flavoured Markdown could support radio boxes (as opposed to the checkboxes we have now in that section of our PR template), any hints would be valued! :)

mdearos:
Hi @tomkralidis,
 
I was just working on the final changes to my PR when I noticed that @webb-ben has raised PR #2215 to fix this issue.
 
The proposed changes are almost identical to the changes I was going to propose so, I am not sure it is worth me continuing to edit my PR when this one appears to do the same thing and cover more providers.
 
I am happy to abandon my PR if you would like me to? Or please feel free to reject it.
 
Thanks for all your help with getting this issue resolved.
 
Also, I was wondering if you have a projected date for the release of version 0.23.0?

webb-ben:
> Hi @tomkralidis,
> 
> I was just working on the final changes to my PR when I noticed that @webb-ben has raised PR #2215 to fix this issue.
> 
> The proposed changes are almost identical to the changes I was going to propose so, I am not sure it is worth me continuing to edit my PR when this one appears to do the same thing and cover more providers.
> 
> I am happy to abandon my PR if you would like me to? Or please feel free to reject it.
> 
> Thanks for all your help with getting this issue resolved.
> 
> Also, I was wondering if you have a projected date for the release of version 0.23.0?

pygeoapi always welcome new contributors! As the are similar, why don't you go ahead with your PR.

I can rebase and edit/review as I think the implementation across different providers requires sufficient time that a PostgreSQL only contribution would be valuable. I just got ahead of myself :)

webb-ben:
I'd also welcome the current changes to `sql.py` to replace the count estimation in https://github.com/cgs-earth/pygeoapi-plugins/blob/master/pygeoapi_plugins/provider/postgresql.py#L48-L66. Much cleaner!

mdearos:
Hi @tomkralidis,

I have made the requested changes and this PR is now ready to be reviewed again.

Also, thank you for the explanation of the process around the PR question "I have already previously agreed to the pygeoapi Contributions and Licensing".

pygeoapi/provider/sql.py
```diff
@@ -127,6 +127,7 @@ def __init__(
         self.id_field = provider_def['id_field']
         self.geom = provider_def.get('geom_field', 'geom')
         self.driver_name = driver_name
+        self.count = str(provider_def.get('count', 'true')).lower() == 'true'
```
webb-ben: Suggest you move this to the base provider. Also would recommend importing/using 'pygeoapi.util.str2bool' https://github.com/geopython/pygeoapi/blob/master/pygeoapi/util.py#L230 instead of parsing the string

mdearos: Hi @webb-ben,

Thank you for your review.

I have made and pushed the changes that you requested. If you are able to take another look at my PR I would really appreciate it.

pygeoapi/provider/base.py
webb-ben, line 37: Import str2bool here

mdearos: Hi @webb-ben,

I was unable to import the str2bool function at the top of this file because it creates a circular dependency. The base.py requires util.py that requires the ProviderTypeError function from the base.py file.

Looking at this there are only really 4 ways forward:
1. Leave the import in the class scope
2. Remove the str2bool import completely and not validate the input
3. Go back to the previous way the input was validated but inside base.py
4. Move the str2bool function to its own file

Can I ask which one of these options you would prefer?

webb-ben: I will defer to Tom. I think 1, 2, or 3 are sufficient. Simplest path would be keep it as is (1).  Alternatively the most correct thing, and what would keep this PR in scope, is to move it back to the SQL provider (3) or remove str2bool validation (2) because we don't apply `str2bool` to the other bool based configuration options in the base provider. Then in #2215 we can think about how to best port into the base provider.

mdearos:
> I'd also welcome the current changes to `sql.py` to replace the count estimation in https://github.com/cgs-earth/pygeoapi-plugins/blob/master/pygeoapi_plugins/provider/postgresql.py#L48-L66. Much cleaner!

Hi @webb-ben,

I have spoken to the team and we would like to try and commit the changes to the repository mentioned. I am not 100% sure when we will be able to do this but we have a ticket in our backlog. We would like to ask some questions about the pygeoapi-plugins contribution and release processes, would it be OK to reach out to you directly via email?

tomkralidis:
> > I'd also welcome the current changes to `sql.py` to replace the count estimation in https://github.com/cgs-earth/pygeoapi-plugins/blob/master/pygeoapi_plugins/provider/postgresql.py#L48-L66. Much cleaner!
> 
> Hi @webb-ben,
> 
> I have spoken to the team and we would like to try and commit the changes to the repository mentioned. I am not 100% sure when we will be able to do this but we have a ticket in our backlog. We would like to ask some questions about the pygeoapi-plugins contribution and release processes, would it be OK to reach out to you directly via email?

Hi @mdearos: most information can be found on the [wiki](https://github.com/geopython/pygeoapi/wiki/) and [development](https://pygeoapi.io/community/psc/) section of the website.

- new/additional plugins are encouraged to be developed externally as we continue to stabilize the core of the project, and can be listed on our [Community Plugins and Themes page](https://github.com/geopython/pygeoapi/wiki/CommunityPluginsThemes)
- Release management is also described on the [wiki](https://github.com/geopython/pygeoapi/wiki/ReleaseManagement)
- Release planning is documented in https://github.com/geopython/pygeoapi/milestones

Hope this helps.

tomkralidis:
FYI #2228 should help with the import issues. @mdearos can you rebase and test the str2bool import again?  Thanks.

tomkralidis:
@mdearos looks like something happened with the rebase?  Can you review/assess/fix?

pygeoapi/provider/base.py
```diff
@@ -57,6 +57,8 @@ def __init__(self, provider_def):
         :returns: pygeoapi.provider.base.BaseProvider
         """
 
+        from pygeoapi.util import str2bool
```
tomkralidis, line 60: Can this be added as a top level import?

mdearos: Hi @tomkralidis,

I am afraid not.

When I make this change I get a circular dependency error because pygeoapi.provider.base imports str2bool from pygeoapi.util, which imports get_provider_default from pygeoapi.provider.__init__, which imports ProviderTypeError from pygeoapi.provider.base.

tomkralidis: ok, thanks @mdearos.  We can keep as is for now.

pygeoapi/provider/base.py
```diff
@@ -65,6 +67,7 @@ def __init__(self, provider_def):
             raise RuntimeError('name/type/data are required')
 
         self.editable = provider_def.get('editable', False)
+        self.count = str2bool(provider_def.get('count', 'true'))
```
tomkralidis: We can probably do `self.count = str2bool(provider_def.get('count', True))` at this stage of the code (one less cast).

tomkralidis (changes requested):

pygeoapi/provider/base.py
mdearos: Hi @tomkralidis,

I have now made this change and I have also fixed the Flake 8 issue by removing the unused str2bool import from the sql.py file.

PR #2183, opened 2025-12-15
Fix image builds for tags

hk60418:
# Overview

Currently, multi-platform (ARM64/AMD64) Docker images are only built for latest because the container workflow doesn't trigger on tag pushes. This PR adds a tag trigger so that release versions (e.g., 0.23.0) also receive multi-platform manifests. Without tagged ARM64 images, developers on Apple Silicon Macs cannot use specific versions and must either build locally or rely on latest, causing friction in development workflows.

# Related Issue / discussion

Could not find a PR relevant for this feature.

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [X] I'd like to contribute feature/bugfix to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution

I agree to the pygeoapi Contributions and Licensing Guidelines.

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -10,6 +10,9 @@ on:
     workflows: ["Build"]
     types: [completed]
     branches: [master]
+  push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
     
 env:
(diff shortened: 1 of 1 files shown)


tomkralidis:
@hk60418 in the future, as mentioned by the PR template, please open an issue for discussion and consensus.

PR #2187, opened 2025-12-16
Fix error for CSV/GeoJSON provider when the bbox parameter has 6 coordinates

### LINKED ISSUE #2186: CSV provider throws TypeError exception for a `bbox` parameter with 6 coordinates
**Description**
The `bbox` parameter allows to pass from 4 to 6 values, i.e. minx,miny,minz,maxx,maxy,maxz. When the parameter with 6 values is requested for a collection based on a CSV/GeoJSON provider then it throws the following exception:

```python
TypeError: box() takes from 4 to 5 positional arguments but 6 were given
``` 

**Steps to Reproduce**
1. Configure a collection with a CSV/GeoJSON provider
2. `curl http://localhost:5000/collections/obs/items?f=json&bbox=0.0,0.0,0.0,0.0,0.0,0.0`

**Expected behavior**
The items are served with a response `HTTP 200`

**Screenshots/Tracebacks**
NA

**Environment**
 - OS: macos Sequoia
 - Python version: 3.12
 - pygeoapi version: 0.23

**Additional context**
The affected line of code is `222` in `pygeoapi/provider/csv_.py`

francbartoli:
# Overview

# Related Issue / discussion

Closes #2186 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/tests/provider/test_csw_provider.py
+++ b/tests/provider/test_csw_provider.py
@@ -1,8 +1,10 @@
 # =================================================================
 #
 # Authors: Tom Kralidis <tomkralidis@gmail.com>
+#          Francesco Bartoli <xbartolone@gmail.com>
 #
 # Copyright (c) 2025 Tom Kralidis
+# Copyright (c) 2025 Francesco Bartoli
 #
 # Permission is hereby granted, free of charge, to any person
 # obtaining a copy of this software and associated documentation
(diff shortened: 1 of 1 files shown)


francbartoli:
@tomkralidis looking at the code more deeply this exception is returned by the `box` method from `shapely.geometry`. Beyond the points already fixed that function is used only in the `util.py` function `bbox2geojsongeometry` which is being used just by the `CSWFacadeProvider` class. I'm going to add only a test to this additional use case which already doesn't give error.

PR #2191, opened 2025-12-19
add datetime representer and allow unicode characters

KatKatKateryna:
# Overview

This PR is fixing the behavior of overwriting example-config.yaml via admin API. Currently, the data is received correctly but YAML writer is representing datetime objects and special characters incorrectly:
image

image

To fix this, YAML writer now has 'allow_unicode=True' and 'Dumper=MyDumper', where MyDumper is an instance of yaml.SafeDumper with correct datetime object representers.

# Related Issue / discussion

This was discovered while working on QGIS pygeoapi-config plugin: correctly sent data is being saved incorrectly on the running pygeoapi instance: https://github.com/byteroad/pygeoapi_config/pull/12 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/util.py
+++ b/pygeoapi/util.py
@@ -212,11 +212,7 @@ def yaml_dump(dict_: dict, destfile: str) -> bool:
         # timestamp in a specified format, without string quotes
         return dumper.represent_scalar(u'tag:yaml.org,2002:timestamp', value)
 
-    # create a dedicated dumper to write datetime to yaml correctly
-    class MyDumper(yaml.SafeDumper):
-        pass
-
-    MyDumper.add_representer(datetime, datetime_representer)
+    yaml.add_representer(datetime, datetime_representer)
 
     lock = FileLock(f'{destfile}.lock')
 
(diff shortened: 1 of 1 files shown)


doublebyte1 (approved):
Thank you @KatKatKateryna ! 🌟 

- Could you tick the box on "Updates to Public Demo"? I don't think the content of this PR is breaking anything on the demo server 👍🏽 
- please run flake8 to fix the formatting of the code:

https://docs.pygeoapi.io/en/latest/development.html#using-flake8

KatKatKateryna:
@doublebyte1  Thanks! I am not sure about the test, because the only relevant check is for 'put_config' method, but it only checks for the return code. In the end of the method, though, it does call 'admin.write(data)' where the change was made, but the affected function only writes data to YAML and never does any validation from that file. So the only way it would affect the test, is if the yaml writer fails for some reason.

KatKatKateryna:
I think the new Yaml dumper is replacing the old one (instead of adding to it), so Path representer seems to be ignored

P.S. I rerun the test locally with the latest commit 'add representer...' and these tests are passing now

kalxas:
Thank you @KatKatKateryna 
Lets see how CI goes now

tomkralidis (approved):

PR #2202, opened 2026-01-13
OpenSearch: fix property evaluation (#2189)

### LINKED ISSUE #2189: query result depends on order of parameters
As reported by @klmn-hun  in the wis2box discussions page:

_if I try to get notification messages from the API by specifying both the metadata_id and the wigos_station_identifier query parameters,
(e.g.
{WIS2BOX_API_URL/collections/messages/items?f=json&wigos_station_identifier=0-348-1-13704&metadata_id=urn%3Awmo%3Amd%3Ahu-hungaromet%3Asurface-weather-observations.gbon)
I get results also for stations with different wigosID.
On the other hand if I modify the order of the query parameters, i.e. first specify metadata_id and after that wigos_station_identifier,
I get results only for the desired station but for both datasets.
So it seems that only the last query parameter is taken into account._

_In pygeoapi package in the module `provider/elasticsearch_.py` when processing the 'properties' parameter there is a loop for the properties and after the loop it checks if prop[1] contains '|' and adds minimum_should_match ='100%' if it doesn't contain.
However this if statement should be inside the loop.
If I put the if statement inside the loop the message query I mentioned above works as expected._

See https://github.com/World-Meteorological-Organization/wis2box/discussions/985#discussioncomment-14282833

tomkralidis:
# Overview
This PR fixes an indentation bug when the OpenSearch provider evaluates query terms with pipes (|).

# Related Issue / discussion
Fixes #2189

# Additional information
Follow on of #2200 
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/opensearch_.py
+++ b/pygeoapi/provider/opensearch_.py
@@ -373,7 +373,7 @@ class OpenSearchProvider(BaseProvider):
 
             LOGGER.debug(f'Query: {query}')
             try:
-                result = self.os_search(index=self.index_name, **query)
+                result = self.os_.search(index=self.index_name, body=query)
                 if len(result['hits']['hits']) == 0:
                     LOGGER.error(err)
                     raise ProviderItemNotFoundError(err)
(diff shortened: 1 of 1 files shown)


aothompson:
Can confirm that testing this hotfix on our dev WIS2 Node running latest version of pygeoapi fixed our complex filter queries

PR #2209, opened 2026-01-15
TinyDB: fix spatial queries against null geometries

tomkralidis:
# Overview
This PR fixes TinyDB spatial queries against features whose geometry is null.

# Related Issue / discussion
https://github.com/geopython/demo.pygeoapi.io/issues/72#issuecomment-3756149186

# Additional information
None
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/tests/provider/test_filesystem_provider.py
+++ b/tests/provider/test_filesystem_provider.py
@@ -54,7 +54,7 @@ def test_query(config):
 
     r = p.get_data_path(baseurl, urlpath, dirpath)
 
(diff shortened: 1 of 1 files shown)


tomkralidis:
CI failure is unrelated/intermittent.

tomkralidis:
> it solves the case, just wondering if a try-except wouldn't be better, in case the geometry is not empty but invalid for example.

Good idea @pvgenuchten.  PR updated.

PR #2215, opened 2026-01-21
Implement optional count in Feature Providers

### LINKED ISSUE #1969: Slow Query Performance in Postgres Provider Due to Full count on Large Tables

webb-ben:
# Overview
Implement `count` field in provider def allowing providers with expensive **count** actions to skip returning a `numberMatched` value in the normal feature collection response when count is false.

# Related Issue / discussion
Closes: https://github.com/geopython/pygeoapi/issues/1969
Pending: https://github.com/geopython/pygeoapi/pull/2174

# Additional information
- Providers that do not have this added either require a count or may return it without performing an expensive count query.
- https://github.com/cgs-earth/pygeoapi-plugins/blob/master/pygeoapi_plugins/provider/postgresql.py offers a Postgis provider with a psuedocount which is faster but less accurate than a full count.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -273,6 +273,7 @@ default.
                 include_extra_query_parameters: false  # include extra query parameters that are not part of the collection properties (default: false)
                 # editable transactions: DO NOT ACTIVATE unless you have setup access control beyond pygeoapi
                 editable: true  # optional: if backend is writable, default is false
+                count: true  # optional: perform a count query for collection queries, default is true
                 # coordinate reference systems (CRS) section is optional
                 # default CRSs are http://www.opengis.net/def/crs/OGC/1.3/CRS84 (coordinates without height)
                 # and http://www.opengis.net/def/crs/OGC/1.3/CRS84h (coordinates with ellipsoidal height)
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -273,7 +273,9 @@ default.
                 include_extra_query_parameters: false  # include extra query parameters that are not part of the collection properties (default: false)
                 # editable transactions: DO NOT ACTIVATE unless you have setup access control beyond pygeoapi
                 editable: true  # optional: if backend is writable, default is false
-                count: true  # optional: perform a count query for collection queries, default is true
+                # count: include `numberMatched` in collection results responses, for providers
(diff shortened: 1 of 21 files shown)


doublebyte1 (commented):
Nice addition! 👍🏽 

Are you also planning to add the other feature providers (e.g.: elasticsearch, opensearch, parquet)?

If not, maybe you could add a column here with "optional count"?

https://github.com/geopython/pygeoapi/blob/a4dc75bf53167b2b218045faca594014cf152341/docs/source/publishing/ogcapi-features.rst?plain=1#L18-L19

tomkralidis:
> Nice addition! 👍🏽
> 
> Are you also planning to add the other feature providers (e.g.: elasticsearch, opensearch, parquet)?
> 
> If not, maybe you could add a column here with "optional count"?
> 
> https://github.com/geopython/pygeoapi/blob/a4dc75bf53167b2b218045faca594014cf152341/docs/source/publishing/ogcapi-features.rst?plain=1#L18-L19

IMHO `count` is an 'opt-in' feature (if not defined, then default).  As a rule, we should not strive for provider parity (while not breaking compatibility).

tomkralidis (approved):

PR #2224, opened 2026-01-28
Cleanup xarray provider

webb-ben:
# Overview
- Do not advertise parameters without units (alternative to #2165)
- Insert T shape in correct order
- Align xarray and xarrayEDR providers with pygeoapi code style guidelines (remove double quotes, trailing commas, etc.)

# Related Issue / discussion

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/tests/provider/test_xarray_zarr_provider.py
+++ b/tests/provider/test_xarray_zarr_provider.py
@@ -73,7 +73,7 @@ def config_no_time(tmp_path):
 def test_provider(config):
     p = XarrayProvider(config)
 
-    assert len(p.fields) == 4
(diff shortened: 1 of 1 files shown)


pygeoapi/provider/xarray_.py
```diff
@@ -126,6 +126,11 @@ def get_fields(self):
                     elif dtype.name.startswith('str'):
                         dtype = 'string'
 
+                    if not value.attrs.get('units'):
```
tomkralidis: `if not value.attrs.get('units') is None:`

PR #2229, opened 2026-02-01
RFC5: apply distance limiting

tomkralidis:
# Overview
This PR applies distance limiting per RFC5 for items and coverages.
# Related Issue / discussion
https://pygeoapi.io/development/rfc/5

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api/__init__.py
+++ b/pygeoapi/api/__init__.py
@@ -1332,3 +1332,53 @@ def evaluate_limit(requested: Union[None, int], server_limits: dict,
     else:
         LOGGER.debug('limit requested')
         return min(requested2, max_)
+
+
+def evaluate_limit_distance(server_limits: dict, collection_limits: dict,
+                            request_bbox: list = []) -> bool:
+    """
+    Helper function to evaluate limit parameter
+
+    :param server_limits: `dict` of server limits
+    :param collection_limits: `dict` of collection limits
+    :param request_bbox: `list` bbox of request
+
+    :returns: `bool` successful distance check
+    """
+
+    exceed_msg = None
+
+    effective_limits = ChainMap(collection_limits, server_limits)
+
+    on_exceed = effective_limits.get('on_exceed', 'throttle')
+    max_distance_x = effective_limits.get('max_distance_x')
+    max_distance_y = effective_limits.get('max_distance_y')
+    max_distance_units = effective_limits.get('max_distance_units')
+
+    LOGGER.debug(f'On exceed: {on_exceed}')
+    LOGGER.debug(f'Maximum distance x: {max_distance_x}')
+    LOGGER.debug(f'Maximum distance y: {max_distance_y}')
+    LOGGER.debug(f'Maximum distance units: {max_distance_units}')
+
+    # TODO: assess distance units
+
+    if not request_bbox:
+        LOGGER.debug('no bbox requested')
+        return True
+
+    if max_distance_x is not None:
+        requested_distance_x = abs(request_bbox[2] - request_bbox[0])
+        if requested_distance_x > max_distance_x:
+            exceed_msg = 'Maximum distance x exceeded'
+
+    if max_distance_y is not None:
+        requested_distance_y = abs(request_bbox[3] - request_bbox[1])
+        if requested_distance_y > max_distance_y:
+            exceed_msg = 'Maximum distance y exceeded'
+
+    if exceed_msg is not None:
+        LOGGER.warning(exceed_msg)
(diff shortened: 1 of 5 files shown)


pygeoapi/api/__init__.py
```diff
@@ -1332,3 +1332,53 @@ def evaluate_limit(requested: Union[None, int], server_limits: dict,
...
+    LOGGER.debug(f'Maximum distance units: {max_distance_units}')
+
+    # TODO: assess distance units
+
+    if not request_bbox:
+        LOGGER.debug('no bbox requested')
+        return True
+
+    if max_distance_x is not None:
+        requested_distance_x = abs(request_bbox[2] - request_bbox[0])
```
webb-ben, line 1370: It might be worth establishing the unit for distance. This as a simple comparison makes some sense when `max_distance_units` are degrees and the bbox is in `OGC:CRS84`. However if the bbox is in `EPSG:3857`, the effective limit would be _much_ smaller.

pygeoapi/api/coverages.py
```diff
@@ -112,6 +112,11 @@ def get_collection_coverage(
     else:
         try:
             bbox = validate_bbox(bbox)
+            server_limits = api.config['server'].get('limits', {})
+            collection_limits = api.config['resources'][dataset].get('limits', {})  # noqa
+
+            _ = evaluate_limit_distance(request.params.get('bbox', []),
```
webb-ben, line 118: Related to previous comment, might be worth transforming the bbox to WGS84 here until the unit is applied?

pygeoapi/api/__init__.py
```diff
@@ -1332,3 +1332,53 @@ def evaluate_limit(requested: Union[None, int], server_limits: dict,
...
+
+    on_exceed = effective_limits.get('on_exceed', 'throttle')
+    max_distance_x = effective_limits.get('max_distance_x')
+    max_distance_y = effective_limits.get('max_distance_y')
+    max_distance_units = effective_limits.get('max_distance_units')
+
+    LOGGER.debug(f'On exceed: {on_exceed}')
+    LOGGER.debug(f'Maximum distance x: {max_distance_x}')
+    LOGGER.debug(f'Maximum distance y: {max_distance_y}')
+    LOGGER.debug(f'Maximum distance units: {max_distance_units}')
```
doublebyte1, line 1361: I am a bit worried about what we can get inside the variable max_distance_units. Does it make sense to create an enum [here](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/models/provider/base.py) and validate it against it?

pygeoapi/api/__init__.py
```diff
@@ -1332,3 +1332,53 @@ def evaluate_limit(requested: Union[None, int], server_limits: dict,
...
+        LOGGER.debug('no bbox requested')
+        return True
+
+    if max_distance_x is not None:
+        requested_distance_x = abs(request_bbox[2] - request_bbox[0])
+        if requested_distance_x > max_distance_x:
+            exceed_msg = 'Maximum distance x exceeded'
+
+    if max_distance_y is not None:
+        requested_distance_y = abs(request_bbox[3] - request_bbox[1])
```
doublebyte1, line 1375: If the bbox uses degrees (e.g: CRS84/WGS84) would it be correct to apply this type of cartesian math? Not an issue for small distances, since the distortion is small; but if we you have larger distances, we could be looking at differences in the requested distance according to the location on earth. The safe solution would be to always project the bbox to a projected coordinated system (e.g.: 3857).

pygeoapi/api/__init__.py
```diff
@@ -1332,3 +1332,53 @@ def evaluate_limit(requested: Union[None, int], server_limits: dict,
...
+
+    :returns: `bool` successful distance check
+    """
+
+    exceed_msg = None
+
+    effective_limits = ChainMap(collection_limits, server_limits)
+
+    on_exceed = effective_limits.get('on_exceed', 'throttle')
+    max_distance_x = effective_limits.get('max_distance_x')
```
doublebyte1, line 1354: it may be worth casting max_distance_x and mast_distance_y to floats, and wrap this into a try/catch block

doublebyte1 (changes requested):

doublebyte1:
Not needed, but nice to have, a more complete limits object here:

https://github.com/geopython/demo.pygeoapi.io/blame/aa624319f69deeec5a9098117340b14d699414c5/services/pygeoapi_master/local.config.yml#L18-L19

PR #2243, opened 2026-02-13
Update base64 encoding to return UTF-8 string

francescoingv:
# Overview
Fixed a simple error in json_serial() not returning a string: Decode base64 encoded JSON object to UTF-8 string.

# Related Issue / discussion
No previous issue: just a code fix.
Difficult to provide a test.

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [ X] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution
- [ X] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

### DIFF
--- a/pygeoapi/util.py
+++ b/pygeoapi/util.py
@@ -378,7 +378,7 @@ def json_serial(obj: Any) -> str:
             return obj.decode('utf-8')
         except UnicodeDecodeError:
             LOGGER.debug('Returning as base64 encoded JSON object')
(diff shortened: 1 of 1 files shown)


tomkralidis:
@francescoingv we need an issue and a test to demonstrate the issue in order to consider this PR.

PR #2258, opened 2026-02-22
PubSub: add support for Kafka (#2259)

tomkralidis:
# Overview
This PR adds Kafka support for Pub/Sub workflows.
# Related Issue / discussion
#2259 

# Additional information
None
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/requirements-pubsub.txt
+++ b/requirements-pubsub.txt
@@ -1,1 +1,2 @@
+kafka-python
 paho-mqtt
--- a/docs/source/pubsub.rst
+++ b/docs/source/pubsub.rst
@@ -85,6 +85,11 @@ Brokers
 
 The following protocols are supported:
 
+.. note::
+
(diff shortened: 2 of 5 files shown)


requirements-pubsub.txt
```diff
@@ -1 +1,2 @@
+kafka-python
```
doublebyte1: Is there a reason why these packages are not in requirements-provider.txt?

doublebyte1, line 1: Not sure if we should mention somewhere `pip3 install -r requirements-pubsub.txt`?

tomkralidis: +1.  Added in docs in this PR.

tomkralidis:
> Thanks for adding this! The generation of async api works as expected.
> 
> Image
> Are you planning to add also unit tests for pubsub at some point?

Thanks @doublebyte1.  Note that some tests are implemented in `tests/api/test_pubsub.py`, most of which are focused on the "discovery" Requirements Class of the draft specification.

PR #2271, opened 2026-03-04
Geoparquet 1.1 spec compliance update

C-Loftus:
# Overview

This PR address #2252 

- If a parquet file has a bbox column, it will be filtered with using the bbox instead of needing to specify multiple x / y columns
- Added support to allow for id columns that are strings other other types. Previously it just used the row index of the data as the id value no matter what. This caused issues when the id was a string.
    - I added a warning log if the id column is non-numeric. Parquet cannot calculate row statistics as easily for strings or other non-numeric types, and as such, queries are likely to be less efficient 
- Cleaned up some of the internals; previously there were some dangling variables or unclear semantics on how the bbox filter works
- Cleaned up tests. There is one class for all naive parquet files (i.e. ones that don't follow the geoparquet spec and naively set a point column) and one for all geoparquet compliant files

# Related Issue / discussion

Closes #2252 

# Additional information

- In the future it would be useful to change all code in this provider to skip the pandas serialization step. It is generally not necessary and adds extra overhead to convert to the intermediate format. However, for the sake of not making too many changes in one PR, I've left it as is.

- I have not changed any of the code for s3fs. It is possible this could be tweaked for better performance. 

# Dependency policy (RFC2)

- [x] I have ensured that this PR meets [RFC2](https://pygeoapi.io/development/rfc/2) requirements

# Updates to public demo

- [x] I have ensured that breaking changes to the [pygeoapi master demo server](https://github.com/geopython/demo.pygeoapi.io) have been addressed
  - [x] https://github.com/geopython/demo.pygeoapi.io/blob/master/services/pygeoapi_master/local.config.yml

# Contributions and licensing

(as per https://github.com/geopython/pygeoapi/blob/master/CONTRIBUTING.md#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution
- [x] I have already previously agreed to the pygeoapi Contributions and Licensing Guidelines

PR #2150, opened 2025-10-31
Fix ParquetProvider geometry handling and add configurable geometry column support

MatthewDownsNR:
# Overview

This PR enhances the ParquetProvider with critical geometry handling fixes and adds support for configurable geometry columns, including WKT format support.

**Key improvements:**
- **Fixed critical index mismatch bug** that caused geometry to appear as `null` in individual item requests (`/collections/.../items/[id]`)
- **Added support for configurable geometry columns** via `geometry_column` parameter in provider configuration
- **Added WKT (Well-Known Text) geometry support** alongside existing WKB support
- **Enhanced geometry source prioritization** with proper fallback handling
- **Improved error handling and logging** for geometry operations
- **Added backward compatibility** for older pygeoapi versions without `crs_transform`

**Bug fixes:**
- Resolved `'list' object attribute 'index' is read-only` error in `get()` method
- Fixed geometry index alignment issues in both `query()` and `get()` methods
- Corrected bbox filtering logic for x/y field configurations

# Related Issue / discussion

This PR addresses multiple issues reported by users:
- Parquet files with existing geometry columns (WKT format) were not supported
- Individual item requests returned `null` geometry despite having valid geometry data
- Users experienced "identifier not found" errors due to geometry processing failures
- The provider only supported x/y coordinate fields, not pre-existing geometry columns

These improvements make the ParquetProvider more robust and compatible with real-world geospatial parquet files that contain geometry in various formats.

# Additional information

**Testing performed:**
- All existing 12 ParquetProvider tests continue to pass
- Added comprehensive testing for WKT geometry column support
- Verified backward compatibility with x/y field configurations
- Tested with both S3 and local file sources
- Confirmed proper handling of missing geometry scenarios

**Configuration examples:**
```yaml
# New geometry column support
providers:
  - type: feature
    name: Parquet
    data:
      source: s3://bucket/data.parquet
    geometry_column: wkt_geom  # Now supported!
    id_field: unique_id

# Existing x/y field support (unchanged)
providers:  
  - type: feature
    name: Parquet
    data:
      source: local_data.parquet
    x_field: longitude
    y_field: latitude
    id_field: id
```

**Breaking changes:** None - all changes are backward compatible.

# Dependency policy (RFC2)

- [x] I have ensured that this PR meets [RFC2](https://pygeoapi.io/development/rfc/2) requirements

# Updates to public demo

[...]

PR #2281, opened 2026-03-09
Parquet batch configuration and documentation for spec updates

C-Loftus:
# Overview

When using the parquet provider, previously there was no config option for `batch_size` or `batch_readahead`. pyarrow defaults to batch sizes of over 100k with a default readahead of 16 batches. This is fine for analytics but in a server use case it is generally far too large and when using parquet over s3, adds a lot of extra data transfer.

I changed the provider to allow these options to have smaller defaults and to allow them to be configurable in the provider definition.

I added comments for documentation for this in the code and in the rst docs. I brought some extra info for the docs from #2271  as well while I was at it.

# Related Issue / discussion

Sorta #2271 / #2150 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/docs/source/publishing/ogcapi-features.rst
+++ b/docs/source/publishing/ogcapi-features.rst
@@ -616,7 +616,7 @@ For GeoParquet data which complies to spec version 1.1, all geometry metadata wi
 detected. 
 
 Note that for any version of parquet, you may optionally specify ``batch_size`` and ``batch_readahead`` in the ``data`` section of the parquet provider config.
-``batch_size`` controls how many rows are fetched per batch. Large batch sizes speed up data processing, but add more I/O, increase latency when fetching data from an object store, and increase memory usage. If not defined it will 
+``batch_size`` controls how many rows are fetched per batch. Large batch sizes speed up data processing, but add more I/O time like increased latency when fetching data from an object store, and . If not defined it will 
 default to 20,000 rows. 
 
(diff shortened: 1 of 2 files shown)


docs/source/publishing/ogcapi-features.rst
```diff
@@ -585,7 +585,7 @@ To publish a GeoParquet file (with a geometry column) the geopandas package is a
       - type: feature
         name: Parquet
         data: 
-          source: ./tests/data/parquet/random.parquet
+          source: ./tests/data/parquet/naive/random.parquet
```
C-Loftus, line 588: updates to new test data path

pygeoapi/provider/parquet.py
```diff
@@ -136,7 +154,8 @@ def __init__(self, provider_def):
             self.fs = None
 
         # Build pyarrow dataset pointing to the data
-        self.ds = pyarrow.dataset.dataset(self.source, filesystem=self.fs)
+        self.ds: pyarrow.dataset.Dataset = \
+            pyarrow.dataset.dataset(self.source, filesystem=self.fs)
```
C-Loftus, line 158: since pyarrow has multiple different objects that could potentially be used in the internals of this provider. (i.e. `pyarrow.parquet`, it is helpful for documentation purposes to explicitly label it). Also helps for type validation in editors with lsp. Otherwise self.ds is assumed to be unknown since pyarrow doesn't provide type info.

Issue #1547, opened 2024-02-23
Use Protocol interfaces and static duck typing design pattern


francbartoli:
**Is your feature request related to a problem? Please describe.**
In order to improve flexibility and composability over inheritance it would be beneficial to improve the design using static duck typing  [PEP 544](https://peps.python.org/pep-0544/)

**Describe the solution you'd like**
This is a very basic example how we can use it to adopt standard dataclasses as replacement of pydantic data models but at the same type leave the flexibility to use them for data validation in a fancy way without touching the core:

```python
from typing import Any, Protocol

from pydantic import RootModel
from pydantic.dataclasses import dataclass as p_dataclass
from dataclasses import dataclass

from pygeoapi.models.openapi import SupportedFormats


class OAPIFormat(Protocol):
    """Interface for OpenAPI Format"""

    root: SupportedFormats


class OAPIPydantic(RootModel):
    root: SupportedFormats = SupportedFormats.YAML.value


@dataclass
class OAPI(OAPIFormat):
    """OpenAPI definition"""

    root:SupportedFormats


@p_dataclass
class OAPI_PydanticDataclass(OAPIFormat):
    """OpenAPI definition"""
    root:SupportedFormats


def check_openapi_format(format: OAPIFormat):
    """Check OpenAPI Format"""
    print(format)
```

**Describe alternatives you've considered**

**Additional context**

vvmruder:
I would be interested to see the move to plain dataclasses as well. Since they are part of python this should be more stable than using pydantic. Currently I'm facing the same problem as described here #1664

Is there a roadmap when this could happen? I could support working on that.

vvmruder:
@francbartoli I really appriceate this decision and lokking forward to use it.

Again, I can offer a hand to support the work on that. Please get in touch.

PR #2307, opened 2026-03-28
Get rid of pydantic and move data validation to standard dataclasses

francbartoli:
# Overview

This PR aims at removing `pydantic` from the core dependencies in favor of the standard `dataclasses` while offering the same level of data validation in the codebase. At the same time it improves the extensibility of the plugin system in order to write custom plugins for data validation with `pydantic` or other external validation libraries.
A new `PluginContext` dataclass has been added to pass dependencies in the plugin system. Downstream projects/forks can extend `PluginContext` via subclassing to integrate a custom validation logic.
The change has a minimal footprint on the plugin design because it guarantees backwards compatibility so existing pygeoapi instances can update with zero-risk impact.

# Related Issue / discussion

Mainly discussed and agreed during the [summit ](https://github.com/geopython/pygeoapi/wiki/Code-Sprint-2024-02-29-to-2024-03-01#notes) with the ultimate goal to get rid of `pydantic` and provide downstream data validation with custom pydantic-based (or similar tools) plugins

https://github.com/geopython/pygeoapi/issues/1547
https://github.com/geopython/pygeoapi/issues/1664

# Additional information

A new [RFC](https://pygeoapi.io/development/rfc/6/) has been raised to voting about this design change

The plan to achieve the above goal is the following:

- [X] Extend the `load_plugin` in `plugin.py` method with an optional context based on a `PluginContext` dataclass
- [X] Refactor `BaseProvider`, `BaseProcessor` and `BaseManager` for backwards compatibility
- [X] Migrate `APIRules` to dataclass
- [X] Migrate `TileSetMetadata` to dataclass
- [X] Migrate `MVTTilesJson` to dataclass
- [X] Remove `Pydantic` from the dependencies

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@
 #
 # Copyright (c) 2026 Tom Kralidis
 # Copyright (c) 2019 Just van den Broecke
-# Copyright (c) 2025 Francesco Bartoli
+# Copyright (c) 2026 Francesco Bartoli
 # Copyright (c) 2025 Angelos Tzotsos
 # Copyright (c) 2023 Bernhard Mallinger
 #
@@ -70,7 +70,6 @@ ARG ADD_DEB_PACKAGES="\
     python3-netcdf4 \
     python3-pandas \
     python3-psycopg2 \
-    python3-pydantic \
     python3-pymongo \
     python3-pyproj \
     python3-rasterio \
--- a/docs/source/plugins.rst
+++ b/docs/source/plugins.rst
@@ -502,6 +502,216 @@ The below template provides a minimal example (let's call the file ``mycooljsonf
            return out_data
 
 
+Example: transaction validation with PluginContext
+--------------------------------------------------
+
+pygeoapi serves the schema of each collection at ``/collections/{id}/schema``, but
+when a transaction (create/update) comes in, providers write the data without validating
+it against that schema.
+
+If you want to fill this gap then you should write a dedicated plugin, i.e. a
+``ValidatedGeoJSONProvider`` plugin for the ``GeoJSON`` provider.  It reads the provider's
+JSON Schema (from ``get_fields()``) and builds a validator that checks incoming features
+on ``create()`` and ``update()``.  In python there are plenty of data validation libraries
+but pygeoapi aims at being dependency least so at first glance the provider code should be
+**technology-agnostic**: it calls the validator's interface without knowing whether the
+validator is implemented with dataclasses, pydantic, or any other library.
+
+The bare minimal, without injecting any ``PluginContext``, is to build a new provider
+(or add the validation logic to the existing one) using the standard python and the
+existing dependencies to validate from its own fields.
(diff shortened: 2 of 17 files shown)


francbartoli:
CI failures are not related to these changes

webb-ben:
Its not entirely clear to me how the introduction of PluginContext is coupled this closesly with the removal of pydantic. Furthermore I am not quite sure I understand the intended utility of it - perhaps this is a documentation question. Where / How would this be used?

francbartoli:
> Its not entirely clear to me how the introduction of PluginContext is coupled this closesly with the removal of pydantic. Furthermore I am not quite sure I understand the intended utility of it - perhaps this is a documentation question. Where / How would this be used?

That's a very good question @webb-ben, I'm happy to document what's the intended usage downstream indeed

francbartoli:
@webb-ben I've added an example about a potential use case for data validation with two variants, one with a default validator (standard library only) and the second with a downstream override (`pydantic` via `ValidatingContext`). 
Let me also clarify why  `PluginContext` and `pydantic` removal are in the same PR.

They are technically independent but share the same design goal: making the plugin system more flexible for downstream projects.

In the new example there is a `ValidatedGeoJSONProvider` that explains how the `PluginContext` would be used. In short, I would consider these key points:

1. The provider is *technology-agnostic*: it calls `self._feature_validator(**properties)` on `create()`/`update()` without knowing whether the validator is a dataclass, a pydantic model, or any other callable. The only contract is: accept `*kwargs`, raise on invalid input. 

2. Without `PluginContext` (pygeoapi core today), the current provider could be improved with a default validator from its own JSON Schema (`provider.fields`) using only dataclasses and `validate_type` from pygeoapi core, no external dependencies. This catches type errors on transactions that pygeoapi does not catch today. For example, a `POST` with `"scalerank": "not_a_number"` to a collection where `scalerank` is declared as `integer` is currently accepted and written to disk silently — this is just an example not in scope for this PR.

3. With PluginContext (downstream), a project that uses pygeoapi as a library can subclass `PluginContext`, inject a stricter validator (e.g. a pydantic `BaseModel` with value ranges, allowed values, required fields), and the same provider code applies those rules without any modification. The validation logic is decoupled from the provider implementation.

Again, this is just one of the possible use cases. The pattern would be the same as `self.logger` in `BaseProvider`: sensible default when no context is provided, override via injection when it is.

mikemahoney218-usgs:
Apologies if I'm way off here, as I might be missing the context (heh) of this change -- but would it be easier and more flexible to just optionally pass the existing APIContext to plugins instead? I'm looking at the fields that PluginContext defines and I don't believe our providers would make use of any of them -- but having access to the query arguments, for instance, would simplify a lot of our workflow. I _believe_ that would still accomplish what you're looking for as well without requiring the new class -- but again, I'd understand if I'm way off!

PR #2308, opened 2026-03-30
Updates to the OGC API - Maps Support

doublebyte1:
# Overview

- Introduced some guardrails for WebMercator, to prevent issues with out of bounds. 
- Since the WMS Facade does not support a bbox-crs different than the crs, there is no need to pass a bbox-crs to the provider; however we must ensure that we always pass the bbox in crs coordinates (transform, if needed).
- Support default values for crs, bbox-crs and bbox.
- Add support for storage_crs, as a configuration option.
- Add Content-Crs and Content-Bbox headers to the response
- Catchup with tests and documentation

This PR also addresses a current issue with the map preview on the collection page, which affects the demo server.

# Additional information

In the collection page, it toggles between imageoverlay and imageoverlay.ocgapi to show the preview map, even when the view is outside the limits of the crs:

Screenshot from 2026-05-05 08-58-41

Screenshot from 2026-05-05 08-58-53

Screenshot from 2026-03-31 09-14-34

TODOS:
- [x]  update unit tests
- [x]  update docs

# Dependency policy (RFC2)

# Updates to public demo

I created an additional PR [here](https://github.com/geopython/demo.pygeoapi.io/pull/78), to add the storage_crs configuration option.

# Contributions and licensing

### DIFF
--- a/pygeoapi/api/maps.py
+++ b/pygeoapi/api/maps.py
@@ -8,6 +8,7 @@
 #          Ricardo Garcia Silva <ricardo.garcia.silva@geobeyond.it>
 #          Bernhard Mallinger <bernhard.mallinger@eox.at>
 #
+# Copyright (c) 2026 Joana Simoes
 # Copyright (c) 2026 Tom Kralidis
 # Copyright (c) 2025 Francesco Bartoli
 # Copyright (c) 2022 John A Stevenson and Colin Blackburn
@@ -61,15 +62,19 @@ CONFORMANCE_CLASSES = [
     'http://www.opengis.net/spec/ogcapi-maps-1/1.0/conf/core'
 ]
 
-DEFAULT_CRS = 'http://www.opengis.net/def/crs/EPSG/0/4326'
+DEFAULT_CRS = 'http://www.opengis.net/def/crs/OGC/1.3/CRS84'
+
+DEFAULT_BBOX = [-180, -90, 180, 90]  # CRS84
 
 CRS_CODES = {
     '4326': 'http://www.opengis.net/def/crs/EPSG/0/4326',
     '3857': 'http://www.opengis.net/def/crs/EPSG/0/3857',
+    'CRS84': 'http://www.opengis.net/def/crs/OGC/1.3/CRS84',
     'http://www.opengis.net/def/crs/EPSG/0/4326': 'http://www.opengis.net/def/crs/EPSG/0/4326',  # noqa
     'http://www.opengis.net/def/crs/EPSG/0/3857': 'http://www.opengis.net/def/crs/EPSG/0/3857',  # noqa
+    'http://www.opengis.net/def/crs/OGC/1.3/CRS84': 'http://www.opengis.net/def/crs/OGC/1.3/CRS84', # noqa
     'EPSG:4326': 'http://www.opengis.net/def/crs/EPSG/0/4326',
-    'EPSG:3857': 'http://www.opengis.net/def/crs/EPSG/0/3857'
+    'EPSG:3857': 'http://www.opengis.net/def/crs/EPSG/0/3857',
 }
 
 
@@ -115,18 +120,31 @@ def get_collection_map(api: API, request: APIRequest,
     query_args['format_'] = request.params.get('f', 'png')
     query_args['style'] = style
 
-    query_args['crs'] = CRS_CODES.get(request.params.get(
-        'crs', DEFAULT_CRS))
-    query_args['bbox_crs'] = CRS_CODES.get(request.params.get(
-        'bbox-crs',
-        api.config['resources'][dataset]['extents']['spatial'].get(
-            'crs', DEFAULT_CRS))
-    )
-
-    if query_args['crs'] is None:
+    # If there is no crs param, we assume the storage_crs;
+    # if it does not exist or is not supported, use CRS84.
+    try:
+        if 'crs' not in request.params:
+            query_args['crs'] = CRS_CODES.get(collection_def.get('storage_crs',
+                                              DEFAULT_CRS), DEFAULT_CRS)
+        else:
+            query_args['crs'] = CRS_CODES.get(request.params['crs'],
+                                              DEFAULT_CRS)
+    except KeyError:
         query_args['crs'] = DEFAULT_CRS
-    if query_args['bbox_crs'] is None:
-        query_args['bbox_crs'] = DEFAULT_CRS
+
+    LOGGER.debug(f'Using crs: {query_args['crs']}')
+
+    # If there is no bbox-crs param, we assume CRS84
+    try:
+        if 'bbox-crs' not in request.params:
+            query_args['bbox-crs'] = DEFAULT_CRS
+        else:
+            query_args['bbox-crs'] = CRS_CODES.get(request.params['bbox-crs'],
+                                                   DEFAULT_CRS)
+    except KeyError:
+        query_args['bbox-crs'] = DEFAULT_CRS
+
+    LOGGER.debug(f'Using bbox-crs: {query_args['bbox-crs']}')
 
     try:
         query_args['width'] = int(request.params.get('width', 500))
@@ -156,7 +174,7 @@ def get_collection_map(api: API, request: APIRequest,
                 exception, api.pretty_print)
 
     except AttributeError:
-        bbox = api.config['resources'][dataset]['extents']['spatial']['bbox']  # noqa
+        bbox = DEFAULT_BBOX
     try:
         bbox = [float(c) for c in bbox]
     except ValueError:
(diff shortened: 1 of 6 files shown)


webb-ben:
Should the bbox-crs not be used to convert the bbox into the requested crs?

doublebyte1:
> Should the bbox-crs not be used to convert the bbox into the requested crs?

Whenever it exists, yes; but this is an optional parameter, so we should have a fallback if we want to support a minimal map query (e.g.: no crs, no bbox-crs, no bbox).

https://github.com/doublebyte1/pygeoapi/blob/b581b8cd93bf8691a5adee34c7ce8fd425784ae0/pygeoapi/api/maps.py#L120-129

webb-ben:
Makes some sense to me. Is there a reason to derive the default CRS from bbox or should we align this with crs list and storage_crs as implemented for OAFeat?

doublebyte1:
> Makes some sense to me. Is there a reason to derive the default CRS from bbox or should we align this with crs list and storage_crs as implemented for OAFeat?

If we don' t have bbox-crs and bbox as parameters, we still need some defaults to provide a map to the user. The defaults are taken from what is defined on the spatial extents of the collections (bbox and bbox-crs), but I am open to more ideas (:

Examples:
https://demo.pygeoapi.io/master/collections/mapserver_world_map/map?f=png&crs=4326
https://demo.pygeoapi.io/master/collections/mapserver_world_map/map?f=png

If the spatial extents of the collection are not defined in the config, I also defined some defaults. But I am actually not sure, if that is even possible?

In the case of our WMS providers, the storage_crs make less sense, because they are basically facades.

webb-ben:
Ahh. Using bbox and bbox-crs in that context makes sense. I was more thinking along the lines of conformance with https://docs.pygeoapi.io/en/latest/crs.html

webb-ben:
Does the map emit its bbox and CRS? It is my understand that to know how to draw a map the client would need to know the extents of the map which I presume is done by the client dictating to the server these values.

Is `/map` without query args a special case where OAMap allows for 200 response without overloading the query args or is there some other way the projected CRS and extent of the map are returned to a client?

doublebyte1:
> Does the map emit its bbox and CRS? It is my understand that to know how to draw a map the client would need to know the extents of the map which I presume is done by the client dictating to the server these values.
> 
> Is `/map` without query args a special case where OAMap allows for 200 response without overloading the query args or is there some other way the projected CRS and extent of the map are returned to a client?

I see your point now. No, we do not give that information back to the client. This is what [is written in the Standard](https://docs.ogc.org/is/20-058/20-058.html):

`NOTE 1:    When no crs parameters are specified, please refer to the Maps API core conformance class to know about the default CRS`

`NOTE 2:    The default CRS of the BBOX is https://www.opengis.net/def/crs/OGC/1.3/CRS84 but the default CRS of the map is the native (storage) CRS`.

Also, the bounding box is not optional. I need to update the code to reflect that, and the fact that the default crs of the bounding box is CRS84, rather than 4326. Regarding the native storage crs, I guess we need the user to declare it in the config? there is a layer crs in the GetCapabilities of the WMS service.

doublebyte1:
> Does the map emit its bbox and CRS? It is my understand that to know how to draw a map the client would need to know the extents of the map which I presume is done by the client dictating to the server these values.
> 
> Is `/map` without query args a special case where OAMap allows for 200 response without overloading the query args or is there some other way the projected CRS and extent of the map are returned to a client?

You were right in your observations (: There is a map collection `storage_crs` in the Standard, and that should be the default crs; in case it **does not exist**, the server should default to CRS84 .

oam-crs

In the case of `bbox-crs`, it should just default to CRS84.

oam-bbox-crs

Ultimately, the  crs and the bounding box are returned to the user in the response headers.

```
Content-Crs: http://www.opengis.net/def/crs/EPSG/0/4326
Content-Bbox: -180.0,-90.0,180.0,90.0
```

I reflected this logic in my latest commits.

pygeoapi/api/maps.py
```diff
@@ -61,15 +62,19 @@
     'http://www.opengis.net/spec/ogcapi-maps-1/1.0/conf/core'
 ]
 
-DEFAULT_CRS = 'http://www.opengis.net/def/crs/EPSG/0/4326'
+DEFAULT_CRS = 'http://www.opengis.net/def/crs/OGC/1.3/CRS84'
```
webb-ben: Can these be a part of `pygeoapi.crs`? We have the concept of `DEFAULT_STORAGE_CRS` and `DEFAULT_CRS` already.

pygeoapi/api/maps.py
```diff
@@ -114,11 +119,32 @@ def get_collection_map(api: API, request: APIRequest,
 
     query_args['format_'] = request.params.get('f', 'png')
     query_args['style'] = style
-    query_args['crs'] = CRS_CODES[request.params.get(
-        'crs', collection_def.get('crs', DEFAULT_CRS))]
-    query_args['bbox_crs'] = CRS_CODES[request.params.get(
-        'bbox-crs', collection_def.get('crs', DEFAULT_CRS))]
-    query_args['transparent'] = request.params.get('transparent', True)
```
webb-ben, line 121: Keep transparency arg parse?

doublebyte1:
CI failures seem unrelated

pygeoapi/api/maps.py
```diff
@@ -160,10 +188,10 @@ def get_collection_map(api: API, request: APIRequest,
             exception, api.pretty_print)
 
     # the transformer function expects the crs to be in a uri format
-    if query_args['bbox_crs'] != query_args['crs']:
+    if query_args['bbox-crs'] != query_args['crs']:
         LOGGER.debug(f'Reprojecting bbox CRS: {query_args["crs"]}')
-        bbox = transform_bbox(bbox, query_args['bbox_crs'], query_args['crs'])
-
+        bbox = transform_bbox(bbox, query_args['bbox-crs'],
+                              query_args['crs'], True)
```
webb-ben:
```suggestion
                              query_args['crs'], always_xy=True)
```
Might be worth keeping this as a keyword argument to maintain clarity about what is happening here

pygeoapi/provider/wms_facade.py
```diff
@@ -43,10 +45,11 @@
 
 CRS_CODES = {
     'http://www.opengis.net/def/crs/EPSG/0/4326': 'EPSG:4326',
-    'http://www.opengis.net/def/crs/EPSG/0/3857': 'EPSG:3857'
+    'http://www.opengis.net/def/crs/EPSG/0/3857': 'EPSG:3857',
+    'http://www.opengis.net/def/crs/OGC/1.3/CRS84': 'urn:ogc:def:crs:OGC:1.3:CRS84' # noqa
```
webb-ben: I get an error when I try to use a URN for WMS. The error says the CRS should be in the format `AUTHORITY:CODE`

doublebyte1:
@webb-ben Thanks a lot for the quick review! 👍🏽 Those are all great comments, addressed in my latest commits.

webb-ben (approved):
+1. Wonder if we could just draw the image with default bbox on the map once for those lower zooms. Starts to become a whole hack though. 

```
diff --git a/pygeoapi/templates/collections/collection.html b/pygeoapi/templates/collections/collection.html
index 1d522c3..887eebe 100644
--- a/pygeoapi/templates/collections/collection.html
+++ b/pygeoapi/templates/collections/collection.html
@@ -169,6 +169,7 @@
     var clampedBounds = L.latLngBounds(clampedSw, clampedNe);
 
     var ogcapi_layer = null;
+    var image_layer = null;
 
     {# if this collection has a map representation, add it to the map #}
     {% for link in data['links'] %}
@@ -179,12 +180,12 @@
             "transparent": true, 
             "bounds": clampedBounds 
         });
+        image_layer = L.imageOverlay("{{ link['href'] }}", clampedBounds, {opacity: 0.7, transparent: true});
         bbox_layer.setStyle({
             fillOpacity: 0
         });
       {% endif %}
     {% endfor %}
-
     // Check bounds and toggle the visibility of the imageOverlay, accordingly
     function toggleOverlayVisibility() {
 
@@ -199,9 +200,9 @@
         var isWithinBounds = (viewWidth <= 360) && (centerLng >= -180 && centerLng <= 180);
 
             if (isWithinBounds) {
-                if (!map.hasLayer(ogcapi_layer)) map.addLayer(ogcapi_layer);
+                if (!map.hasLayer(ogcapi_layer)) map.addLayer(ogcapi_layer); map.removeLayer(image_layer);
             } else {
-                if (map.hasLayer(ogcapi_layer)) map.removeLayer(ogcapi_layer);
+                if (map.hasLayer(ogcapi_layer)) map.removeLayer(ogcapi_layer); map.addLayer(image_layer);
             }
         }
     }
```

pygeoapi/templates/collections/collection.html
```diff
@@ -179,6 +180,10 @@ <h3>{% trans %}Storage CRS{% endtrans %}</h3>
             "transparent": true, 
             "bounds": clampedBounds 
         });
+        image_layer = L.imageOverlay("{{ link['href'] }}", clampedBounds, {opacity: 0.7, transparent: true});
+         bbox_layer.setStyle({
+             fillOpacity: 0
+         });
         bbox_layer.setStyle({
             fillOpacity: 0
         });
```
webb-ben, line 189: Duplication of the bbox fill

tomkralidis (approved):

PR #2317, opened 2026-04-16
catch EDR instance queries in Starlette

tomkralidis:
# Overview
This PR handles Starlette greedy routing for EDR instance queries.

# Related Issue / discussion

# Additional information
None
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/starlette_app.py
+++ b/pygeoapi/starlette_app.py
@@ -523,6 +523,11 @@ async def get_collection_edr_query(request: Request,
     if 'collection_id' in request.path_params:
         collection_id = request.path_params['collection_id']
 
+        if '/instances/' in collection_id:
+            tokens = collection_id.split('/instances/')
+            collection_id = tokens[0]
+            instance_id = tokens[-1]
+
     if 'instance_id' in request.path_params:
         instance_id = request.path_params['instance_id']
 


pygeoapi/starlette_app.py
```diff
@@ -523,6 +523,11 @@ async def get_collection_edr_query(request: Request,
     if 'collection_id' in request.path_params:
         collection_id = request.path_params['collection_id']
 
+        if '/instances/' in collection_id:
+            tokens = collection_id.split('/instances/')
+            collection_id = tokens[0]
+            instance_id = tokens[-1]
+
     if 'instance_id' in request.path_params:
         instance_id = request.path_params['instance_id']
```
webb-ben, line 532: Does this function make use of `collection_id`, `instance_id`, `location_id`? or are they provided as expected inputs that L523-532 resolves from the request parameters?

webb-ben: Moreover, what is the circumstance in which `instance_id` is derived from this PR instead of the subsequent lines that parse the Route params for instance_id
```python
    if 'instance_id' in request.path_params:
        instance_id = request.path_params['instance_id']
```

tomkralidis: > Moreover, what is the circumstance in which `instance_id` is derived from this PR instead of the subsequent lines that parse the Route params for instance_id
> 
> ```python
>     if 'instance_id' in request.path_params:
>         instance_id = request.path_params['instance_id']
> ```

`instance_id` is utilizied in the above when the route is `/collections/{collection_id}/instances/{instance_id}`.

When the route is `/collections/{collection_id}/instances/{instance_id}/position` (or other `instance_id` based query types), `instance_id` is not picked up, which is what this PR fixes for the EDR query type routing.

webb-ben: Ahh:
`path returns the rest of the path, including any additional / characters.`

webb-ben (approved):

PR #2153, opened 2025-11-04
Fix STAC API endpoints on Starlette

rajadain:
# Overview

This is in support of the USGS NHGF work with @dblodgett-usgs and @sjordan29.

These endpoints were added recently in #2129, but without the `await`, we get the following error:

```
Traceback (most recent call last):
  File "/python/conda/lib/python3.12/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/python/conda/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/python/conda/lib/python3.12/site-packages/starlette/applications.py", line 113, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/python/conda/lib/python3.12/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/python/conda/lib/python3.12/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/python/conda/lib/python3.12/site-packages/starlette/middleware/cors.py", line 85, in __call__
    await self.app(scope, receive, send)
  File "/python/conda/lib/python3.12/site-packages/starlette/middleware/exceptions.py", line 63, in __call__
    await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
  File "/python/conda/lib/python3.12/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    raise exc
  File "/python/conda/lib/python3.12/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
    await app(scope, receive, sender)
  File "/python/conda/lib/python3.12/site-packages/starlette/routing.py", line 716, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/python/conda/lib/python3.12/site-packages/starlette/routing.py", line 736, in app
    await route.handle(scope, receive, send)
  File "/python/conda/lib/python3.12/site-packages/starlette/routing.py", line 462, in handle
    await self.app(scope, receive, send)
  File "/python/conda/lib/python3.12/site-packages/starlette/routing.py", line 716, in __call__
[...]

PR #2340, opened 2026-05-11
Further Fixes for STAC Search

### LINKED ISSUE #2354: Bug: STAC Item Search Breaks
**Description**

When attempting to search STAC items, the API breaks.

**Steps to Reproduce**

- Initialize a STAC search using pystac_client
- Execute `search.item_collection()`
- See the failure

**Expected behavior**
The search succeeds and returns results.

**Screenshots/Tracebacks**

Failure 1, because of double slash on this line:

https://github.com/geopython/pygeoapi/blob/0c756e1ecd3ca4e70d7bd7a0874bd4503ecd9975/pygeoapi/api/stac.py#L287

Image

Failure 2, because assets are treated as lists, not dicts:

Image

The [STAC item spec specifies](https://stacspec.org/en/about/stac-spec/) that links are lists, and assets are dicts:

Image

**Environment**
 - OS: Linux via Docker running on macOS
 - Python version: 12
 - pygeoapi version: 0.23.4

### LINKED ISSUE #2355: Bug: Links and Assets not sourced correctly from pgSTAC
**Description**
pygeopapi expects links and assets to be top-level items, as specified in [the STAC spec](https://stacspec.org/en/about/stac-spec/).

However, in pgSTAC, the collections and items schemas do not include them as top-level items, but store them in the `contents` JSONB column:

https://github.com/stac-utils/pgstac/blob/08ce5991e7840f51993d508c715fad44f889e5e5/src/pgstac/pgstac.sql#L889-L899

https://github.com/stac-utils/pgstac/blob/08ce5991e7840f51993d508c715fad44f889e5e5/src/pgstac/pgstac.sql#L2067-L2077

To access them when using a pgSTAC back-end, we should fetch from `feature.properties.contents.links` and `feature.properties.contents.assets` here:

https://github.com/geopython/pygeoapi/blob/0c756e1ecd3ca4e70d7bd7a0874bd4503ecd9975/pygeoapi/api/stac.py#L382-L384

**Steps to Reproduce**

- Setup pystac_client against a pygeoapi instance backed by pgSTAC
- Run `search.item_collection()`
- Observe the links and assets are incomplete and missing from the output:

Image

**Expected behavior**
The links and assets are sourced correctly from the pgSTAC backend:

Image

**Environment**
 - OS: Linux via Docker running on macOS
 - Python version: 12
 - pygeoapi version: 0.23.4

### LINKED ISSUE #2356: Bug: Instantaneous STAC datetime not correctly initialized
**Description**
Currently, pygeopapi checks if `start_datetime` and `end_datetime` exist, and if they do, initializes a temporal range. However, if they do not, but `datetime` does exist, it does not correctly initialize a temporal instant:

https://github.com/geopython/pygeoapi/blob/0c756e1ecd3ca4e70d7bd7a0874bd4503ecd9975/pygeoapi/api/stac.py#L484-L491

Because of this lack of initialization, this value is later overridden by other values:

https://github.com/geopython/pygeoapi/blob/0c756e1ecd3ca4e70d7bd7a0874bd4503ecd9975/pygeoapi/api/stac.py#L493-L506

We should use the `datetime` if available, and only override if not.

**Environment**
 - OS: Linux via Docker running on macOS
 - Python version: 12
 - pygeoapi version: 0.23.4

rajadain:
# Overview

Minor fixes that allow the STAC search endpoint to work correctly. This is in support of the USGS NHGF work.

# Related Issue / discussion

This is a continuation of https://github.com/geopython/pygeoapi/pull/2153.

Closes #2354 
~~Closes #2355~~
Closes #2356 

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/api/stac.py
+++ b/pygeoapi/api/stac.py
@@ -379,9 +379,11 @@ def search(api: API, request: Union[APIRequest, Any]) -> Tuple[dict, int, str]:
                     geom = from_geojson(json.dumps(feature['geometry']))
                     feature['bbox'] = geom.bounds
 
-                for la in ['links', 'assets']:
-                    if feature.get(la) is None:
-                        feature[la] = []
+                if feature.get('links') is None:
+                    feature['links'] = []
+
+                if feature.get('assets') is None:
+                    feature['assets'] = {}
 
     stac_api_response['numberReturned'] = len(stac_api_response['features'])
 
--- a/pygeoapi/api/stac.py
+++ b/pygeoapi/api/stac.py
@@ -284,7 +284,7 @@ def landing_page(api: API,
         'rel': 'search',
         'type': FORMAT_TYPES[F_JSON],
         'title': l10n.translate('STAC API search', request.locale),
-        'href': f"{api.base_url}/stac-api//search?f={F_JSON}"
+        'href': f"{api.base_url}/stac-api/search?f={F_JSON}"
     }]
 
     return headers, status, to_json(content, api.pretty_print)
--- a/pygeoapi/api/stac.py
+++ b/pygeoapi/api/stac.py
@@ -490,18 +490,20 @@ def get_temporal(feature: dict) -> dict:
     if datetime_ is None and None not in [start_datetime, end_datetime]:
         LOGGER.debug('Temporal range partially exists')
     elif datetime_ is not None:
+        value['datetime'] = datetime_
         LOGGER.debug('Temporal instant exists')
 
     LOGGER.debug('Attempting to derive temporal from GeoJSON feature')
     LOGGER.debug(feature)
-    if feature.get('time') is not None:
+    if value.get('datetime') is None and feature.get('time') is not None:
         if feature['time'].get('timestamp') is not None:
             value['datetime'] = feature['time']['timestamp']
         if feature['time'].get('interval') is not None:
             value['start_datetime'] = feature['time']['interval'][0]
             value['end_datetime'] = feature['time']['interval'][1]
 
-    if feature['properties'].get('created') is not None:
+    if value.get('datetime') is None \
+       and feature['properties'].get('created') is not None:
         value['datetime'] = feature['properties']['created']
 
     if not value:


pygeoapi/api/stac.py
```diff
@@ -379,9 +379,17 @@ def search(api: API, request: Union[APIRequest, Any]) -> Tuple[dict, int, str]:
                     geom = from_geojson(json.dumps(feature['geometry']))
                     feature['bbox'] = geom.bounds
 
-                for la in ['links', 'assets']:
-                    if feature.get(la) is None:
-                        feature[la] = []
+                if feature.get('links') is None:
```
tomkralidis, line 382: What is the content model that this update is based on?  Is `properties.content.links` and `properties.content.assets` part of STAC proper?

pygeoapi/api/stac.py
```diff
@@ -488,18 +496,20 @@ def get_temporal(feature: dict) -> dict:
     if datetime_ is None and None not in [start_datetime, end_datetime]:
         LOGGER.debug('Temporal range partially exists')
     elif datetime_ is not None:
+        value['datetime'] = datetime_
         LOGGER.debug('Temporal instant exists')
 
     LOGGER.debug('Attempting to derive temporal from GeoJSON feature')
     LOGGER.debug(feature)
-    if feature.get('time') is not None:
+    if not value.get('datetime') and feature.get('time') is not None:
```
tomkralidis: Can we do `if value.get('datetime') is None and feature.get('time') is not None:` for readability?

pygeoapi/api/stac.py
```diff
@@ -488,18 +496,20 @@ def get_temporal(feature: dict) -> dict:
...
         if feature['time'].get('timestamp') is not None:
             value['datetime'] = feature['time']['timestamp']
         if feature['time'].get('interval') is not None:
             value['start_datetime'] = feature['time']['interval'][0]
             value['end_datetime'] = feature['time']['interval'][1]
 
-    if feature['properties'].get('created') is not None:
+    if not value.get('datetime') \
```
tomkralidis: Can we do `if value.get('datetime') is None and feature['properties'].get('created') is not None:` for readability?

tomkralidis (changes requested):
See change requests.  It would be valuable to clarify the problems the PR attempts to address/fix (and/or feel free to create an issue).  Thanks!

rajadain:
Thank you for taking a quick look! I'll add more context and motivation for the changes, and make the readability improvements today. 🙏

pygeoapi/api/stac.py
rajadain: Done in 72e7a7f

pygeoapi/api/stac.py
rajadain: Done in 72e7a7f

pygeoapi/api/stac.py
rajadain: Yes. As can be seen in [the STAC Spec](https://stacspec.org/en/about/stac-spec/):

image

`links` is a list, and `assets` is a dict.

rajadain:
Apologies for the long delay in getting back to this.

I created cards specifying the individual issues we need to fix: #2354, #2355, #2356. I've also marked this PR as closing those bugs. I've also added more detail to the individual commits, better explaining their relevance and motivation. And I've incorporated your keen recommendations to improve readability.

Would appreciate a second look at your convenience. Thank you!

pygeoapi/api/stac.py
```diff
@@ -379,9 +379,17 @@ def search(api: API, request: Union[APIRequest, Any]) -> Tuple[dict, int, str]:
                     geom = from_geojson(json.dumps(feature['geometry']))
                     feature['bbox'] = geom.bounds
 
-                for la in ['links', 'assets']:
-                    if feature.get(la) is None:
-                        feature[la] = []
+                if feature.get('links') is None:
+                    feature['links'] = feature\
```
tomkralidis: Thanks.  I am seeing:

```python
                        .get('properties', {})\
                        .get('content', {})\
                        .get('links', [])
```

A STAC Item indeed has `links` and `assets`, but they are at the root of the STAC item, not in an object called `properties.content`.

rajadain: Indeed. This is quirk of the pgSTAC back-end. Please see #2355 for a more detailed explanation. I can also add a comment here pointing to that to keep the context near the code.

Also open to alternative approaches to address this discrepancy.

tomkralidis: In this case, I would suggest removing this given it is backend specific. You can work around this by adding your own feature provider plugin which reads PgSTAC and returns items such that the STAC code in pygeoapi is reading content on the absence of a backend specific data model.

rajadain: That makes perfect sense. I'm going to remove this commit from this PR and explore creating a pgSTAC back-end. Thank you!

rajadain: That commit is gone, this is ready for another look at your convenience 🙏

tomkralidis (approved):

rajadain:
Thanks for the thoughtful review!

tomkralidis:
Thanks for the valuable contributions and fixes!

PR #2345, opened 2026-05-19
Pooling

### LINKED ISSUE #2344: PostgreSQL/SQL provider connection pool is not configurable and never recycles
## Description

`get_engine()` in `pygeoapi/provider/sql.py` creates the SQLAlchemy engine with:

```python
engine = create_engine(conn_str, connect_args=connect_args, pool_pre_ping=True)
```

No pool parameters are passed, so SQLAlchemy uses the default `QueuePool` (`pool_size=5`, `max_overflow=10`) with `pool_recycle=-1`. The `pool_size` connections are therefore held open for the entire life of the worker process
and are never recycled.

In a multi-process deployment this is significant. With N replicas and W gunicorn workers, the connection floor is `N * W * pool_size` connections that sit IDLE on the database indefinitely, plus burst overflow. On a deployment with 4 replicas x 4 workers we observed connections idle on the PostgreSQL server for **several days**, eventually exhausting `max_connections` (shared
with other services) and producing:

```
org.postgresql.util.PSQLException: FATAL: remaining connection slots are
reserved for roles with the SUPERUSER attribute
```

There is currently no way to bound or recycle the pool from configuration.

## Proposal

Allow the connection pool to be configured per provider via the existing `options:` block, exposing `pool_size`, `max_overflow`, `pool_recycle`, `pool_timeout` and `pool_pre_ping`. These would be parsed in `store_db_parameters()`, kept out of the DBAPI `connect_args`, and passed to `create_engine()`. The values are stored as a hashable tuple so `get_engine()` remains `@functools.cache`-able (engine sharing per process is preserved).

The same change must be applied to the PostgreSQL process manager (`pygeoapi/process/manager/postgresql.py`), which also calls `get_engine()`.

## Open question for the PSC / maintainers

Should the default `pool_recycle` remain `-1` (no behaviour change; pure opt-in feature) or change to a finite value such as `3600` (recycle hourly; removes the "idle forever" footgun for everyone, but is a behaviour change for all existing users)?

I propose merging with **`pool_recycle=-1` as the default** (zero behaviour change) and treating a finite default as a separate follow-up decision, but I'm happy to go whichever way the PSC prefers.

## Environment

- pygeoapi: master
- Python: 3.12
- Driver: `postgresql+psycopg2`, SQLAlchemy QueuePool

I'd like to contribute a fix (code + test + docs) for this.

KoalaGeo:
# Overview

Makes the SQLAlchemy connection pool of the SQL provider configurable per provider via the existing `options:` block, exposing `pool_size`, `max_overflow`, `pool_recycle`, `pool_timeout` and `pool_pre_ping`.

Previously `get_engine()` called `create_engine(conn_str, connect_args=connect_args, pool_pre_ping=True)` with no pool sizing or recycle, so the default `QueuePool` held `pool_size` connections open for the life of each worker process and never recycled them. In multi-process deployments this produces a large number of permanently-IDLE server-side connections (we saw connections idle for days, eventually exhausting `max_connections`). There was no way to bound or recycle the pool from configuration.

Changes:

- `store_db_parameters()` now extracts the five pool keys from `options`,   coerces them to their declared types, and stores them as a sorted, hashable `tuple` (`self.db_pool_options`). They are popped out of `options` so they  are **not** forwarded to the DBAPI as `connect_args`. 
- `get_engine()` takes a `pool_options` tuple parameter and applies  `**dict(pool_options)` to `create_engine()`. It stays `@functools.cache`-able because the parameter is a hashable tuple, so engine sharing per process is preserved; providers with differing pool config correctly get distinct engines.
- `pygeoapi/process/manager/postgresql.py` also calls `get_engine()`; its call site is updated to pass `self.db_pool_options` so the manager does not lose `pool_pre_ping` or skip recycling.

**Backward compatibility:** defaults preserve current behaviour exactly — `pool_size=5`, `max_overflow=10`, `pool_pre_ping=True`, and `pool_recycle=-1` (SQLAlchemy's default, i.e. the current effective behaviour).

This PR is therefore a pure, opt-in feature add with no behaviour change for existing users. (See the issue for discussion of whether a finite default `pool_recycle` should be adopted as a separate follow-up.)

New tests and documentation are included.

# Related Issue / discussion

Closes https://github.com/geopython/pygeoapi/issues/2344.

# Additional information

Example configuration:

```yaml
providers:
  - type: feature
    name: PostgreSQL
    data:
      host: 127.0.0.1
      port: 5432
      dbname: test
      user: postgres
      password: postgres
      search_path: [osm, public]
    options:
      pool_size: 2          # persistent connections per worker process
      max_overflow: 3       # short-lived burst capacity
      pool_recycle: 300     # recycle connections older than 5 minutes
      pool_timeout: 30
    id_field: osm_id
    table: hotosm_bdi_waterways
    geom_field: foo_geom
```

Note (documented): because `get_engine()` is `@functools.cache`-d on its full argument set, providers that share a database **must** use identical pool options to continue sharing a single engine per worker; differing options intentionally yield separate engines.

# Dependency policy (RFC2)

No new dependencies are introduced; only the standard library and the already-required SQLAlchemy are used.

# Updates to public demo

- [x] I have ensured that breaking changes to the pygeoapi master demo server have been addressed
 - [x] No changes required: defaults preserve existing behaviour, so the demo `local.config.yml` does not need to change.

# Contributions and licensing

- [x] I'd like to contribute a bugfix/feature (configurable SQL connection pool) to pygeoapi. I confirm that my contributions to pygeoapi will be compatible with the pygeoapi license guidelines at the time of contribution


webb-ben:
Is there a reason to not pop the attributes from the connect_args inside of get_engine? This would consolidate a bit of the complications noted in the PR between hashing and the manager using get_engine. Maybe I am missing something

pygeoapi/provider/sql.py
```diff
@@ -615,6 +616,24 @@ def store_db_parameters(
         connection_data.get('search_path') or
         options.pop('search_path', ['public'])
     )
+    # Connection-pool tuning. These are popped out of ``options`` so they
+    # are NOT passed to the DBAPI as connect_args, and are coerced to a
+    # hashable form so get_engine() can stay functools.cache()-able.
+    # Defaults keep SQLAlchemy's QueuePool sizing but, unlike SQLAlchemy's
+    # default of -1, recycle connections after an hour so that pooled
+    # connections cannot sit IDLE on the server indefinitely.
```
ricardogsilva: This part of the comment seems to be outdated, as you end up setting the default value of `pool_recycle` to `-1`

docs/source/publishing/ogcapi-features.rst
```diff
@@ -703,11 +703,27 @@ These are optional and if not specified, the default from the engine will be use
...
+             # Extra short-lived connections allowed above pool_size.
+             max_overflow: 10
+             # Recreate connections older than this many seconds. -1 (the
+             # default) never recycles; set a finite value (e.g. 300) so
+             # pooled connections cannot sit IDLE on the server indefinitely.
+             pool_recycle: -1
+             # Seconds to wait for a connection from the pool before erroring.
+             pool_timeout: 30
+             # Test connections with a lightweight ping before use.
+             pool_pre_ping: true
```
ricardogsilva, line 719: All of these new parameters need to be added to the config schema at

`pygeoapi/resources/schemas/config/pygeoapi-config-0.x.yml`

This will make it possible to test a pygeoapi configuration for correctness even before starting up the server.

pygeoapi/provider/sql.py
```diff
@@ -615,6 +616,24 @@ def store_db_parameters(
...
+        'max_overflow': 10,
+        'pool_recycle': -1,   # SQLAlchemy default; preserves current behaviour
+        'pool_timeout': 30,
+        'pool_pre_ping': True,
+    }
+    self.db_pool_options = tuple(sorted(
+        (key, type(default)(options.pop(key, default)))
+        for key, default in pool_defaults.items()
+    ))
+
```
ricardogsilva: In my opinion this could be made easier to read and less complex by:

- Storing `self.db_pool_options` as a `dict` instead of a `tuple`, and defer tuple creation to when `get_engine` is called;
- Relying on the types of passed `options` already being correct. Adding these new parameters to the config JSON Schema (as I mentioned in my other comment) would mean that the type of each parameter would already be documented and would be enforceable by doing a validation of the config.

Also, note that your implementation contains a subtle bug when trying to parse `pool_pre_ping`. If the original value had been: 

```python
{'pool_pre_ping': 'False'}  # I'm passing a string with the value of "False"
```

then the outcome would be:

```python
# type(True)("False")
True
```

In other words, `bool("False")` is actually `True` because non-empty strings are truthy.

```diff
-pool_defaults = {
-    'pool_size': 5,
-    'max_overflow': 10,
-    'pool_recycle': -1,   # SQLAlchemy default; preserves current behaviour
-    'pool_timeout': 30,
-    'pool_pre_ping': True,
-}
-self.db_pool_options = tuple(sorted(
-    (key, type(default)(options.pop(key, default)))
-    for key, default in pool_defaults.items()
-))
+self.pool_defaults = {
+    'pool_size': options.pop('pool_size', 5),
+    'max_overflow': options.pop('max_overflow', 10),
+    'pool_recycle': options.pop('pool_recycle', -1),  # SQLALchemy default - never release connections
+    'pool_timeout': options.pop('pool_timeout', 30),
+    'pool_pre_ping': options.pop('pool_pre_ping', True),
+}
```

pygeoapi/provider/sql.py
```diff
@@ -134,6 +134,7 @@ def __init__(
             self.db_user,
             self._db_password,
             self.db_conn,
+            self.db_pool_options,
```
ricardogsilva: as per my other comment, in my opinion it would be clearer if the tuple would be generated here, perhaps also accompanied with a comment mentioning that this is made as a way to enable making use of `functools.cache`.

Also, in [modern Python](https://docs.python.org/3/whatsnew/3.7.html#summary-release-highlights), a dict's insertion ordering is preserved, so I don't think sorting the tuple would be needed. 


```suggestion
            tuple(self.db_pool_options.items()),  # convert to hashable type, for using with functools.cache
```

tests/provider/test_sql_pool_options.py
```diff
@@ -0,0 +1,99 @@
...
+    obj = _Dummy()
+    store_db_parameters(obj, dict(CONN), {})
+    pool = dict(obj.db_pool_options)
+    # Defaults must match pre-existing effective behaviour:
+    # pool_pre_ping was hardcoded True; pool_recycle was unset (-1).
+    assert pool['pool_size'] == 5
+    assert pool['max_overflow'] == 10
+    assert pool['pool_timeout'] == 30
+    assert pool['pool_pre_ping'] is True
+    assert pool['pool_recycle'] == -1
```
ricardogsilva: This test seems unnecessary to me - when this PR gets merged, the behavior it implements will become the current behavior, so the test looses its relevancy.

tests/provider/test_sql_pool_options.py
```diff
@@ -0,0 +1,99 @@
...
+        obj, dict(CONN),
+        {'pool_size': 2, 'max_overflow': 3, 'pool_recycle': 300},
+    )
+    pool = dict(obj.db_pool_options)
+    assert pool['pool_size'] == 2 and isinstance(pool['pool_size'], int)
+    assert pool['max_overflow'] == 3
+    assert pool['pool_recycle'] == 300
+    # untouched keys keep defaults
+    assert pool['pool_timeout'] == 30
+    assert pool['pool_pre_ping'] is True
```
ricardogsilva: This test would be unnecessary if you'd go with my suggestion above, of storing `db_pool_options` as a `dict` instead of a `tuple` and you'd rely on the configuration being valid after having added the JSON schema bits that are missing.

tests/provider/test_sql_pool_options.py
```diff
@@ -0,0 +1,99 @@
...
+
+def test_dict_valued_options_still_filtered():
+    obj = _Dummy()
+    store_db_parameters(
+        obj, dict(CONN),
+        {'pool_size': 2, 'zoom': {'min': 0, 'max': 22}},
+    )
+    assert 'zoom' not in obj.db_options
+    assert dict(obj.db_pool_options)['pool_size'] == 2
+
```
ricardogsilva: This test seems to be unnecessary, as it is not testing the changes you made in this PR. It verifies that the contents of `obj.db_options` are correct. 

IMO this PR does not make any changes that would warrant this verification, unless you would not trust the behavior of `dict.pop`, which is a Python builtin.

tests/provider/test_sql_pool_options.py
```diff
@@ -0,0 +1,99 @@
...
+    store_db_parameters(b, dict(CONN), {'pool_size': 2})
+    # identical config -> identical key -> shared engine via functools.cache
+    assert a.db_pool_options == b.db_pool_options
+    assert hash(a.db_pool_options) == hash(b.db_pool_options)
+
+    c = _Dummy()
+    store_db_parameters(c, dict(CONN), {'pool_size': 9})
+    # differing pool config -> distinct key (separate engine, by design)
+    assert c.db_pool_options != a.db_pool_options
+
```
ricardogsilva: This is testing Python's own implementation of how tuples are hashed, so I don't think it is relevant to include in pygeoapi.

tests/provider/test_sql_pool_options.py
```diff
@@ -0,0 +1,99 @@
...
+def test_pool_options_coexist_with_search_path():
+    obj = _Dummy()
+    store_db_parameters(
+        obj, dict(CONN),
+        {'search_path': ['published', 'public'], 'pool_size': 4},
+    )
+    assert obj.db_search_path == ('published', 'public')
+    assert dict(obj.db_pool_options)['pool_size'] == 4
+
```
ricardogsilva: This test seems unnecessary, as it does not test the functionality introduced in this PR

ricardogsilva (commented):
Just leaving my two cents here - I'm not a core committer so take these with a grain of salt.

Overall I agree with the PR, as adding these connection-related options seems relevant - thanks for your work and I look forward to having it merged!

Personally, I would simplify the implementation a bit, by relying on pygeoapi's JSON Schema document for the validation of the config.

And I would not include most of these tests, which I see as not being relevant.

pygeoapi/provider/sql.py
ricardogsilva: I just found out that the upcoming Python 3.15 will come with a new [frozendict](https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-frozendict) type. This means this new dict type will be hashable, thus being usable as part of a cacheable function.

It will likely take a long while before pygeoapi can rely on this though, but nice to see this becoming a standard feature in Python.

KoalaGeo:
> Is there a reason to not pop the attributes from the connect_args inside of get_engine? This would consolidate a bit of the complications noted in the PR between hashing and the manager using get_engine. Maybe I am missing something

That's a good shout, I'll refactor

tomkralidis:
@KoalaGeo any update on this PR?

docs/source/publishing/ogcapi-features.rst
KoalaGeo: Done :-)

tests/provider/test_sql_pool_options.py
KoalaGeo: removed

tests/provider/test_sql_pool_options.py
KoalaGeo: Agreed

tests/provider/test_sql_pool_options.py
KoalaGeo: Agreed

tests/provider/test_sql_pool_options.py
KoalaGeo: Agreed

KoalaGeo:
@tomkralidis think that's addressed everything raised by @webb-ben and @ricardogsilva — let me know if I've missed anything.

Quick summary of what changed:

- `get_engine()` now pops the pool keys out of `connect_args` itself (@webb-ben's suggestion). That removes the separate `db_pool_options` tuple entirely, and the manager call site no longer needs any special handling — it just passes `**self.db_options` like the provider does. It stays `functools.cache`-able since the kwargs are hashable scalars.
- Added `pool_size` / `max_overflow` / `pool_recycle` / `pool_timeout` / `pool_pre_ping` to the config JSON Schema (@ricardogsilva), so a config can be validated before the server starts. That's now what enforces the types, so I've dropped the manual `type(default)(...)` coercion — which also removes the `bool("False") is True` gotcha you spotted.
- Removed the outdated `pool_recycle` comment, and trimmed the tests down to a single behavioural test covering the pool/`connect_args` split in `get_engine()`.

Defaults are unchanged, so existing deployments behave exactly as before.

tests/provider/test_sql_pool_options.py
tomkralidis, line 1: Can you add a source code header ([example](https://github.com/geopython/pygeoapi/pull/2345/changes#diff-6902cec8691e6d183dfae638fdab863a09ee6d6ef2a1d1e5c4ab203fbf21f2e8)) to this (new) file, of course with only your authorship and copyright accordingly?

tomkralidis (changes requested):

tomkralidis:
> @tomkralidis think that's addressed everything raised by @webb-ben and @ricardogsilva — let me know if I've missed anything.
> 
> Quick summary of what changed:
> 
> * `get_engine()` now pops the pool keys out of `connect_args` itself (@webb-ben's suggestion). That removes the separate `db_pool_options` tuple entirely, and the manager call site no longer needs any special handling — it just passes `**self.db_options` like the provider does. It stays `functools.cache`-able since the kwargs are hashable scalars.
> * Added `pool_size` / `max_overflow` / `pool_recycle` / `pool_timeout` / `pool_pre_ping` to the config JSON Schema (@ricardogsilva), so a config can be validated before the server starts. That's now what enforces the types, so I've dropped the manual `type(default)(...)` coercion — which also removes the `bool("False") is True` gotcha you spotted.
> * Removed the outdated `pool_recycle` comment, and trimmed the tests down to a single behavioural test covering the pool/`connect_args` split in `get_engine()`.
> 
> Defaults are unchanged, so existing deployments behave exactly as before.

Thanks @KoalaGeo!  See last minor change request on source code header for `tests/provider/test_sql_pool_options.py`

KoalaGeo:
Done :-)

tomkralidis (approved):

PR #2346, opened 2026-05-21
add translation support for data elements

tomkralidis:
# Overview
This PR adds support for elements that are returned as part of JSON (not HTML) responses, as well as updates various French translations.
# Related Issue / discussion
None

# Additional information
None
# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/l10n.py
+++ b/pygeoapi/l10n.py
@@ -249,11 +249,11 @@ def translate(value: str, language: Union[Locale, str]):
         raise LocaleError('language is not a str or Locale')
 
     # First try fast approach: directly fetch expected language key
-    translation = value.get(locale2str(language)
-                            if hasattr(language, 'language') else language)
+    translation_ = value.get(locale2str(language)
+                             if hasattr(language, 'language') else language)
 
-    if translation:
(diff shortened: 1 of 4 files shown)


pygeoapi/l10n.py
```diff
@@ -237,6 +251,7 @@ def translate(value: str, language: Union[Locale, str]):
     # First try fast approach: directly fetch expected language key
     translation = value.get(locale2str(language)
```
GeoSander: @tomkralidis You are essentially overwriting the globally imported `translation` method from `gettext` here in the local scope. This is not a big deal in this case since you're doing the `if translation` afterwards, but to avoid potential issues in the future, perhaps it's better to either rename this variable here or rename the import?

GeoSander (commented):
Looks good to me, Tom. Thanks for adding this!

pygeoapi/l10n.py
tomkralidis: Thanks @GeoSander, fixed.

GeoSander (approved):
👍

PR #2352, opened 2026-06-03
HTML: add safe bbox function for dateline crossing

tomkralidis:
# Overview
This PR adds support to HTML to render (collection) polygons supporting cases with dateline crossing.  Note that data is not changed per se, as the bbox adjustment is solely for visualization purposes.

# Related Issue / discussion
Downstream in https://github.com/World-Meteorological-Organization/wis2box/issues/1103

# Additional information
Current master:
image
This PR:

image

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/static/js/default.js
+++ b/pygeoapi/static/js/default.js
@@ -1,0 +1,46 @@
+/******************************************************************
+*
+* Authors: Tom Kralidis <tomkralidis@gmail.com>
+*
+* Copyright (c) 2026 Tom Kralidis
+*
+* Permission is hereby granted, free of charge, to any person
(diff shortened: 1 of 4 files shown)


webb-ben:
Would this not affect a bbox of the whole world? Does the bbox function always get passed WGS84?

tomkralidis:
> Would this not affect a bbox of the whole world? Does the bbox function always get passed WGS84?

Updated.  The bbox function is working off 4326 as a rule (just as existing functionality to plot is bound to 4326).

pvgenuchten (approved):
Much cleaner Tom, thanx

Issue #2364, opened 2026-06-19
pygeoapi fails to throw a 404 code, when it does not find a tile (mvt-proxy)


doublebyte1:
**Description**

The [martin](https://github.com/maplibre/martin)  tile server returns a 500 error (rather than 404), when it does not find a tile. On the mvt-proxy provider, pygeoapi carries that error code, failing to meet  Abstract test A.7, Requirement 6: /req/core/tc-error: "	verify Error Responses"

Tile that exists:

```
curl -I localhost:3000/srup_arvores_point/0/0/0
HTTP/1.1 204 No Content
vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
date: Fri, 19 Jun 2026 12:49:01 GMT

```

Tiles that does not exist:
```
curl -I localhost:3000/srup_arvores_point/0/1/100
HTTP/1.1 500 Internal Server Error
content-length: 60
vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
content-type: text/plain; charset=utf-8
date: Fri, 19 Jun 2026 12:46:55 GMT

```

**Steps to Reproduce**

Request a tile out of bound in pygeoapi, using the MVT-proxy provider and martin as a backend:

```
curl -I localhost:5000/collections/srup_arvores_point/tiles/WebMercatorQuad/0/0/100?f=pbf
HTTP/1.1 500 INTERNAL SERVER ERROR
Server: Werkzeug/3.1.8 Python/3.12.0
Date: Fri, 19 Jun 2026 12:52:02 GMT
Content-Type: application/vnd.mapbox-vector-tile
X-Powered-By: pygeoapi 0.24.dev0
Content-Language: en-US
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: *
Content-Length: 112
Connection: close

```

**Expected behavior**
Receive a 404 error

**Environment**
 - OS: Ubuntu 24.04.4 LTS
 - Python version: Python 3.12.3
 - pygeoapi version:  0.24.dev0 (latest master)

PR #2365, opened 2026-06-19
Force a 404 code for tiles out of bounds

doublebyte1:
# Overview

# Related Issue / discussion

https://github.com/geopython/pygeoapi/issues/2364

# Additional information

On the mvt-proxy provider, we are not sure what the behaviour of the upstream server is regarding tiles out of bounds; however, we need to make sure that if the tile is out of bounds we return a 404. This PR forces that behaviour on mvt-proxy, by evaluating if the 500 error was caused by requesting a tile outside the limits of the TMS. This also ensures that we pass the conformance test.

```
curl -I localhost:5000/collections/srup_arvores_point/tiles/WebMercatorQuad/0/0/100?f=pbf
HTTP/1.1 404 NOT FOUND
Server: Werkzeug/3.1.8 Python/3.12.0
Date: Fri, 19 Jun 2026 12:58:03 GMT
Content-Type: application/vnd.mapbox-vector-tile
X-Powered-By: pygeoapi 0.24.dev0
Content-Language: en-US
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: *
Content-Length: 14
Connection: close

```

Screenshot from 2026-06-19 13-56-56

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/provider/mvt_proxy.py
+++ b/pygeoapi/provider/mvt_proxy.py
@@ -176,7 +176,7 @@ class MVTProxyProvider(BaseMVTProvider):
                     else:
                         resp = session.get(f'{base_url}/{layer}/{z}/{x}/{y}{url_query}')  # noqa
 
-                    if resp.status_code == 404:
+                    if resp.status_code == 404 or resp.status_code == 500:
                         if (self.is_in_limits(self.get_tilematrixset(tileset), z, x, y)): # noqa
                             return None
                         raise ProviderTileNotFoundError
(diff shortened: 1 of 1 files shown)


pygeoapi/provider/mvt_proxy.py
```diff
@@ -176,7 +176,7 @@ def get_tiles(self, layer=None, tileset=None,
                     else:
                         resp = session.get(f'{base_url}/{layer}/{z}/{x}/{y}{url_query}')  # noqa
 
-                    if resp.status_code == 404:
+                    if resp.status_code == 404 or resp.status_code == 500:
```
tomkralidis: Can we update to:

```python
if resp.status_code in [404, 500]:
```

tomkralidis (changes requested):

tomkralidis (approved):

Issue #2370, opened 2026-06-24
improve flow through ogcapi-cov


pvgenuchten:
**Is your feature request related to a problem? Please describe.**

- as an average user i find the user flow through coverage-api troublesome
- https://demo.pygeoapi.io/master/collections/gdps-temperature 
   - Clear ui difference between vector datasets and cov datasets, so i understand their different behaviour 
   - Linkage to https://demo.pygeoapi.io/master/collections/gdps-temperature/queryables is not indicated
   - unclear list of links (which one to choose)
  <img width="483" height="300" alt="Image" src="https://github.com/user-attachments/assets/f5c921d8-ca84-4942-b5f6-b33f70baac9d" />

- https://demo.pygeoapi.io/master/collections/gdps-temperature/coverage -> 500 Invalid format requested

**Describe the solution you'd like**

- more user friendly guidance
- prevent format exception at coverage root (using browser)

**Describe alternatives you've considered**

either minimal metadata in html at most endpoints, or provide a coverage client (as map/diagram)

PR #2373, opened 2026-06-24
Minor coverages improvements

webb-ben:
# Overview
Based on https://github.com/geopython/pygeoapi/issues/2370, introduce minor coverage improvements:
- Add queryables to coverage
- Add HTML representation and advertise at `/collections/{cid}`. This is a shallow copy of the EDR HTML template with small modifications. The HTML is served from browsers with default accept headers
image
image
- Advertise Coverage from /collection HTML
image

# Related Issue / discussion
https://github.com/geopython/pygeoapi/issues/2370

# Additional information

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing


pvgenuchten (approved):
thanx @webb-ben, changes are usefull

let's discuss in timisuara how we can best improve the link display for a next PR

@tomkralidis can we release before foss4g-eu? a change like this would be usefull for the training

tomkralidis:
We will be releasing 0.24 after FOSS4G-EU.

doublebyte1 (approved):
+1

pygeoapi/api/collection.py
```diff
@@ -218,6 +218,7 @@ def gen_collection(api, request, dataset: str,
         }])
 
     if collection_data_type in ['feature', 'coverage', 'record']:
+        data['itemType'] = collection_data_type
```
tomkralidis: A coverage is not an itemType.  Remove.

tomkralidis (changes requested):

pygeoapi/api/collection.py
webb-ben: Can change. As far as I understood, the itemType is never emitted to the client. This is only used inside the Jinja templates to add additional links. When I first tried the approach used for tiles, I wasn't getting the link to pop up. Will have to look back later.

webb-ben: image

pygeoapi/templates/collections/collection.html
```diff
@@ -74,20 +74,40 @@ <h3>{% trans %}Schema{% endtrans %}</h3>
...
+            </li>
+          </ul> 
+        {% endif  %}
+
+        {% if 'coverage' in provider['type'] %}
+        <h3>{% trans %}Coverage{% endtrans %}</h3>
+        <ul>
+          <li>
+            <div>
+              <a title="{% trans %}Browse Coverage{% endtrans %}" href="{{ data['collections_path'] }}/{{ data['id'] }}/coverage">
```
tomkralidis: Add to `locale/en/LC_MESSAGES/messages.po`:

```
msgid "Browse Coverage"
msgstr ""
```

pygeoapi/templates/collections/collection.html
```diff
@@ -74,20 +74,40 @@ <h3>{% trans %}Schema{% endtrans %}</h3>
...
+                {% trans %}Display Coverage of{% endtrans %} "{{ data['title'] }}"</a></div>
```
tomkralidis, line 97: Add to `locale/en/LC_MESSAGES/messages.po`:

```
msgid "Display Coverage of"
msgstr ""
```

tomkralidis (changes requested):

pygeoapi/api/collection.py
pvgenuchten: if not a coverage, what itemtype is it?
or do we expect a dedicated cov folder in template directory, like edr,tiles,map?
image

pygeoapi/templates/collections/coverage/index.html
tomkralidis, line 1: Rename to `pygeoapi/templates/collections/coverage/index.html`

tomkralidis (changes requested):

pygeoapi/api/collection.py
tomkralidis: It is not an itemType at all.  An itemType is something that is returned from an `.../items/{itemId}`.

The resulting directory / filepath for the template should be `pygeoapi/templates/collections/coverage/index.html` (change now requested).

pygeoapi/api/coverages.py
```diff
@@ -226,7 +226,7 @@ def get_collection_coverage(
         }]
 
         content = render_j2_template(api.tpl_config, tpl_config,
-                                     'collections/coverages/coverage.html',
+                                     'collections/coverages/index.html',
```
tomkralidis: `collections/coverage/index.html` (and rename actual template file).

tomkralidis (changes requested):

pvgenuchten:
Thanx @webb-ben and @tomkralidis, useful discussion and great work!

pvgenuchten:
the .../coverage page is nice with the coverage view, but it looks a bit empty, maybe good to repeat the title/abstract/keywords of the coverage here. The difference with the /collections/grid request page is minimal then, but should be ok.

the json and json-ld links in the top right, can we extend it to include the cov-json and grib formats? (json-ld should probably be removed, because clicking it throws invalidparamvalue, it is also used to embed schema.org on the page, this should probably also be removed, not sure if easy to remove, because this comes from inherited templates)

webb-ben:
The coverage HTML is a near direct copy of the EDR HTML in which the JSON-LD does not work either. I'd be interested to also return the links for these in the CovJSON (related to https://github.com/opengeospatial/ogcapi-environmental-data-retrieval/issues/666)

Out of scope for this PR for sure and not sure how actually helpful it is, but there is https://covjson.org/context.jsonld

tomkralidis (approved):

PR #2379, opened 2026-06-25
fix: normalize trailing slash in root URL (#2361)

### LINKED ISSUE #2361: The url configuration paramater does not respect the trailing slash as the application root
**Description**

the url configuration paramater does not respect the trailing slash as the application root

`url: http://localhost:5000/somepath/api/` so base urls are generated without the trailing slash

**Steps to Reproduce**

add `url: http://localhost:5000/somepath/api/` the pygeoapi yml configuration

**Expected behavior**

The expectation is that the url respects the trailing slash for a base url and links are generated appropriately

so setting `http://localhost:5000/somepath/api/` should generate root level urls as `http://localhost:5000/somepath/api/?f=jsonld` or  `http://localhost:5000/somepath/api/` NOT `http://localhost:5000/somepath/api?f=jsonld?` or `http://localhost:5000/somepath/api`

**Screenshots/Tracebacks**

**Environment**
 - OS: RHEL 9.x
 - Python version: 3.12
 - pygeoapi version: 0.23.0

**Additional context**

C1-BA-B1-F3:
Fixes #2361

Root URL links in landing page, collections, and item endpoints were constructed as `base_url?f=json` without a slash before the query string. When users configure a URL with a trailing slash (e.g., `http://localhost:5000/somepath/api/`), the generated URLs would incorrectly omit the slash (e.g., `http://localhost:5000/somepath/api?f=json`).

Changed root URL link construction to always include a `/` before the query string, producing correct URLs like `http://localhost:5000/somepath/api/?f=json`.

### DIFF
--- a/pygeoapi/api/__init__.py
+++ b/pygeoapi/api/__init__.py
@@ -662,17 +662,17 @@ def landing_page(api: API,
         'rel': request.get_linkrel(F_JSON),
         'type': FORMAT_TYPES[F_JSON],
         'title': l10n.translate('This document as JSON', request.locale),
-        'href': f"{api.base_url}?f={F_JSON}"
+        'href': f"{api.base_url}/?f={F_JSON}"
     }, {
         'rel': request.get_linkrel(F_JSONLD),
         'type': FORMAT_TYPES[F_JSONLD],
         'title': l10n.translate('This document as RDF (JSON-LD)', request.locale),  # noqa
-        'href': f"{api.base_url}?f={F_JSONLD}"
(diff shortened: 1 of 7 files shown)


pvgenuchten:
This is one of the known incompatibilities between stac and ogcapi, stac is a browsable folder structure, so expects slashes, ogcapi is an api without slashes

tomkralidis:
Agree @pvgenuchten this is a longstanding discussion upstream of pygeoapi.  Closing.

Issue #2376, opened 2026-06-25
move jsonfg from plugin to default


pvgenuchten:
**Is your feature request related to a problem? Please describe.**

I noticed a [plugin jsonfg](https://github.com/cgs-earth/pygeoapi-plugins/blob/master/pygeoapi_plugins/formatter/jsonfg.py), by now is it an idea to move it to default status, since in may it was adopted by OGC as formal standard

**Additional context**

which is the formal way to request jsonfg from a server?

accept: draft: application/vnd.ogc.fg+json / final: application/fg+json (although not registered at IANA yet)

do we support the shortcut ?f=fg or f=jsonfg

webb-ben:
The pygeoapi formatter implementation uses Formatter.f as the f= query arg match and Formatter.mimetype for Accept header.

As implemented in downstream plugin, https://reference.geoconnex.us/collections/hu02/items?f=jsonfg resolves to JSON-FG

pvgenuchten:
which accept header is that? application/fg+json or application/vnd.ogc.fg+json

or is it this one? doesn't that conflict with default geojson?

https://github.com/cgs-earth/pygeoapi-plugins/blob/c8fd875b4f7b2775968cb22e3dc38ff5d2e7a666/pygeoapi_plugins/formatter/jsonfg.py#L70

webb-ben:
Yep. Will fix in a sec

pvgenuchten:
hi, sorry, just noticed this discussion https://github.com/opengeospatial/ogc-feat-geo-json/issues/129

seems the group decided to not register a dedicated accept header, but use an additional profile parameter (wonder what the geojson community thinks about that)

- to request using accept header: Accept: application/geo+json; profile="http://www.opengis.net/def/profile/OGC/0/jsonfg"
- for the f=xx shortcut, they suggest ?f=json&profile=jsonfg

```
curl \
  -H 'Accept: application/geo+json; profile="http://www.opengis.net/def/profile/OGC/0/jsonfg"' \
  https://example.org/collections/buildings/items
```

wow, this stuff is complex

webb-ben:
pygeoapi does not have that functionality at the moment. As valuable as it would be to have JSON-FG I do have questions about creating a new content negotiation profile

Will revert to application/geo+json seeing as pygeoapi looks for application/json

pvgenuchten:
I will discuss with tom and those present in timisuara next week

doublebyte1:
[...]

PR #2384, opened 2026-06-28
show alternate formats which are listed in links

pvgenuchten:
# Overview

- move the links section to a generic template, reducing the code and be more uniform
- loop through list of links, to see which alternates are available, and show those on the toolbar 
- only include json-ld link if a representation in json-ld is available (by checking links)
- it updates the json-ld include js, to modern js

# Related Issue / discussion

this came up in #2376; in coverages, there is no json-ld, still the json-ld link was shown, clicking it causes an error
on the other hand there are no links to other formats, such as jsonfg, csv, shapefile, grib2 etc
this PR solves the issue, it still needs testing against various providers/services, as it may give unexpected results

# Additional information

in pygeoapi links are either advertised in response.links or response.data.links, so check both for the links element, some format such as cov-json do not have a links element, in that case the html data seems also not having the links element.

image

a mapping would be needed to translate mimetype (from link@type) to link-text, maybe the ?f=xxx format is expected, or the trans module can be used?

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

### DIFF
--- a/pygeoapi/templates/_base.html
+++ b/pygeoapi/templates/_base.html
@@ -37,7 +37,7 @@
     <meta property="og:description"  content="{{ config['metadata']['identification']['description'] }}" />
 
     <meta property="og:type" content="website" />
-    {% for link in data['links'] %}
+    {% for link in (links or data['links']) %}
       {% if (link['rel']=="self" and link['type']=="text/html") %}
       <link rel="canonical" href="{{ link['href'].split('?')[0] }}?f=html" />
       {% elif (link['rel']!="canonical") %}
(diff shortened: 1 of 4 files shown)


webb-ben:
While there are genuine limits to serving an incomplete view of the collection, might this overflow for a collection like
```
curl -s "https://reference.geoconnex.us/collections/hu02?f=json" \
  | jq -r '.links[] | select(.rel == "items") | .href'
https://reference.geoconnex.us/collections/hu02/items?f=json
https://reference.geoconnex.us/collections/hu02/items?f=jsonld
https://reference.geoconnex.us/collections/hu02/items?f=html
https://reference.geoconnex.us/collections/hu02/items?f=csv
https://reference.geoconnex.us/collections/hu02/items?f=csv
https://reference.geoconnex.us/collections/hu02/items?f=jsonfg
https://reference.geoconnex.us/collections/hu02/items?f=shp
https://reference.geoconnex.us/collections/hu02/items?f=gpkg
https://reference.geoconnex.us/collections/hu02/items?f=kml
https://reference.geoconnex.us/collections/hu02/items?f=parquet
```

pvgenuchten:
Yes, great point @webb-ben, i wondered if we can cluster similar requests and showing their formats as a list? Eg:

item detail (json, jsonld, html, csv, …)

PR #2396, opened 2026-08-04
RFC7: add support for validator plugins

tomkralidis:
# Overview
This PR implements [RFC7](https://pygeoapi.io/development/rfc/7/), adding support for validation plugin facilities in support of data validation on item transactions.

# Related Issue / discussion

# Additional information

RFC7: https://pygeoapi.io/development/rfc/7

# Dependency policy (RFC2)

# Updates to public demo

# Contributions and licensing

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pygeoapi. I confirm that my contributions to pygeoapi wilxl be compatible with the pygeoapi license guidelines at the time of contribution


justb4 (commented):
Very useful addition. What I don't understand: in the configuration example a Python Module path is specified like `name: path.to.validator`. But in `plugin.py` a `name: Python path` is defined like:
```
'validator': {
        'GeoJSON': 'pygeoapi.validator.geojson.GeoJSONValidator'
```
Maybe I did not understand, or is the idea one can also refer by name, like `GeoJSON`, to a pre-configured `validator` in `plugin.py`?

tomkralidis:
> Very useful addition. What I don't understand: in the configuration example a Python Module path is specified like `name: path.to.validator`. But in `plugin.py` a `name: Python path` is defined like:
> 
> ```
> 'validator': {
>         'GeoJSON': 'pygeoapi.validator.geojson.GeoJSONValidator'
> ```
> 
> Maybe I did not understand, or is the idea one can also refer by name, like `GeoJSON`, to a pre-configured `validator` in `plugin.py`?

Yes, validators that are shipped "on board" with pygeoapi can be referrred to by a shortname (i.e. `GeoJSON`), which evaluates to the dotted path, in configuration.  Configuration can also explicitly point to a dotted path just the same; the result of both approaches is a dotted path/class that is instantiated accordingly.

justb4:
> > Very useful addition. What I don't understand: in the configuration example a Python Module path is specified like `name: path.to.validator`. But in `plugin.py` a `name: Python path` is defined like:
> > ```
> > 'validator': {
> >         'GeoJSON': 'pygeoapi.validator.geojson.GeoJSONValidator'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe I did not understand, or is the idea one can also refer by name, like `GeoJSON`, to a pre-configured `validator` in `plugin.py`?
> 
> Yes, validators that are shipped "on board" with pygeoapi can be referrred to by a shortname (i.e. `GeoJSON`), which evaluates to the dotted path, in configuration. Configuration can also explicitly point to a dotted path just the same; the result of both approaches is a dotted path/class that is instantiated accordingly.

Ok, suggestion to add this to the documentation, which only mentions/suggests "module path".

tomkralidis:
> > > Very useful addition. What I don't understand: in the configuration example a Python Module path is specified like `name: path.to.validator`. But in `plugin.py` a `name: Python path` is defined like:
> > > ```
> > > 'validator': {
> > >         'GeoJSON': 'pygeoapi.validator.geojson.GeoJSONValidator'
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Maybe I did not understand, or is the idea one can also refer by name, like `GeoJSON`, to a pre-configured `validator` in `plugin.py`?
> > 
> > 
> > Yes, validators that are shipped "on board" with pygeoapi can be referrred to by a shortname (i.e. `GeoJSON`), which evaluates to the dotted path, in configuration. Configuration can also explicitly point to a dotted path just the same; the result of both approaches is a dotted path/class that is instantiated accordingly.
> 
> Ok, suggestion to add this to the documentation, which only mentions/suggests "module path".

+1, docs updated.

pygeoapi/api/itemtypes.py
```diff
@@ -797,6 +797,28 @@ def manage_collection_item(
             HTTPStatus.BAD_REQUEST, headers, request.format,
             'InvalidParameterValue', msg)
 
+    if action in ['create', 'update']:
+        if p.validator is not None:
+            LOGGER.debug('Provider is configured for validation')
+            LOGGER.debug('Loading validator')
+            try:
+                v = load_plugin('validator', {'name': p.validator['name']})
```
francbartoli, line 806: JSON schema recompiles at each call, maybe we can cache the compiled schema?

pygeoapi/api/itemtypes.py
```diff
@@ -797,6 +797,28 @@ def manage_collection_item(
...
+            except Exception:
+                msg = 'Invalid validator configured'
+                return api.get_exception(
+                    HTTPStatus.INTERNAL_SERVER_ERROR, headers, request.format,
+                    'NoApplicableCode', msg)
+
+            LOGGER.debug('Validating item')
+            try:
+                v.validate(request.data)
+            except Exception as err:
```
francbartoli: Suggest to raise `ValidatorGenericError as err` and use `err.http_status_code` / `err.ogc_exception_code` because an invalid payload returns `500`

francbartoli: a non-JSON body crashes because `JSONDecodeError` object has no attribute `user_msg`. Suggest to add more tests to cover these cases

pygeoapi/api/itemtypes.py
```diff
@@ -797,6 +797,28 @@ def manage_collection_item(
...
+                msg = err.user_msg or 'Item is not valid, please check and validate payload'  # noqa
```
francbartoli, line 817: Validation details never reach the client, nor even the log line below

francbartoli (changes requested):

pygeoapi/api/itemtypes.py
tomkralidis: This is by design (Postel's Law).  Validation errors are emitted to the logger.

Example payload:
```json
    {
      "geoemetry": {
        "type": "Point",
        "coordinates": [
          -130.44472222222223,
          54.28611111111111
        ]
      },
      "type": "Feature",
      "properties": {
        "id": 1972,
        "foo": "bar",
        "title": null
      },
      "id": 48693
}
```

Error emitted to log on POST `/collections/foo`, for example:

```
[2026-08-16T22:55:27Z] {/Users/tomkralidis/Dev/pygeoapi/validator/pygeoapi/validator/geojson.py:94} ERROR - Invalid GeoJSON payload: ["[]: 'geometry' is a required property"]
```

By default, validation errors in the shipped GeoJSON validator are not emitted (so as not to echo potentially dangerous input).

If a validator plugin chooses to populate `user_msg`, then this is indeed emitted to the client.

In summary, the validator plugin has the option to be as explicit as desired.

docs/source/transactions.rst
```diff
@@ -17,3 +17,33 @@ It should be made clear that authentication and authorization is beyond the resp
...
+custom business rules as needed to ensure data is valid prior to adding or updating a given provider backend.
+
+Given the example below:
+
+.. code-block:: yaml
+
+   providers:
+       - type: feature
+         name: Elasticsearch
+         data /path/to/file
```
webb-ben: Fix YAML syntax

docs/source/configuration.rst
```diff
@@ -286,6 +286,8 @@ default.
                 storage_crs: http://www.opengis.net/def/crs/OGC/1.3/CRS84  # optional CRS in which data is stored, default: as 'crs' field
                 storage_crs_coordinate_epoch: 2017.23  # optional, if storage_crs is a dynamic coordinate reference system
                 always_xy: false  # optional should CRS respect axis ordering
+                validator:
+                    name: path.to.validator  # Python path of validation definition
```
webb-ben, line 290: Do we think it's clear enough that validator(s) require the editable field to apply? 

Will also note the config usually does a plural for the entries (resources, providers, formatters)

pygeoapi/resources/schemas/config/pygeoapi-config-0.x.yml
```diff
@@ -637,6 +637,15 @@ properties:
                                           type: boolean
                                           description: whether to include extra query parameters
                                           default: false
+                                      validator:
```
webb-ben: validator is an array of single property objects right?

docs/source/configuration.rst
tomkralidis: > Will also note the config usually does a plural for the entries (resources, providers, formatters)

Validator is meant to be singular, as stacking validators is not in scope for the design.

pygeoapi/validator/geojson.py
```diff
@@ -0,0 +1,98 @@
...
+            msg = 'Partial validation not supported'
+            raise ValidatorValidationError(msg)
+
+        LOGGER.debug(f'Validating against {SCHEMA_FILE}')
+        try:
+            data_payload = json.loads(data)
+        except json.decoder.JSONDecodeError as err:
+            msg = 'Error decoding GeoJSON'
+            LOGGER.error(f'{msg}: {err}')
+            raise ValidatorValidationError(msg)
```
francbartoli, line 82: This request at the moment
```shell
curl -s -w "\nHTTP %{http_code}\n" \                                                                                                                                       
  -X POST "http://127.0.0.1:5011/collections/obs/items" \
  -H "Content-Type: application/geo+json" \
  -d '{"type":"Feature"}'
```
would return:
```json
{
    "code":"InvalidParameterValue",
    "type":"InvalidParameterValue",
    "description":"Item is not valid, please check and validate payload"
}
```
HTTP 500

Suggest to change into:

 `raise ValidatorValidationError(msg, user_msg=f'{msg}: {err}')`

that gives HTTP 400 with:
```json
{
    "code":"InvalidParameterValue",
    "type":"InvalidParameterValue",
    "description":[
        "[]: 'properties' is a required property",
        "[]: 'geometry' is a required property"
    ]
}
```

francbartoli (changes requested):

tests/validator/test_geojson_validator.py
```diff
@@ -0,0 +1,95 @@
...
+
+def test_valid_geojson_data(validator_def, valid_geojson_data):
+    v = GeoJSONValidator(validator_def)
+    assert v.validate(valid_geojson_data) is None
+
+
+def test_invalid_geojson_data(validator_def, invalid_geojson_data):
+    v = GeoJSONValidator(validator_def)
+    with pytest.raises(ValidatorValidationError):
+        v.validate(invalid_geojson_data)
```
francbartoli, line 95: Suggest to add one more test:
```python
def test_transaction_validation_answers_400(validator_def, api_):
    req = mock_api_request(data=b'{"type": "Feature"}')
    headers, status, content = manage_collection_item(
        api_, req, 'create', 'obs')
    assert status == HTTPStatus.BAD_REQUEST
    assert json.loads(content)['code'] == 'InvalidParameterValue'
```

francbartoli (changes requested):

pygeoapi/validator/base.py
```diff
@@ -0,0 +1,81 @@
...
+        Initialize object
+
+        :param validator_def: validator definition
+
+        :returns: pygeoapi.validator.base.BaseValidator
+        """
+
+        self.errors = []
+
+    def validate(self, data: Any, partial: bool = False) -> list:
```
francbartoli: `data` is promising `Any` which cannot be respected, i.e. a `dict`. The raw strict type `bytes` should be required:
```python
def validate(self, data: bytes, partial: bool = False) -> None:
    """
    Validate a raw request payload

    :param data: raw request body as received bytes
    :param partial: `bool` of whether data to be validated is a
                    partial resource (default `False`)

    :returns: `None`; raises `ValidatorValidationError` on invalid
              payloads, with `user_msg` carrying the details
    """

    raise NotImplementedError()
```

pygeoapi/validator/base.py
```diff
@@ -0,0 +1,81 @@
...
+    def __init__(self, validator_def):
+        """
+        Initialize object
+
+        :param validator_def: validator definition
+
+        :returns: pygeoapi.validator.base.BaseValidator
+        """
+
+        self.errors = []
```
francbartoli: This seems to be dead and never used

pygeoapi/validator/geojson.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self, validator_def):
+        """
+        Initialize object
+
+        :returns: pygeoapi.validator.geojson.GeoJSONValidator
+        """
+
+        super().__init__(validator_def)
+
+    def validate(self, data: Any, partial: bool = False) -> None:
```
francbartoli: This would be changed accordingly to the raw `data` type:
```python
def validate(self, data: bytes, partial: bool = False) -> None:
...
```

docs/source/plugins.rst
```diff
@@ -503,6 +503,42 @@ The below template provides a minimal example (let's call the file ``mycooljsonf
...
+
+   from pygeoapi.validator.base import BaseValidator, ValidatorValidationError
+
+   class MyCoolDataValidator(BaseValidator):
+       def __init__(self, validator_def):
+           """Inherit from parent class"""
+
+           super().__init__(validator_def)
+
+       def validate(self, data: Any, partial: bool = False) -> None:
```
francbartoli: `data: Any` --> `data: bytes`

francbartoli (changes requested):

docs/source/plugins.rst
```diff
@@ -503,6 +503,40 @@ The below template provides a minimal example (let's call the file ``mycooljsonf
...
+           """Inherit from parent class"""
+
+           super().__init__(validator_def)
+
+       def validate(self, data: bytes, partial: bool = False) -> None:
+           if partial:  # plugin does not support partial updates to a given item (PATCH)
+               msg = 'Partial validation not supported'
+               raise ValidatorValidationError(msg)
+
+           # data is a dict of incoming data, validate accordingly
```
francbartoli: data cannot be a dict now unless parse before, suggest to add a `try...except` with the parsing function to adapt this example

pygeoapi/validator/base.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, validator_def):
+        """
+        Initialize object
+
+        :param validator_def: validator definition
+
+        :returns: pygeoapi.validator.base.BaseValidator
+        """
+
+    def validate(self, data: bytes, partial: bool = False) -> list:
```
francbartoli: Return type is `None`

pygeoapi/validator/base.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Validate a data structure
+
+        :param data: `bytes` of data
+        :param partial: `bool` of whether data to be validated is a
+                        partial resource (default `False`)
+
+        :returns: `list` of validation errors
```
francbartoli: `None` instead of `list`

pygeoapi/validator/geojson.py
```diff
@@ -0,0 +1,97 @@
...
+            msg = 'Partial validation not supported'
+            raise ValidatorValidationError(msg)
+
+        LOGGER.debug(f'Validating against {SCHEMA_FILE}')
+        try:
+            data_payload = json.loads(data)
+        except json.decoder.JSONDecodeError as err:
+            msg = 'Error decoding GeoJSON'
+            LOGGER.error(f'{msg}: {err}')
+            raise ValidatorValidationError(msg)
```
francbartoli, line 82: Is it intentional to not raise `ValidatorValidationError(msg, user_msg=f'{msg}: {err}')`? That would be coherent with the log and would tell the client the reason of the returned status code. Without the client won't never see what's breaking the contract

francbartoli (changes requested):

pygeoapi/validator/geojson.py
tomkralidis: Yes this is by design (Postel's Law).  A custom validator can expose additional information as required.

The GeoJSON validator is a simple example that does basic GeoJSON checking, or @francbartoli do you see utility of using in production?  If there is no concern about echo'ing user input, we can consider.

francbartoli: if it is just an example that doesn't pretend to be production-ready then I would suggest to advertise in the docs

francbartoli (approved):

tomkralidis:
Merging given [RFC7 vote](https://pygeoapi.io/development/rfc/7/) and PR approval.  Thanks for the valuable feedback and comment!

